PR #12, opened 2017-05-15
Add base, text and dot style support from theme.

valinaga:
adding base, text and dot style to theme which propagates to calendar and day component.


tautvilas:
Hey, I will not merge this pull request, because we are planning to have more flexible day style customization via different means. This solution only works for basic marking type dates but not for interactive type marking

PR #13, opened 2017-05-15
Add ability to change first day of week. Add maxDate parameter to Calendar and CalendarList.

nulleof:


slorber:
Hi,

Wouldn't it be nice to add the firstDay parameter to the locale settings also?

Issue #2, opened 2017-05-09
Compatibility with react-native-web


peggyrayzis:
Hey there! 🙋 This calendar module looks awesome. We would like to use it in our `react-native-web` project but are running into an issue because of the `.ios` and `.android` file extensions for some of the assets. My suggestion is to provide a default for these assets so our Webpack build can resolve them.

The files that are giving us trouble are:

- `src/calendar/img/previous.png`
- `src/calendar/img/next.png`
- `src/agenda/style.js`

If you add these files, then we should be good to go! The only question now is whether we use iOS or Android styling as the default for the web - what do you all think? 

cc @kkemple

tautvilas:
Hi, thanks for reporting this. I set IOS files as defaults for v1.2.2. Android and iOS design differ just a tiny bit.

peggyrayzis:
@tautvilas I think we need to reopen this issue. Just tried the new release and it's still failing at `src/calendar/img/next.png` and `src/calendar/img/previous.png`. AFAIK, Webpack can't resolve images dynamically @2x or @3x like React Native can - currently the default is set at `next@2x.png` and `previous@2x.png`.

Looks like someone opened up a similar issue w/ Webpack, but it was never resolved (https://github.com/webpack/webpack/issues/3516). I'm going to follow up with Webpack to see if we can get some more info.

tautvilas:
I will try adding asset versions without `@` maybe they will be interpreted the same way as `@1x`

peggyrayzis:
Yeah, that's definitely a good solution for the short term. I think this may be a broader issue beyond this project - ideally, we would like our Webpack config to be able to resolve image paths to the appropriate scale just like in React Native. I know it's possible because [Haul](https://github.com/callstack-io/haul) does it. It's probably going to take a custom resolver though.

Once we have this all figured out, would you like me to add a section to the README on how to integrate with `react-native-web`?

tautvilas:
hey tried adding images without extension. Check if it works. If we manage to get it working with react-native-web you can prepare WEB-HOWTO.md file in pull request and I will merge it to readme or add a link to it in main readme.

peggyrayzis:
Thanks for making that change! The build runs without errors and the calendar renders! 🎉 Only problem now is that the arrows don't show up because `tintColor` (how the arrow images are filled) is a React Native only property with no plans to be supported in RNW in the future (https://github.com/necolas/react-native-web/issues/362). 😭 

[...]

PR #28, opened 2017-05-25
feature(calendar): support custom arrow components

kkemple:
This PR brings full web compatibility to address issue #2

Changes made:
Remove mixed import/module.exports for Webpack 2 support
Allow `<CalendarHeader />` component to take `leftArrow` and `rightArrow` props allowing any component to be used for arrow replacement.

scores-and-schedule

### DIFF
--- a/src/calendar/header/index.js
+++ b/src/calendar/header/index.js
@@ -1,13 +1,8 @@
-import React, {Component} from 'react';
-import {ActivityIndicator} from 'react-native';
-import {
-  View,
-  Text,
-  TouchableOpacity,
-  Image
-} from 'react-native';
+import React, { Component } from 'react';
+import { ActivityIndicator } from 'react-native';
+import { View, Text, TouchableOpacity, Image } from 'react-native';
 import styleConstructor from './style';
-import {weekDayNames} from '../../dateutils';
+import { weekDayNames } from '../../dateutils';
 
 class CalendarHeader extends Component {
   constructor(props) {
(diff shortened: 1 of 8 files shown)


kkemple:
hmm, it seems import is breaking the build, missing babel plugin i imagine, I will take a look today

src/calendar/header/index.js
```diff
@@ -42,11 +42,17 @@ class CalendarHeader extends Component {
     if (!this.props.hideArrows) {
       leftArrow = (
         <TouchableOpacity onPress={this.substractMonth} style={this.style.arrow}>
-          <Image source={require('../img/previous.png')} style={this.style.arrowImage}/></TouchableOpacity>
+          {this.props.renderArrow
+            ? React.cloneElement(this.props.renderArrow('left'))
```
ajwhite: What's the intention behind cloning the element here if you're not adding properties to it? Versus just
```js
? this.props.renderArrow('left')
```

kkemple: actually that is a good point, in the last commit it was explicit `leftArrow` and `rightArrow` props so doing `React.cloneElement(this.props.leftArrow)` seemed better than `<this.props.leftArrow />`. Now that its a method there is no reason for it. Will remove. :smile:

ajwhite: Cool! I was almost hoping to have learned something new 😆

tautvilas:
Yes, there is no babel plugin. I was hoping to keep unit tests simple without need of transpilation. Unfortunately webpack does not support module.exports :/

kkemple:
I was able to update test setup with minimal intrusion into current setup, I hope this is acceptable! if not more than willing to make any adjustments :smile:

kkemple:
:tada: thanks!! Awesome work on this module by the way! Love it!

tautvilas:
@kkemple wanted to ask some claification about module import syntax. I see that this file:
https://github.com/wix/react-native-calendars/blob/master/src/dateutils.js

Still uses oldschool require = module.exports

Is this ok for webpack? I am considering possibility of moving back to jasmine without babel, because babel config is giving us problems with react native. Maybe it would be possible to use require/module.exports in `src/interface.js` too?

kkemple:
so sorry for delayed response, yes using require/module.exports is fine its mixing import/module.exports that webpack can't parse

Issue #39, opened 2017-05-31
Flow and proptypes


sibelius:
What do you think of adding proptypes and flow types for this package?

tautvilas:
Prop types could be nice addition, not sure about flow - did not use it

sibelius:
Flow look like this:

- https://github.com/jwohlfert23/react-native-tag-input/blob/master/index.js
- https://github.com/react-community/react-navigation/blob/master/src/TypeDefinition.js

tautvilas:
Merged prop types PR #43 do not plan of adding flow types for now

PR #43, opened 2017-06-01
Add propTypes to components

sibelius:
see #39

I did not add any `isRequired`, but I think we should


tautvilas:
Thanks! no need for `isRequired` because no props are required by default

andrerfneves:
Awesome ^^

PR #56, opened 2017-06-10
Fixing typo in git clone

yemarnevets:


tautvilas:
Hey, I think my version is correct.

This will not work for me: `git clone git@github.com/wix/react-native-calendars.git`

yemarnevets:
Thanks for the reply! What OS are you using? I'm on a mac so that might be
why?

tautvilas:
I'm on mac too. do you get error if you do `git clone git@github.com:wix/react-native-calendars.git`?

yemarnevets:
It did earlier, but I can try again later.
It might've just been an unrecognised key error...

Issue #55, opened 2017-06-08
Any plan for providing gesture?


Symous:
Is there any plan for providing gesture support ? such as swipe down to expand the agenda ?

tautvilas:
it's something we were thinking about a lot (swipe down was an original idea). However smooth swipe probably can be implemented only using native solutions. A good library for that task would be https://github.com/wix/react-native-interactable. However I want that this lib by default would be not requiring native dependencies. So Maybe dragging can be optional feature and you would not have to link native libs if you want to use tappable knob. Also it would be interesting to se JS-only swipeable solution and how performant it is.

tautvilas:
I see know that you meant gesture and what I meant is dragging knob to expand agenda. Maybe implementing guesture would be easy? What do you suggest?

Symous:
@tautvilas it's possible to implement guesture without linking native libs. just like a Drawer(Side Bar).I will try to achieve this and pull a request later since I'm not familiar with gesture of React Native now. :-)

Louis345:
This sounds really awesome.

tautvilas:
Solved by #85

PR #57, opened 2017-06-11
Add optional native gesture handler (makes the agenda calendar natively draggable)

agarant:
React-native-interactable make it possible to have very fluid interactions. This pull request add the possibility to drag the calendar down.

As suggested in #55, the functionality is optional. If draggable is set to false, the current option will be used (click on the knob to expand the calendar).

I did not have an iphone to test on IOS. 

There is probably some work left to do (and doc), but I wanted to have feedback on it (whether or not it make sense for this library to have this kind of integration) before putting more work into it.

draggable_calendar

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
   "license": "MIT",
   "dependencies": {
     "prop-types": "^15.5.10",
+    "react-native-interactable": "0.0.9",
     "xdate": "^0.8.0"
   },
   "peerDependencies": {
--- a/src/agenda/index.js
+++ b/src/agenda/index.js
@@ -183,12 +183,12 @@ export default class AgendaView extends Component {
       this.calendar.scrollToDay(day, CALENDAR_OFFSET, true);
     }
     
-    // if (this.props.loadItemsForMonth) {
-    //   this.props.loadItemsForMonth(xdateToData(day));
-    // }
-    // if (this.props.onDayPress) {
-    //   this.props.onDayPress(xdateToData(day));
-    // }
+    if (this.props.loadItemsForMonth) {
+      this.props.loadItemsForMonth(xdateToData(day));
+    }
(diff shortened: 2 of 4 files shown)


Louis345:
This is so awesome.

tautvilas:
Hey, this is amazing! I will check it out on iphone, we definitely want to have this feature.

tautvilas:
Hi for some reason I get red screen on android agenda. I only did `react-native link react-native-interactable` after `yarn install`. On ios if `draggable=false` there is some problem because it is not possible to interact with scroll view. If `draggable=true` no errors and I can interact with scroll view + drag but the calendar view looks broken. Attaching screens.
screen shot 2017-06-12 at 13 18 56
screen shot 2017-06-12 at 17 41 27

Louis345:
Just a concern: If we use react-native-interactable , won't that create a dependency?  As of now, react-native-interactable is not working with expo so that might break something. If this has already been  considered, then just ignore  me.

sibelius:
`react-native-interactable` will an optional

agarant:
Im glad that there is interest in having this feature in! I updated the reference on my fork which should fix the broken design. I was not able to reproduce the red screen on Android, I will keep looking. 

However I broke few things that I need to fix while doing this. Also, when draggable={false}, the animation is sometime interrupted because the scrollToDay is triggered during the animation. I was able to prevent this by changing the way the animation is done (using translateY instead of height which allows to set useNativeDriver to true). However, I have some trouble with scrollToDay, it seems that the offset does not necessarily have the same effect.

jkomyno:
+1, very interested in this feature, kudos to @agarant

suyogkc:
@agarant  This will be a really good addition to this library . Awesome :+1:

tautvilas:
Solved with #85

PR #66, opened 2017-06-20
Updates to <Calendar> styles to allow customization

kcfgl:
This pull request gives the ability to set appStyle props to change the fontFamily and fontSize for <Calendar> component. 

The default "System" font and default sizes are retained and used as a backup if the props are not defined.

There likely is a more elegant way to do this and if you have any suggestions I'd be happy to make modifications.

**New props**

- textDayFontFamily
- textMonthFontFamily
- textDayHeaderFontFamily
- textTodayFontFamily
- textDayFontSize
- textMonthFontSize
- textDayHeaderFontSize

**Original**
screen shot 2017-06-20 at 12 17 38 pm

**With props added**
screen shot 2017-06-20 at 12 16 44 pm

### DIFF
--- a/src/calendar/day/basic/style.js
+++ b/src/calendar/day/basic/style.js
@@ -11,7 +11,8 @@ export default function styleConstructor(theme={}) {
     },
     text: {
       marginTop: 4,
-      fontSize: 16,
+      fontSize: appStyle.textDayFontSize,
+      fontFamily: appStyle.textDayFontFamily,
       fontWeight: '300',
       color: appStyle.dayTextColor,
       backgroundColor: 'rgba(255, 255, 255, 0)'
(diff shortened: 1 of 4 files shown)


tautvilas:
Hi, thanks for pull request. Is `System` font family something that is defined in the docs? Cause I can't find it.. Also plz check lint

kcfgl:
It looks like it hasn't been properly documented but I was able to find references in some commits as well as https://github.com/facebook/react-native/blob/235b16d93287061a09c4624e612b5dc4f960ce47/React/Base/RCTConvert.m#L612

They have defined a font called "System" to be the default font for the platform. 

And I'll check lint, looks like I missed something pretty simple!

sibelius:
System works fine

PR #67, opened 2017-06-20
Add option to disable all dates and provide array of dates to enable

kcfgl:
This pull request creates two new <Calendar> props: 
```
// Collection of dates that should be enabled. Default = {}
    enabledDates: PropTypes.array,

// Disable dates by default. Default = false
    disabledByDefault: PropTypes.bool,
```
They can be used like this:
```
<Calendar
    ...
    disabledByDefault = {true}
    enabledDates = {['2017-06-20', '2017-06-23', '2017-06-25']}
    ...
/>
```
During the initial styling, if disabledByDefault == true, then it checks if the day is included in the enabledDates array.  

In pressDay it also checks to see if the new props exist and simply returns; if true.  It uses the same XDate library to format the day into this string format 'yyyy-MM-dd.'

There may be a better way to handle checking the array, especially with extremely large data sets. 

**Screenshot**
screen shot 2017-06-20 at 4 43 59 pm

**LIMITATION**
This does not work with date ranges.  For example, if I have the array of values above, the user can click on 6/20 and then on 6/25 and complete the range.  I don't believe this is something that you will want to solve for in the library, and should instead be handled in the individual app code.

If there are modification suggestions or areas you need improved, please let me know and I'll try to take care of them.

### DIFF
--- a/src/calendar/index.js
+++ b/src/calendar/index.js
@@ -115,7 +115,7 @@ class Calendar extends Component {
     const maxDate = parseDate(this.props.maxDate);
 
     if(this.props.enabledDates && this.props.disabledByDefault) {
-      if(!this.props.enabledDates.includes(day.toString("yyyy-MM-dd"))) {
+      if(!this.props.enabledDates.includes(day.toString('yyyy-MM-dd'))) {
         return;
       }
     }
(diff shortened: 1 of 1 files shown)


asantos00:
Is this planned to be merged?

tautvilas:
The idea itself is not bad, however I don't like this implementation. I would propose such implementation: disabledByDefault - this controls if all dates are disabled by default. However we should not have enabledDates prop and just use markedDates prop with disabled:false attributes for days that should be enabled.

asantos00:
All right, I'm going to work on that!

asantos00:
Done at https://github.com/wix/react-native-calendars/pull/193

tautvilas:
Solved in #193

PR #82, opened 2017-07-04
renderItem callback could receive the date

slorber:
I don't really see why renderEmptyDate would receive it and not renderItem. I need this because my last day item is always a CTA button to "add availability". Without the date I have to resort on trying to get back the date from the items list of my component which is a bit annoying


slorber:
Hmmm actually not so sure it would work, only the first row receives the date right
?

tautvilas:
yes with this fix only first item in day will receive date

slorber:
Hi,

I'll contribute new PR's to this project later when I'm back working with calendars, so I'm closing this one for now

PR #85, opened 2017-07-07
Draggable agenda calendar

overengineered:
#55 Allows user to drag the knob in agenda calendar to expand it. Also expands the calendar if user simply presses on the knob.

Uses native driver, so should be performant.


naxiaoguang:
how to use?

tautvilas:
checked on android & ios, looks great 👍

Louis345:
This is great. How can we use this?

PR #91, opened 2017-07-12
Replaced example with pure JS dependencies, as per issue #45

jkomyno:
- **Motivation**: https://github.com/wix/react-native-calendars/issues/45
- **Linting status on edited files**: 0 errors

I've used [create-react-native-app](https://github.com/react-community/create-react-native-app), so just install the deps and run `yarn run android` or `yarn run ios`. Thanks to this PR you can now try the example directly on Expo.

Issue #45, opened 2017-06-01
Use only js based packages for example


sibelius:
What do you think of refactor the example to use create-react-native-app, and use react-navigation instead of react-native-navigation

I have this error when running the example:

`RCCManager.h file not found`

tautvilas:
I think this probably would be a good move because example module would become easier to maintain

jkomyno:
I'm doing this, PR incoming soon

jkomyno:
@tautvilas done, please review it when you have time

tautvilas:
Currently decided to stick with RNN

PR #109, opened 2017-07-20
Created pure JS example, fixing #45

jkomyno:
@tautvilas https://github.com/wix/react-native-calendars/pull/91 had some issues, so I have done the example again in pure JS. It seems that either `yarn` or `expo` sometimes need `yarn start -- --cache-clear` to install the correct deps. Npm 4 works completely fine though.


tautvilas:
I have resolved my yarn issue. It was this bug: https://github.com/yarnpkg/yarn/issues/2515

I will review & merge this soon.

tautvilas:
I get this error when doing `npm run ios` :(
screen shot 2017-07-20 at 18 11 29

jkomyno:
I ain't getting that, I was able to get it working both in Android and iOS. Can you please try removing `node_modules` and use `npm` instead of `yarn`?
Have you cleared your cache (e.g. watchman, npm cache clear, rn packager)?

sibelius:
this is a problem with react-native 

"react": "16.0.0-alpha.6",
    "react-native": "0.44.0",

try to remove `^` from package.json

tautvilas:
using react native 0.44.0 with yarn gives me another error. npm with 45.1 works fine (no problem)
screen shot 2017-07-20 at 18 35 06

It seems that react-native-navigation example is more stable than default `react-native-create-app` solution with `reactive-navigation` :/

jkomyno:
@sibelius, I have removed node_modules and have run `npm i` and didn't have any issue. Note that I'm using npm@4.0.3. However I can remoe the "^" from package.json, no problem.

@tautvilas With create-react-native-app you have a different way of upgrading/downgrading react-native version. Please take a look [here](https://github.com/react-community/create-react-native-app/blob/master/VERSIONS.md).

tautvilas:
npm works for me too no problem, but I am not sure we can have example project that does not support yarn

jkomyno:
@tautvilas you're actually right, maybe we should let the team behind `yarn` know that we're having this kind of unexpected behaviour. Imho this isn't related to create-react-native-app

tautvilas:
Apparently doing the npm cache clean and rebooting computer solved the issue with AccesibilityInfo module :) however I noticed that `yarn link` does not work with expo. This makes it difficult to develop module (when I want to check changes that I do in realtime on example project). Does anybody know how this could be solved?

slorber:
@tautvilas some other opensource projects (like RN Storybook) are using [Haul](https://github.com/callstack-io/haul) packager instead of Metro as it supports linking

Metro does not support linking which makes it hard to contribute to RN projects and also use tools like Lerna

tautvilas:
Decided to stick with react-native-navigation for now

Issue #99, opened 2017-07-16
Changing state on onMonthChange doesn't move to the next/prev month


lgalant:
I'm trying to show the busy days in the calendar using the markedDates property. For the month I'm currently showing I build the markedDays hashtable dynamically from a SQLite database. 
When I click the next arrow in the calendar I use onMonthChange to call the method that updates the markedDates list. The list is updated correctly, however the month doesn't move to the next.  
 I believe I was able to isolate the issue using the code provided in the example. Here's an excerpt of ./example/src/screens/calendar.js with my changes:

`export default class CalendarsScreen extends Component {
  constructor(props) {
    super(props);
    this.state = {
      markedDates :{
        '2017-07-24': {selected: true, marked: true},
        '2017-07-25': {marked: true},
        '2017-07-30': {marked: true},
      }
    };
    this.onDayPress = this.onDayPress.bind(this);
  }

  monthChange(month) {
    console.log('month changed2',month);
    this.setState({
      markedDates :{
        '2017-07-01': {selected: true, marked: true},
        '2017-07-02': {marked: true},
      }
    });

  }
  render() {
    return (
      <ScrollView style={styles.container}>
        <Text style={styles.text}>Calendar with selectable date and arrows</Text>
        <Calendar
          onDayPress={this.onDayPress}
          onMonthChange={(month) => {this.monthChange.bind(this)(month)}}
          style={styles.calendar}
          current={'2017-07-17'}
          hideExtraDays
          markedDates={ this.state.markedDates}
`

If I comment out the "setState" call on the monthChange method, the next/prev arrows work correctly. However if I uncomment the setState, the month doesn't change when the arrows are clicked.

Great work btw!

lgalant:
I solved the problem by changing method updateMonth on src/calendar/index.js to this:

`  updateMonth(day, doNotTriggerListeners) {
    if (day.toString('yyyy MM') === this.state.currentMonth.toString('yyyy MM')) {
      return;
    }
    if (!doNotTriggerListeners) {
      const currMont = day.clone();
      if (this.props.onMonthChange) {
        this.props.onMonthChange(xdateToData(currMont));
      }
      if (this.props.onVisibleMonthsChange) {
        this.props.onVisibleMonthsChange([xdateToData(currMont)]);
      }
    }
    this.setState({
      currentMonth: day.clone()
    })
  }`

If you are OK I can get you a PR
Thanks!

tautvilas:
please do so that I can see a diff

[...]

PR #112, opened 2017-07-22
remove componentWillReceiveProps

lgalant:
This is the proposed fix for issue #99 
 componentWillReceiveProps method is resetting the currentMonth when a current date is passed as props overriding the next/prev arrow behavior. The current month behavior is already covered by the constructor.


tautvilas:
this is rendering optimization that should not be removed

PR #116, opened 2017-07-27
Add fallback to support RN version < 0.44

booker-laputa:
To check if ViewPropTypes is available, if no, fallback to use View.propTypes


booker-laputa:
Unified at https://github.com/wix/react-native-calendars/pull/118

PR #118, opened 2017-07-27
Add fallback support to RN < 0.44

booker-laputa:


nathsimpson:
I don't think this was quite right :( I've made a PR https://github.com/wix/react-native-calendars/pull/1089

PR #145, opened 2017-08-23
Add missing "backgroundColor" theme prop

slorber:
This made me loose some time due to not understanding why `calendarBackground` was not applying to the reservation-list


tautvilas:
Thanks!

PR #147, opened 2017-08-23
Agenda: allow custom markedDates

slorber:
current markedDates=items default setting is not flexible enough for my usecase which if to mark the date only if one row of the day match a predicate


tuanmai:
👍 for this PR. Please consider to merge this 😄

tuanmai:
@slorber I think you should update the README too.

slorber:
I've added it to readme example

tautvilas:
Thanks for PR. also will add markingType prop possible for agenda, otherwise this feature would not be complete

PR #157, opened 2017-08-25
Day: allow custom marking to override state marking

slorber:
Currently if I provide 

`markedDays={{"dayString": {disabled: false}}}` to Calendar it won't necessarily "not mark" the day, particularly if the Calendar has computed state="disabled"

I think it makes sense to let the user override the state computated in parent calendar if he wants to


tautvilas:
Resolved this in 1.8.0 version

slorber:
thanks :)

PR #161, opened 2017-08-28
Add hideDayNames prop

maksimdegtyarev:
Add hideDayNames props. If its `true`, day names arent visible. Now it looks like this:
iphone 6 ios 10 3 14e8301 2017-08-28 11-19-33


tautvilas:
Thanks for PR!

maksimdegtyarev:
You're welcome ;)

PR #164, opened 2017-09-01
fix(calendar-list): fixes onscroll returning early on web

kkemple:
Hey hey! Thanks again for such an amazing library! Really love the API and its such a pleasure to work with! We are using this library for our cross-platform component lib and on web the calendar list view does not load months once the scroll into view due to the onScroll handler returning early for any platform that is not `android`. In a `react-native-web`/universal component system this causes web to not work.

____
early-return
____

I have updated the check to only return specifically for ios and this fixes the issue. Let me know if you need me to do anything else! Thanks again for all of the hard work!

___
with-ios-specific-check
___

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 npm-debug.log
 .idea
 yarn.lock
+.vscode
--- a/src/calendar-list/index.js
+++ b/src/calendar-list/index.js
@@ -188,7 +188,7 @@ class CalendarList extends Component {
   }
 
   onScroll(event) {
-    if (Platform.OS !== 'android') {
+    if (Platform.OS !== 'android' && Platform.OS !== 'web') {
       return;
     }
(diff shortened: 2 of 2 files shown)


kkemple:
ugh, just saw my prettier formatting got in there, will fix today

tautvilas:
Hi man, sorry for late reply - was on holidays :) Could you fix the formatting so I could better see what is the fix? thx

kkemple:
oh, I thought i did already! sorry about that 😊 
I'll get this pushed today

kkemple:
okay, updated!

kkemple:
thanks! 🎉

tautvilas:
Actually yesterday migrated calendar to flatList, so this PR became irrelevant, plz check if everything works for you with latest version

kkemple:
will do!

PR #167, opened 2017-09-02
Add missing prop disableMonthChange to Calendar

### LINKED ISSUE #166: Calendar disableMonthChange prop is currently not implemented
## Description

The prop `disableMonthChange` on Calendar currently does nothing.

## Expected Behavior

Currently `CalendarList` sets the prop `disableMonthChange` on the `Calendar` component, from what I can see, this prop is to make sure the month is not changed the user clicks on the greyed out days of the next/prev month (in case `hideExtraDays` is set to `true`).

## Observed Behavior

The prop does nothing, since it's not implemented, clicking on the greyed out days does change the current month on that specific calendar in the calendar listing, making the calendar listing really strange, with duplicated months on it.

This was running the code on master as of today.

Gifs are better than words:

issue

JCMais:
Fixes #166


tautvilas:
Good catch, thanks :)

Issue #100, opened 2017-07-16
More powerful theming


wvteijlingen:
I would like to use this great component in an app that used bold fonts by default. The current theming options don't allow overriding the font weight.

I think it would be nice if we could override all the styles, by passing in a React.StyleSheet object. This means that we could override things such as font weights, margins, paddings etc. It would also make sure that theming is flexible enough without having to add separate theme options for each item.

I think this would be a more flexible and future-friendly way to go about theming. What do you guys think?

tautvilas:
Similar to #10

tautvilas:
I'm open for suggestions how more flexible theming could be implemented

SukhKalsi:
As this has been linked in with issue #139 I would say the first suggestion is to remove any `fontWeight` whenever there is a defined `fontFamily` - kind of defeats having custom fonts IMO.

wvteijlingen:
Hmm yeah, good points @tautvilas.

I think we nice to look at the calendar and determine the elements that it is comprised of. For example: `Day`, `DayNumber`, `DayTitle`, `MonthTitle`, `PreviousArrow`, `NextArrow` etc.

Each of these elements then has its own style options. If you look at the current theming options there are already a lot of variables that can be grouped together. For example: `textDayFontFamily` + `textDayFontSize`, `textDayHeaderFontSize` +  `textDayHeaderFontFamily`.

This could perhaps be turned into an API like this:

```javascript
<Calendar elementStyles={elementStyles} />

const elementStyles = StyleSheet.create({
  day: {
    // User can style the day box here
  },
  daySelected: {
    // User can style the selected day box here
  },
  dayNumber: {
    // User can style the day number here
  },
  dayNumberSelected: {
    // User can style the selected day number here
  },
  dayTitle: {
    // User can style the day box here
  }
}
```

The Calendar component then merges it's internal StyleSheet with the sheet provided by the user. This allows the user to override practically everything that you allow to override, while at the same time not exploding the `theme` prop with separate keys.

wvteijlingen:
I just checked https://github.com/christopherdro/react-native-calendar, and it uses the same theming options I mentioned.

aforty:
100% agree that @wvteijlingen's approach is the way to go. This should allow anyone to override anything while still allowing this module to ship with its own default look and feel.

This can be done without interfering with implementation details.

asantos00:
[...]

PR #174, opened 2017-09-13
Allow month header to be left justified and have styled font weight

bluwave:


tautvilas:
Hi, currently we are not sure how to implement flexible styling. This PR adds a lot of very specific settings to theme, but I would like to keep theme more basic. Please join #100 thread for discussion how to implement more flexible theming.

PR #184, opened 2017-09-22
Custom rendering a day when markingType is 'period'

mhotakai:
I needed to add another row for each day in the calendar with a custom text:

screen shot 2017-09-22 at 13 36 48

This works only if markingType is 'interactive'. 

I wasn't sure how it should be done with simple markingType.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,8 @@ LocaleConfig.defaultLocale = 'fr';
   firstDay={1}
   // Hide day names. Default = false
   hideDayNames={true}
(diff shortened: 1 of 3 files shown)


tuanmai:
@mhotakai could you also add `renderDay` attribute to `CalendarList` and `Agenda` too?

tautvilas:
renderDay is a good idea, but maybe it should be implmemented in generic way for all components.

tautvilas:
Please check how to ovrerride day component if you need more flexibility:

https://github.com/wix/react-native-calendars#overriding-day-component

PR #191, opened 2017-09-28
Add styling customization options

asantos00:
This PR adds a few styling customisation options as well as some minor fixes.

However, in the future (maybe in a new version, cause it's a breaking change) it would be cool to have a more structured way of styling each component so we can allow full customisation of the styles while keeping a default.

Example:

```
const headerStyle = {
  // all the header styles,
  ...appStyle.header
}

const dayStyle = {
  // all the day styles,
  ...appStyle.day
}
```

This way it would be easy for the user to change whatever we wants regarding styles (instead of just some props that we keep adding so we can customize) and we can even use the "normal" stylesheet api, instead of names like `textSectionTitleColor` it turns out even harder to use and document.

### DIFF
--- a/src/calendar/header/style.js
+++ b/src/calendar/header/style.js
@@ -16,7 +16,8 @@ export default function(theme={}) {
       fontFamily: appStyle.textMonthFontFamily,
       fontWeight: '300',
       color: appStyle.monthTextColor,
-      margin: 10
+      margin: 10,
+      ...appStyle.monthText
     },
     arrow: {
       padding: 10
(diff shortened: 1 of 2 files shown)


tautvilas:
Hi, please check out #100 discussion about styling

tautvilas:
resolved using stylesheets ovveride https://github.com/wix/react-native-calendars#advanced-styling

asantos00:
Sorry If I'm mistaken, but when you use stylesheets override doesn't that overrides the styles defined on the markedDates?

tautvilas:
hey, could you give an example?

asantos00:
Hey

Imagine that I define the color of the selected days by using this theme override you talked about, can I override that selected days color in my `markedDates` object? (that happens when I want to have multiple selected periods with different colors)

PR #201, opened 2017-10-11
Added horizontal mode for calendarlist

ynsgnr:
Simply add horizontal or horizontal={true} to props. In horizontal mode all items rendered inside a square view due to scroll issues. Also you can set calendarHeight now but its not fully tested.


tautvilas:
Hey, thanks for PR, this could be merged only if solution is working good on ios+droid and not only in rectangle.

PR #202, opened 2017-10-13
year flow 

bansalayush:
on click of year a list of year appears from which user can select a particular year. 
GIF -> 
https://user-images.githubusercontent.com/9965653/30317918-395f9a32-97c9-11e7-9025-3fe408f95d77.gif


tautvilas:
Hi maybe this could be considered if year selector was implemented as a separate react component. Also don't understand why dateutils.spec.js was deleted

bansalayush:
Hey @tautvilas , will implement year selector as a separate component .

bansalayush:
Hi , sorry didn't notice that dateutils.spec.js was deleted .Didn't do it purposefully .
And if year-selector is to be implemented as a separate component then I suppose there will be a need of navigator within Calendar component to navigate b/w month and year list .

PR #204, opened 2017-10-16
Hotfix/standardization

Zidail:
## What

 - Updates `dateutils` and its spec to use `const` and `let` instead of `var`
 - Replaces `xdate` with `XDate` in `dateutils`

## Why

Consistency.
`var` is almost never used except in the date util. `let` and `const` are also present in the file.
Other files mention `xdate` as a param and `XDate` as the import, but the util file uses `xdate` for the import.


tautvilas:
Thanks!

Issue #138, opened 2017-08-15
CalendarList always shows minimum of three months in scroll


SukhKalsi:
## Description

CalendarList component always shows three months.

## Expected Behavior

When setting the props, I'd expect the CalendarList to show less than three months if required

## Observed Behavior

No matter what I set in the props the CalendarList will default to showing minimum of three months.

## Environment

* `npm ls react-native-calendars`: react-native-calendars@1.5.6
* `npm ls react-native`: react-native@0.45.1

Also specify:

1. Phone/emulator/simulator & version: iOS Simulator on iOS 10.3

## Reproducible Demo

Below will show the current month, along with the next two months. I only want the current month and the next month to be shown. Also if i wish to set the `futureScrollRange` to be zero, to only show the current month.
  
```
import { CalendarList } from 'react-native-calendars';

class Demo extends Component {

  render() {
    return (
      <CalendarList
        pastScrollRange={0}
        futureScrollRange={1}
      />
  }
}
```

tautvilas:
Hi, yes there is currently such limitation.  Maybe you could also overcome this by using two <Calendar /> components instead of <CalendarList> ?

SukhKalsi:
Out of curiosity where in this library are we forcing minimum of three months?

tautvilas:
I think here:

https://github.com/wix/react-native-calendars/blob/master/src/calendar-list/index.js#L52

There would be some bugs if there are less than 3 months, I think. need to check it.

nulleof:
I set `pastScrollRange={0}` and `futureScrollRange={1}` and see error `No item for index 2`.
It is working correctly when `futureScrollRange` is greater or equal than 2.

nulleof:
I've added a PR to fix this behavior https://github.com/wix/react-native-calendars/pull/207

tautvilas:
resolved in 1.14.1

PR #207, opened 2017-10-18
CalendarList always shows minimum of three months in scroll

nulleof:
Related to issue https://github.com/wix/react-native-calendars/issues/138

Allows to show desired amount of months in given range.

Fixes problem with `No item for index 2` when months range is lower than 3.


tautvilas:
I want to explain my implementation a bit: pushing text into rows means that calendar on these rows would not be rendered. Calendar is very heavy component so we want to reduce rendering of it as much as possible. That is why visibleRowsChange function decides what calendars should be drawn my putting either date objects or text in rows array. In your fix you push date object to all rows that means it should significantly cripple calendar list speed in big calendar lists.

nulleof:
Well, I see now. I'll return old limitation back.

tautvilas:
thanks!

PR #213, opened 2017-10-25
Implement color marking for basic marking type

AstRonin:


AstRonin:
I have finished resolving conflict

tautvilas:
Please update readme and add a small example to example project

AstRonin:
Ok, added. Also I changed name of the param to **docColor** in case if you will decide to combine Marked and Period flows.

PR #223, opened 2017-11-06
Bug fix for multi-dot markers

jaruyot:
Fixed bug when multi-dot markers change and the change doesn't re-render the multi-dot day component.


tautvilas:
@jaruyot plz review

PR #233, opened 2017-11-15
Allow changes to a marked day's font size and weight

karandkanwar:


tautvilas:
Advanced styling could be done with stylesheet override: https://github.com/wix/react-native-calendars#advanced-styling

Let's keep the marking api simple

PR #241, opened 2017-11-20
just made it a little bit more flexible

KirillGudkov:
It allows use custom components instead of activity indicator

### DIFF
--- a/src/agenda/reservation-list/index.js
+++ b/src/agenda/reservation-list/index.js
@@ -179,7 +179,7 @@ class ReactComp extends Component {
 
   render() {
     if (!this.props.reservations || !this.props.reservations[this.props.selectedDay.toString('yyyy-MM-dd')]) {
-      return (<ActivityIndicator style={{marginTop: 80}}/>);
+      return this.props.renderEmptyData();
     }
     return (
       <FlatList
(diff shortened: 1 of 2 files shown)


src/agenda/reservation-list/index.js
```diff
@@ -179,7 +178,7 @@ class ReactComp extends Component {
 
   render() {
     if (!this.props.reservations || !this.props.reservations[this.props.selectedDay.toString('yyyy-MM-dd')]) {
-      return (<ActivityIndicator style={{marginTop: 80}}/>);
+      return this.props.renderEmptyData();
```
albinekb: I think you should set the `ActivityIndicator` that was here before as default, so there's no breaking change? 🤔

albinekb: yes, but you could still override it with something else than a spinner, but it should default to show the same ActivityIndicator as before, to not break other peoples apps 👍

tautvilas:
thanks for PR!

tautvilas:
I fixed one bug, there was ActivityIndicator import missing in reservation-list/index.js

albinekb:
Will fix @tautvilas 🙏

Issue #149, opened 2017-08-23
Disabled days => disable onPress?


slorber:
Currently when a calendar item is marked as disabled, it's still possible to press on it.

Is it a mistake or intended behavior?

If it is intended, can we permit to disable press as well?

tautvilas:
Handling on press depends on client. By disable press you mean remove touchable effect?

slorber:
@tautvilas my usecase is to provide custom marked dates with the agenda, and to disable all dates of the past. Disabling dates is possible thanks to my other PR but after being disabled they still can be clicked in the agenda

nathanloisel:
This is my code for ignoring onDayPress on disabled dates :
`_onDayPress = (day) => {
    const { disabledDates } = this.props;
    const value = new Date(day.dateString);
    if (disabledDates.some(date => moment(date).format('YYYY-MM-DD') === day.dateString)) {
      return;
    }
    this.setState({ value }, () => {
      this.props.onChange(value);
      this._closeCalendar();
    });
  };`

slorber:
@nathanloisel I'm using the Agenda, and I can't override the `onDayPress`, there's only a callback that is called but I can't prevent day press from this callback:

```

  _chooseDayFromCalendar(d) {
    this.chooseDay(d, !this.state.calendarScrollable);
  }

  chooseDay(d, optimisticScroll) {
    const day = parseDate(d);
    this.setState({
      calendarScrollable: false,
      selectedDay: day.clone()
    });
    if (!optimisticScroll) {
      this.setState({
        topDay: day.clone()
      });
    }
    this.setScrollPadPosition(this.initialScrollPadPosition(), true);
    this.calendar.scrollToDay(day, this.calendarOffset(), true);
    if (this.props.loadItemsForMonth) {
      this.props.loadItemsForMonth(xdateToData(day));
    }
    if (this.props.onDayPress) {
      this.props.onDayPress(xdateToData(day));
    }
  }
```

Also, your code probably works but shouldn't it be made easier by the lib?

I think it's a decent behavior to expect disabled dates to not be pressable no?

slorber:
@tautvilas if you don't want to make disabled dates unpressable (which imho you should), would you accept a PR with a solution like this:

```javascript
chooseDay(d, optimisticScroll) {
    const day = parseDate(d);

    if (this.props.onDayPress) {
      const onDayPressReturn = this.props.onDayPress(xdateToData(day));
      if ( onDayPressReturn === false ) {
[...]

PR #255, opened 2017-11-29
Disabled touch events for dates marked "disabled"

MyGuySi:
This PR addresses #149 with a potential solution. Thought I'd kick off the PR to discuss possible solutions.

I think a few users feel touch events should be disabled by default for dates which are marked "disabled". I agree that this is the best solution but I'm open to other suggestions such as an additional prop to toggle this behaviour.

### DIFF
--- a/src/calendar/day/basic/index.js
+++ b/src/calendar/day/basic/index.js
@@ -88,7 +88,15 @@ class Day extends Component {
       textStyle.push(this.style.todayText);
     }
     return (
-      <TouchableOpacity style={containerStyle} onPress={this.onDayPress}>
+      <TouchableOpacity
+        style={containerStyle}
+        onPress={this.onDayPress}
+        disabled={
+          typeof marking.disabled !== 'undefined'
+            ? marking.disabled
+            : this.props.state === 'disabled'
+        }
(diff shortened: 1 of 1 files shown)


tautvilas:
thanks for PR

kcfgl:
@tautvilas I have the latest version, and it looks like this is merged, but I can still click on dates and register the onDayPress which is making it a bit more challenging to ensure my markedDates is accurately reflected.  I'm using a CalendarList.  Here is the first part of the CalendarList ...

```
<CalendarList
                    onDayPress={this.onDayPress}
                    style={styles.calendar}
                    markedDates={this.state.markedDates}
                    disabledByDefault = {true}
```
I see that the dates change color to the disabled color when I add disabledByDefault = {true}.  But I can still click on any date. 

Thanks again for bringing the ability to disable by default, its going to be great!

tautvilas:
Hi, maybe you can just filter disabled days in your onPress handler? I think there are cases when people want to display disabled dates and have them pressable too

kcfgl:
@tautvilas Makes sense.  Thanks for letting me know this is the expected behavior.

PR #274, opened 2017-12-14
Change months correctly 

### LINKED ISSUE #273: Month doesn't change although hideExtraDays set
## Description

I set `hideExtraDays` on a CalendarList, so the days of the preceding and the following month aren't visible. When `current` date is set on today (14 Dec) and I change `current` prop to a date in the first week of the following month (this case: 03 Jan 2018) the calendar doesn't scroll to the new month.

## Expected Behavior

In case `hideExtraDays` is set, `scrollToMonth()` should additionally check if the dates are in different months.

## Observed Behavior

The diff calculation in `scrollToMonth()` is not precise enough.
`diffMonths = diffMonths < 0 ? Math.ceil(diffMonths) : Math.floor(diffMonths);`

For the given time (14 Dec till 03 Jan) it returns something like 0.6 which gets floored to 0.

## Environment

* `npm ls react-native-calendars`: react-native-calendars@1.15.0 
* `npm ls react-native`: react-native@0.50.0 

Simulator, iPad Air, iOS 10.2

webwelten:
This should fix #273 

So, actually this more the hard approach to solving that problem as it doesn't consider if hideExtraDays is disabled. In that case it should take into account of the new current date is among those disabled days. I would like to help and continue the work but I'm not sure how to know that the new date is disabled because that's different in every month.

Please understand this pull request as a request for more information so that I can finish with a proper solution. 

https://github.com/wix/react-native-calendars/issues/273

### DIFF
--- a/src/calendar-list/index.js
+++ b/src/calendar-list/index.js
@@ -84,7 +84,8 @@ class CalendarList extends Component {
     const month = parseDate(m);
     const scrollTo = month || this.state.openDate;
     let diffMonths = this.state.openDate.diffMonths(scrollTo);
-    diffMonths = diffMonths < 0 ? Math.ceil(diffMonths) : Math.floor(diffMonths);
+    const isAnotherMonth = this.state.openDate.getMonth() !== scrollTo.getMonth();
+    diffMonths = (diffMonths < 0 || isAnotherMonth) ? Math.ceil(diffMonths) : Math.floor(diffMonths);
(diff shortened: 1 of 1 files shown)


tautvilas:
hi, thanks for PR, I solved this problem a bit differently, let me know if it fixes your usecase: 

https://github.com/wix/react-native-calendars/commit/81b0c876651898d96539679e0be07b02a6625ae2

PR #295, opened 2018-01-05
add sub title for date

troyz:
Hello, 
Add sub title for a date, we need the feature sometimes such as when booking hotel, usually there should have a price text under the date.

sub title for date


sibelius:
this should be a custom dayComponent instead

https://github.com/wix/react-native-calendars#overriding-day-component

maybe we can have some examples of custom day components on example

troyz:
Thanks, get it.

PR #319, opened 2018-01-29
Fix/add scrollsToTop prop

AbdallaMohamed:
This PR makes the scrollsToTop prop of the CalendarList configurable through a prop


tautvilas:
Thank's for PR, I'm considering to make scrollsToTop default false. Because it really does not makes sense in most of cases. What do you think?

AbdallaMohamed:
@tautvilas Totally agree. I just thought I make my PR backward compatible, but I totally agree it should be `false` by default. Thanks for accepting the PR by the way!

PR #321, opened 2018-01-29
getDrawingStyle: use marking data from given parameter instead of from current props

Agontuk:
Faced this issue while resetting `markedDates` to empty object. The selected dates disappeared, because `textColor` was still set to white (background was also white). Also `getDrawingStyle ` should not use `this.props.marking`, since it's getting the marking data from the parameter.


tautvilas:
nice catch! thanks for PR

PR #331, opened 2018-02-07
Support for reservation list content styling

ldrvoderic:


tautvilas:
Can you elaborate how this would work?

ldrvoderic:
@tautvilas agenda list (reservation list) from the repository is implemented like a `FlatList` React Native component. This component also inherits React Native `ScrollView` props and one of those props is also [contentContainerStyle](https://facebook.github.io/react-native/docs/scrollview.html#contentcontainerstyle). These styles will be applied to the scroll view content container which wraps all of the child views.

Here is usage example of styling applied to `Agenda` component (`stylesheet.agenda.list.content`):
`
<Agenda
    ...
    theme={{
        'stylesheet.agenda.list': {
            content: {
                paddingBottom: 40
            }
         }
    }}
/>
`

PR #355, opened 2018-02-20
Fix warning on dev about keyExtractor as string on VirtualizedList

joaopaulofilho:
By Chaging keyExtractor's type to string on agenda/reservation-list FlatList.


joaopaulofilho:
There is another PR with this fix.

PR #366, opened 2018-03-06
calendar refs

marconett:
added refs to allow calling "addMonth" from outside the calendar component.

with this, and
```
<Calendar
  onRef={ref => (this.cal = ref)}
```

I am able to call `this.cal.addMonth(1)` in the parent component.

there's probably a cleaner way of doing this, but imo, those methods should be accessible.

### DIFF
--- a/src/calendar/index.js
+++ b/src/calendar/index.js
@@ -94,10 +94,10 @@ class Calendar extends Component {
   }
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


tautvilas:
Hi, Why not just use 

ref={calendar => this._calendar = calendar} 
?

Also your code would throw error if onRef is not specified I think

PR #376, opened 2018-03-13
[bug-fix] fix subtract month onPressRight bug

PrestonWinstead:
There was a bug where the months were moving backwards when I clicked the right arrow. It was a simple mistake of calling subtractMonth instead of addMonth.


tautvilas:
Thanks for PR!

PR #378, opened 2018-03-14
Single marking type added

xxmuaddib:
Added a new day component which gives opportunity to customize container and text elements of each day.


tautvilas:
Hey, thanks for PR. I prefer if longPress and new day marking types would go as separate PRs. longPress functionality is really good, but I might have some comments for the new day style.

PR #382, opened 2018-03-15
Make multi-dot strategy to respect provided selectedColor

pinguinjkeke:
```selectedColor``` in the multi-dot section inside README.md is not working and not logically correct because code cannot decide which dot ```selectedColor``` property would be applied if we have multiple dots on one day.

This PR moved ```selectedColor``` one level up (look README.md commit) and adds missing functionality.
I'm using it in my project. It would be cool if you merge this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,8 +140,8 @@ Multi-Dot marking
 
 Use markingType = 'multi-dot' if you want to display more than one dot. Both the Calendar and CalendarList control support multiple dots by using 'dots' array in markedDates. The property 'color' is mandatory while 'key' and 'selectedColor' are optional. If key is omitted then the array index is used as key. If selectedColor is omitted then 'color' will be used for selected dates.
 ```javascript
-const vacation = {key:'vacation', color: 'red'};
-const massage = {key:'massage', color: 'blue'};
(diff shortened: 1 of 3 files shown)


tautvilas:
Hi,  as I understand currently selectedColor for dots works correctly - this setting means what color to change the dot color to when the day is selected. So it makes sense that it is specified per/dot. What you want to do is probably to have ability to change selected day (not dot) color. Can you improve your PR to address only this usecase?

pinguinjkeke:
https://github.com/pinguinjkeke/react-native-calendars/blob/3d672fa0aa4b8a48fa74660a736ca04149a71e25/src/calendar/day/multi-dot/index.js#L69

There is wrong name. Not ```selectedColor``` but ```selectedDotColor``` for dots. So it's a mispell inside README.md. I can rename ```selectedColor``` to ```selectedDotColor``` inside README.md.

Will it be ok?

tautvilas:
You're right. Yes if you could fix this mistake in Readme would be great. Also I saw that there is such mistake in example project where multidot is used, could you add a naming fix there too?

pinguinjkeke:
It's all fixed now

tautvilas:
Thanks!

PR #385, opened 2018-03-16
[Style] Sets date margin by platform type

markholland:
### Motivation

On iOS the number inside the date circle is not vertically aligned but is on Android.

This PR sets the margin top of the date to be by platform type

### Before
before

### After
after

### Android
android

### DIFF
--- a/src/calendar/day/basic/style.js
+++ b/src/calendar/day/basic/style.js
@@ -12,7 +12,7 @@ export default function styleConstructor(theme={}) {
       alignItems: 'center'
     },
     text: {
-      marginTop: 4,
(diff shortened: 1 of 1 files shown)


tautvilas:
Hi, the number is not centered because if you have date marking dot has to be displayed below number. If you want to center style and don't use day marking you can override default style using stylesheets (check readme).

markholland:
But with a date marking dot it still differs between platforms. The fix ensures that it looks the same on both Android & iOS and covers with or without marking dots.

**iOS without fix**
before

**iOS with fix**
after

**Android**
android

PR #391, opened 2018-03-20
Dynamic calendarHeight

tigo0:
## Description
This PR makes calendarHeight dynamic.

## What changed
calendarHeight is passed as props in the CalendarList. If no calendarHeight props is passed, it takes the default as it was => 360.

### DIFF
--- a/src/calendar-list/index.js
+++ b/src/calendar-list/index.js
@@ -12,6 +12,7 @@ import Calendar from '../calendar';
 import CalendarListItem from './item';
 
 const calendarHeight = 360;
+
 class CalendarList extends Component {
   static propTypes = {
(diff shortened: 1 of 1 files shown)


tautvilas:
Not sure how this worked for you because there is a bug. In constructor this should be done:

this.getItemLayout = this.getItemLayout.bind(this);

Without it calendars are not rendered properly in the list after scrolling

tigo0:
This was follow up from my other branch, Horizontal Calendar List where i bind the getItemLayout. I guess thats why I missed it. Thanks

johanntony:
i Use CalendarList ,with no set height props,when i page to next month,the height did not update ,how did i do.

PR #393, opened 2018-03-21
CalendarList accepts calendarHeight prop

bspaulding:
Trying to use a custom styled Calendar in CalendarList leads to problems with scroll calc if your calendar height is different than the default styles.


tautvilas:
this was solved via other PR, thanks

PR #394, opened 2018-03-21
Generic showLoadingIndicator

tigo0:
### Background
Whether to show `displayLoadingIndicator` depends on the state of the `markedDates` passed to the `Calendar` with some conditions. This is not always convenient and this PR gives better control on when to show the `indicator`.

### Description
Introduces `showLoadingIndicator` prop to the calendar. Pass bool to showLoadingIndicator to show/hide loader.

### What changed
Added new prop to Calendar that triggers loadIngIndicator next to the month name. Priority is given to `displayLoadingIndicator` if both are specified.


tautvilas:
Hello, could you update Readme with documentation about this prop?

tigo0:
@tautvilas ReadMe is updated. IMO the other way of showing the loading indicator is way too confusing and specific. Maybe you can consider deprecating/removing it so it won't confuse people. I believe it is in peoples` interest to have more control over it. just a thought )

src/calendar-list/index.js
```diff
@@ -49,7 +49,7 @@ class CalendarList extends Component {
     this.pastScrollRange = props.pastScrollRange === undefined ? 50 : props.pastScrollRange;
     this.futureScrollRange = props.futureScrollRange === undefined ? 50 : props.futureScrollRange;
     this.style = styleConstructor(props.theme);
-    this.calendarWidth = this.props.calendarWidth || width;
+    this.calendarWidth = props.calendarWidth;
```
Inbal-Tish: changes in this file are irrelevant.

src/calendar/index.js
```diff
@@ -246,13 +248,16 @@ class Calendar extends Component {
     }
     let indicator;
     const current = parseDate(this.props.current);
-    if (current) {
+    if (current && this.props.displayLoadingIndicator) {
```
Inbal-Tish: move this condition inside the `if (current)` and else it inside.

src/calendar/index.js
```diff
@@ -246,13 +248,16 @@ class Calendar extends Component {
...
       const lastMonthOfDay = current.clone().addMonths(1, true).setDate(1).addDays(-1).toString('yyyy-MM-dd');
       if (this.props.displayLoadingIndicator &&
           !(this.props.markedDates && this.props.markedDates[lastMonthOfDay])) {
         indicator = true;
       }
     }
+    else if (this.props.showLoadingIndicator) {
```
Inbal-Tish: no need for the if when you assign its value to the `indicator` variable

tigo0:
@Inbal-Tish Sorry, unfortunately I don't have the time to look into this now. This was related to a work I no longer connect with. If someone wants to take over and push this, please go ahead, otherwise I'd suggest closing this until someone else feels the need to re-implement. Thanks

Issue #307, opened 2018-01-16
Fixed day width for hidden days


skrywus:
https://github.com/wix/react-native-calendars/blob/3bd7ae49c6092e8eeaabfb56bb99bf1326a741f1/src/calendar/index.js#L157

Because of this fixed value I am not able to change day width which I need to change

PR #408, opened 2018-03-28
Move hardcoded width of hidden extra days to style

xxmuaddib:
The problem was that if we change the day width (in our example the width was being calculated according to the screen dimensions), hidden extra days stayed 32 and that caused misplace of days in rows that included hidden extra days.

Related to #307


tautvilas:
renamed the style name to dayContainer

PR #409, opened 2018-03-28
Single marking day component

xxmuaddib:
Single marking day component gives the opportunity to mark each day with custom styles - both container and text.

We needed to put days in circles, which could have:
- Background color
- Border
- Elevation (Shadow)
- Just text
- Rectangles
- etc.

What you can achieve with the component can be seen in the image attached.

marking5


tautvilas:
Thank you

tautvilas:
I renamed this marking type from 'single' to 'custom'

PR #411, opened 2018-03-28
CalendarList initialScrollIndex not working in RN 0.44 or Minor

ErickMaeda:
The flatlist has the parameter initialScrollIndex but in React Native 0.44 or minor doesn't work because it doesn't have this props.

So when ```componentDidMount()```it's scrolling to the middle of the CalendarList


tautvilas:
Sorry, will not accept PRs for old RN versions.

ErickMaeda:
Ok, I'll update the RN version.

PR #417, opened 2018-04-02
add remove clipped subviews option

gaboelnuevo:
Added removeClippedSubviews={} to CalendarList and Agenda.

Fix an issue with tab navigator of react-navigation passing removeClippedSubviews={false} to Agenda


tautvilas:
Hi, did you measure that there is performance increase if you add this option? i think there have been some problems with it in the past

gaboelnuevo:
No, because is only an option that I need to activate. The default behavior is the same

Issue #383, opened 2018-03-16
[Feature Request] Looking for free time view on Agenda


wmlutz:
## Description

Looking for a time-proportional view of agenda (i.e. gaps for unfilled space). I'd like my users to visually see when they have free time.

## Expected Behavior
An option with agenda view:
`time_proportional={true}`
that makes agenda items proportional to time, as well as showing the unused time between them. In Google Cal's app, similar to going from "Schedule" to "Day" views.

pinguinjkeke:
You can achieve this behaviour by filling items object with empty dates and Agenda example showing it.

```js
state = {
  items: {
    '2018-02-13': [{ text: 'event1' }, { text: 'event2' }],
    '2018-02-14': {}, // nothing happens here, empty date will be rendered
  }
};
```

wmlutz:
That works for empty days, I'm looking for something for empty time. I'm now realizing it would need to fundamentally change how Agenda view works more than I really thought. Events would need to have a start time and an end time. Although this would be a cool feature build in the future, I think maybe I should try to find another library.

![screenshot_20180316-174059](https://user-images.githubusercontent.com/6496048/37546624-0b78dd9c-2944-11e8-956b-1830624ab71d.png)

Luckygirlllll:
@wmlutz Is that  possible to set a time in the Agenda component right now?

wmlutz:
@Luckygirlllll Including a time in an 'item' object for the Agenda is do-able. But item objects are not standardized in any way so you can customize how an Agenda item is displayed with renderItem. Although this is similar to Agenda in a lot of ways, I think this may be it's own thing.

aprilmintacpineda:
I tweaked the code a little bit and came up with this (please disregard the styles, it's not really something that I'm concerned about as of the moment. And yeah, you can style this however you like.). I made this possible by providing the `previousItem` to the `renderDay` and `renderItem`. I honestly don't understand the need to have these two props. I think we can just do `renderItem` then provide the `currentItem` and `previousItem` to that prop which will return a react-native element, this will allow the developer using this library to render whatever they like on the agenda view.

<img width="346" alt="screen shot 2018-04-04 at 11 31 01 am" src="https://user-images.githubusercontent.com/21032419/38287220-17b15e38-37fc-11e8-8f3b-cd887ab52149.png">

Paul-Todd:
do you have a plan to merge this PR soon? It would be incredibly useful

aperdec:
[...]

PR #419, opened 2018-04-04
response to https://github.com/wix/react-native-calendars/issues/383

aprilmintacpineda:
#### The problem

- We need to be able to display free time on the agendas.
- `renderItem` and `renderDay` are identical and therefore redundant.

#### Solution

remove `renderDay` prop and use `renderItem` providing the `currentItem` and `previousItem` to the prop.

Doing so, will remove the redundancy and also gives freedom to the user to render out the content of the agenda.

###### renderItem signature

```
<Agenda
  // ... other props
  renderItem={(currentItem, previousItem) => {
    if (!previousItem) {
      // this is the first item on the list. Do something about it, or just leave it alone.
    }

    return (
      <View>
        <Text>time:</Text>
        <Text>Agenda!</Text>
      </View>
    );
  }}
/> 
```

screen shot 2018-04-04 at 12 42 24 pm

#### Breaking changes

- removed `renderDay`.
- `renderItem` receives `currentItem` as first parameter and `previousItem` as second parameter. Note that the `firstItem` is now gone.

#### Follow ups needed

- Update `README.md`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,8617 @@
+{
+  "name": "react-native-calendars",
+  "version": "1.17.7",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "absolute-path": {
+      "version": "0.0.0",
+      "resolved": "https://registry.npmjs.org/absolute-path/-/absolute-path-0.0.0.tgz",
+      "integrity": "sha1-p4di+9rftSl76ZsV01p4Wy8JW/c=",
+      "dev": true
+    },
+    "accepts": {
+      "version": "1.2.13",
+      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.2.13.tgz",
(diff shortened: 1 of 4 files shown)


aprilmintacpineda:
https://github.com/wix/react-native-calendars/issues/383

aprilmintacpineda:
#### Updates

give `nextItem` to `renderItem` as well.

###### renderItem signature

```js
<Agenda
  // ... other props
  renderItem={(currentItem, previousItem, nextItem) => {
    if (!previousItem) {
      // this is the first item on the list
    }

    if (!nextItem) {
      // this is the last item on the list
    }

    return (
      <View>
        <Text>time:</Text>
        <Text>Agenda!</Text>
      </View>
    );
  }}
/> 
```

tautvilas:
Hello, I don't see renderDay as reduntant. For people that want to have simple API both renderDay and renderItem are necessary. I suggest that we would add a new prop 'renderRow'. If this prop is provided then this function will be used instead of renderDay+renderItem

PR #421, opened 2018-04-05
Added basic loading spinner

mharrisweb:
Added `ActivityIndicator` instead of `text`


tautvilas:
Hi, I think  If user wants to provide custom placeholder content this should be done via prop.

Also animated spinner would degrade performance if it is being rendered in off-screen calendars

tautvilas:
closing due to inactivity

PR #424, opened 2018-04-06
Pass onScroll and scrollEventThrottle from Agenda to ReservationList

pinguinjkeke:
It's necessary feature when you working with animations depending on ReservationList.
It allows to:
- Pass custom ```onScroll``` callback function.
- Pass ```scrollEventThrottle```
- Pass ```Animated.Value``` that will change on FlatList's ScrollView y changes. It uses AnimatedFlatList and Animated.Event to make it performant, fast and smooth.

I.e. I need to hide header when scrolling ReservationList.

P.S. There is no roadmap for project. What do you think about internal refactoring like using class arrow functions instead of calling .bind(this) inside constructor, implementing defaultProps?


tautvilas:
Hi, I will think about this PR not sure if it should be merged. It is exposing agenda internals.

Regarding arrow functions - it is considered to be bad practice to have them in render function because in every render new function instance is created causing performance issues

pinguinjkeke:
@tautvilas no, I'm talking about using arrow functions inside class properties.

tautvilas:
sorry, this is not gonna be merged

PR #436, opened 2018-04-12
Fix: Clicking on an arrow to change months is difficult

GustasI:


tautvilas:
Thx @GustasI did you make sure hit box is not overlapping day elements?

PR #447, opened 2018-04-16
Add viewabilityConfig prop to CalendarList

tigo0:
### Description
`onMonthChanged` (on `ViewabilityItemsChanged`) is triggered only when the view is perfect fit. With `viewabilityConfig` we get more control on when to trigger `onMonthChanged` based on `viewabilityHelper.js` defined by `react-native`.

### What changed
Added `viewabilityConfig` prop to `CalendarList` that takes an object defined by `viewabilityHelper.js`

### DIFF
--- a/src/calendar-list/index.js
+++ b/src/calendar-list/index.js
@@ -49,8 +49,9 @@ class CalendarList extends Component {
     this.pastScrollRange = props.pastScrollRange === undefined ? 50 : props.pastScrollRange;
     this.futureScrollRange = props.futureScrollRange === undefined ? 50 : props.futureScrollRange;
     this.style = styleConstructor(props.theme);
-    this.calendarWidth = this.props.calendarWidth || width;
+    this.calendarWidth = props.calendarWidth;
     this.calendarHeight = props.calendarHeight;
+    this.viewabilityConfig = props.viewabilityConfig;
 
     const rows = [];
     const texts = [];
@@ -200,13 +201,18 @@ class CalendarList extends Component {
         initialScrollIndex={this.state.openDate ? this.getMonthIndex(this.state.openDate) : false}
         getItemLayout={this.getItemLayout}
         scrollsToTop={this.props.scrollsToTop !== undefined ? this.props.scrollsToTop : false}
+        viewabilityConfig={this.viewabilityConfig}
       />
     );
   }
 }
 
 CalendarList.defaultProps = {
-  calendarHeight: 360
+  calendarHeight: 360,
+  calendarWidth: width,
+  viewabilityConfig: {
+    viewAreaCoveragePercentThreshold: 100,
+  }
 };
 
 export default CalendarList;
--- a/README.md
+++ b/README.md
@@ -331,6 +331,8 @@ If you implement an awesome day component please make a PR so that other people 
   scrollEnabled={true}
   // Enable or disable vertical scroll indicator. Default = false
   showScrollIndicator={true}
+  // Set viewabilityConfig defined by viewabilityHelper.js. Default: viewAreaCoveragePercentThreshold=100
+  viewabilityConfig={viewabilityConfigObject}
   ...calendarParams
 />
 ```


tautvilas:
is there some specific bug that this helps to solve?

tigo0:
Yes. Currently `onViewableItemsChanged` is triggered when there is a perfect fit of the vieawble item in the window. This is valid for `HorizontalCalendar` with `pagingEnabled` as well. Unfortuantely, some phones (tested on Nexus 5x) never assume perfect fit on scrolling and fitting the page in the window - so far my investigation showed that this is due to miscalculation (maybe?) in `react-native`.
 `React-native` outputs window width of `411.428571429` (phone width / density, e.g. 1080 / 2.625) on that specific phone (Its been reported that it doesnt work on galaxy s8+ as well) and while scrolling with `pagingEnabled` the RN calculations are off with ~`0.0021...` in its offset. So its never a perfect fit with the `pagingEnabled`.
 I managed to work around it by passing `viewabilityConfig` with `itemVisiblePercentThreshold: 50`.

src/calendar-list/index.js
```diff
@@ -200,13 +201,18 @@ class CalendarList extends Component {
...
     );
   }
 }
 
 CalendarList.defaultProps = {
-  calendarHeight: 360
+  calendarHeight: 360,
+  calendarWidth: width,
+  viewabilityConfig: {
+    viewAreaCoveragePercentThreshold: 100,
```
chrismllr: So this would mean the list item would have to take up 100% of the viewport - Wouldnt we want something more along the lines of

```
itemVisiblePercentThreshold: 25
```
Which would mean the FlatList item would determine it is in view after 25% of it is visible?

I may be misinterpreting viewabilityConfig, but wanted to ensure the right default prop here

tigo0: I left the default prop to 100, since up until now it was taking the full viewport, and left the opportunity for developers to decide whether they want to change that. Furthermore, I didn't spot that issue with CalendarList's without the pagination enabled. (`pagingEnabled` is valid prop only for horizontal lists), therefore I decided to leave it as it was with the opportunity to change on the fly. Also the behaviour of firing up the event is different with the threshold, in my case it was not fitting having 25, therefore the suggestion for 50 in my comment. But as I said, its up to the developer to decide if/how to use it. Open for changes if requested tho. )

chrismllr: Sounds good!

sutaka:
+1
I have the same issue.

alburkerk2:
Any update on this ? 

I wanted to create the same PR. 

It's been over a year now since this one was created, can I help in anything to make this one merged soon ?

src/calendar-list/index.js
Inbal-Tish: I think it's best to remove this default so it will be passed only if the user pass it, otherwise, we leave the RN defaults.

src/calendar-list/index.js
```diff
@@ -49,8 +49,9 @@ class CalendarList extends Component {
     this.pastScrollRange = props.pastScrollRange === undefined ? 50 : props.pastScrollRange;
     this.futureScrollRange = props.futureScrollRange === undefined ? 50 : props.futureScrollRange;
     this.style = styleConstructor(props.theme);
-    this.calendarWidth = this.props.calendarWidth || width;
+    this.calendarWidth = props.calendarWidth;
```
Inbal-Tish: This is unnecessary

Inbal-Tish:
@tigo0 Is this issue still relevant after RN upgrade to 0.61.4? Also, please resolve conflicts. Thanks

tigo0:
@Inbal-Tish Unfortunately, this was opened almost 2 years ago and I no longer maintain the app related to this issue, and I am not aware if it persist. I don't really have the time to re-investigate. If someone can take over will be nice, otherwise I'd suggest closing until further investigation. Thanks

Inbal-Tish:
@tigo0 Thank you for your replay. I'll close this PR and hope that this issue has been resolved or that someone will open a new one if necessary. Thanks!

PR #449, opened 2018-04-17
😈 Fix initialListSize calculate error

iamkun:
It seems the list size here should be the sum of this.pastScrollRange and this.futureScrollRange rather than product.


tautvilas:
nice catch :)

PR #450, opened 2018-04-17
😈 Fix remove unused state

iamkun:
Seems the state `initialized ` has never been used. Shall we remove it for less confusion?


tautvilas:
thanks :)

PR #461, opened 2018-04-24
Allow to show arrows in CalendarList

christophehurpeau:
Keep the same default to `true`, but allow to show them when `<CalendarList hideArrows={false} />`


tautvilas:
thanks

PR #462, opened 2018-04-24
Fix scrollToDay and scrollToMonth in CalendarList when horizontal

christophehurpeau:


tautvilas:
@tigo0  plz review

tautvilas:
@christophehurpeau could you explain what was the bug? thanks

christophehurpeau:
With the horizontal mode enabled, without this the height is used to calculate the scroll offset, so unless the height is the same than the width it doesn't work.
Let's say you have a width of 500 and a height of 400. You are on month 1 and want to scroll to month 2: without the fix, the offset is set to 400 and you will end up beetween the month 1 and 2

tautvilas:
thanks

PR #467, opened 2018-04-26
Added container style prop to CalendarList

LenaZhuk:
These styles will be applied to the scroll view content container.

### DIFF
--- a/src/calendar-list/index.js
+++ b/src/calendar-list/index.js
@@ -8,7 +8,7 @@ import XDate from 'xdate';
 import {xdateToData, parseDate} from '../interface';
 import styleConstructor from './style';
 import dateutils from '../dateutils';
-import Calendar from '../calendar';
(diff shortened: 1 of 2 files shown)


LenaZhuk:
Need this style because basic style not allowed padding. For this purpose in RN contentContainerStyle was added to FlatList.

screenshot at 17 5-59-15

Inbal-Tish:
@LenaZhuk Hi. Please check if 'calendarWidth' and the recently added 'calendarStyle' can resolve you need for paddings. I prefer to add fewer props if there are others that can be used.

kchasiotis:
@Inbal-Tish `calendarWidth` and `calendarStyle` both refer to the `CalendarListItem` component. Those attributes do not cover the use case of the current pull request. This pull request attempts to set a style in the whole container, not the item.

Issue #485, opened 2018-05-15
bug: Agenda header is scrolling to wrong position when minimized


paolord:
## Description

Supplied Agenda view with selected date prop and the header is scrolled to either the wrong week or an empty space between months in the calendarlist

## Expected Behavior

In the default/minimized state, the header should be scrolled to the correct week containing either the current date or selected date

## Observed Behavior

It's displaying either an empty space or the wrong week

![bug](https://user-images.githubusercontent.com/1720393/40046380-727bcbc2-585f-11e8-88e2-2910d4617a91.png)

possibly similar to https://github.com/wix/react-native-calendars/issues/480

Inbal-Tish:
@paolord Is this issue still relevant?

MuhammedRafi:
@Inbal-Tish Yes ,this is still there

Inbal-Tish:
Please see comment on PR #486

PR #486, opened 2018-05-15
fix: Agenda header is scrolling to wrong position when minimized #485

paolord:
I was using version 1.17.0 before upgrading to 1.19.3, compared the agenda view code and tested  returning the scrollToDay function call in the onLayout function. Fixed it on my implementation.

### DIFF
--- a/src/agenda/index.js
+++ b/src/agenda/index.js
@@ -135,6 +135,7 @@ export default class AgendaView extends Component {
   onLayout(event) {
     this.viewHeight = event.nativeEvent.layout.height;
     this.viewWidth = event.nativeEvent.layout.width;
+    this.calendar.scrollToDay(this.state.selectedDay.clone(), this.calendarOffset(), false);
     this.forceUpdate();
   }
 


tautvilas:
Is the fix needed in newest version?

tautvilas:
Is there some issue ticket related to this fix?

paolord:
Issue #485, I referenced it in the title.

Inbal-Tish:
@paolord Hi. Is this issue still relevant? I can't reproduce it. Can you send a snippet where I can see the issue?

narender2031:
@paolord after applying these changes. Still not showing dates for last of the scroll. 
Screenshot 2020-04-21 at 1 08 52 PM

rdewolff:
This bug is still present.

paolord:
I made this change around 2 years ago. When I used to have react-native projects. The code around this fix has probably changed a lot. I'll have to test this again with the latest versions.

Villar74:
this not fixing the problem if pastScrollRange and futureScrollRange is 1

PR #491, opened 2018-05-16
Agenda - Add onWeekChange event

AdrianKujawski:
Hi,

I added 2 things:
- props firstDay
- event onWeekChange

Event fires when week changes while scrolling agenda list. It return first and last day of visible week.


tautvilas:
Hi, i added firstDay to props, I think onWeekChange is a bit excessive, client can implement this logic on his end if he needs it inside onDayChange

PR #504, opened 2018-05-28
feat(calendar/header): add custom weekdays rendering component

x86kernel:
Add custom weekdays rendering component


tautvilas:
Hi currently calendars API does not make user of components passed directly. I suggest using simple View for rendering and allowing to override style with stylesheet overrides or custom theme props (the same way styling can be customized for other component)

PR #530, opened 2018-06-21
feat(calendar/header): add disableArrowLeft/disableArrowRight props

lfkwtz:
Add `disableArrowLeft` and `disableArrowRight` props to Calendar header component. Also added a `disabledArrowColor` prop to the theme object


tautvilas:
Wouldn't it be better just to user renderArrow to implement custom logic for disabling? I'm afraid this will make API too complicated

lfkwtz:
I see it the other way personally - I see disabling an arrow as a baseline feature, and creating a custom component to conditionally trigger that logic seems like overkill.

src/calendar/header/index.js
```diff
@@ -63,16 +65,16 @@ class CalendarHeader extends Component {
   }
 
   onPressLeft() {
-    const {onPressArrowLeft} = this.props;
-    if(typeof onPressArrowLeft === 'function') {
+    const { onPressArrowLeft } = this.props;
```
Inbal-Tish: Please remove all formatting changes

Inbal-Tish:
@lfkwtz I agree that disabling the arrows should be a feature. Please remove all formatting changes so I can review the relevant ones and please make sure the disable functionality can dynamically be updated. Also, resolve conflict, please. Thanks

lfkwtz:
cal

@Inbal-Tish should be good to go now

PR #546, opened 2018-07-04
Add custom header

jfernandogt:
This PR allow to use a custom header component and set a specific month. 

This is an example that you can create using this features:

### DIFF
--- a/src/calendar/index.js
+++ b/src/calendar/index.js
@@ -41,7 +41,7 @@ class Calendar extends Component {
     // If firstDay=1 week starts from Monday. Note that dayNames and dayNamesShort should still start from Sunday.
     firstDay: PropTypes.number,
 
-    // Date marking style [simple/period/multi-dot/multi-period]. Default = 'simple' 
+    // Date marking style [simple/period/multi-dot/multi-period]. Default = 'simple'
     markingType: PropTypes.string,
 
     // Hide month navigation arrows. Default = false
@@ -93,6 +93,7 @@ class Calendar extends Component {
 
     this.updateMonth = this.updateMonth.bind(this);
     this.addMonth = this.addMonth.bind(this);
+    this.setMonth = this.setMonth.bind(this);
     this.pressDay = this.pressDay.bind(this);
     this.longPressDay = this.longPressDay.bind(this);
     this.shouldComponentUpdate = shouldComponentUpdate;
@@ -151,6 +152,10 @@ class Calendar extends Component {
 
   addMonth(count) {
     this.updateMonth(this.state.currentMonth.clone().addMonths(count, true));
+  }
+
+  setMonth(month) {
+    this.updateMonth(this.state.currentMonth.clone().setMonth(month, true));
   }
 
   renderDay(day, id) {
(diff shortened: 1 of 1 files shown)


josema88:
+1

tautvilas:
wouldn't it make more sense to have 'hideHeader' prop and you could render whatever you want on top?

jfernandogt:
I don't think so, because I need to set the month and know which is the current month (in the same locale). I think is more complicate use your idea.

gzepeda:
+1

tautvilas:
Well you could know what is month by listening onDayChange event from calendar. And you can set the month using 'current' prop of calendar.

lecong:
Thanks for source code.

But, this changed do note have on master branch.
Why?

jfernandogt:
@tautvilas 
If I use your idea I need to create a logic to handle the state (for the current month) and create a function to listen when the day change, I think is better that the component handle it on its own and only render the JSX with the new style.

Zapopozapopo:
Hi, I like both approaches "hideHeader" or "Custom Header". Should I wait for this functionality soon?

apostopher:
@jfernandogt @tautvilas I like the `custom header` approach. We already have `renderArrow` and `dayComponent`. So adding `renderHeader` seems right to me.
Also renaming the `dayComponent` to `renderDay` will add consistancy. what do you think?

jfernandogt:
@apostopher I think is better to change the dayComponent to renderDay. I hope that @tautvilas accept this PR.

src/calendar/index.js
```diff
@@ -258,20 +263,41 @@ class Calendar extends Component {
...
-          monthFormat={this.props.monthFormat}
-          hideDayNames={this.props.hideDayNames}
-          weekNumbers={this.props.showWeekNumbers}
-          onPressArrowLeft={this.props.onPressArrowLeft}
-          onPressArrowRight={this.props.onPressArrowRight}
-        />
+        {
+          this.props.renderHeader
+          ? this.props.renderHeader({
+            theme: this.props.theme,
```
Inbal-Tish: Please move all props to an object and use it on both instead of this massive duplicated code

src/calendar/index.js
```diff
@@ -258,20 +263,41 @@ class Calendar extends Component {
...
+            weekNumbers: this.props.showWeekNumbers,
+            onPressArrowLeft: this.props.onPressArrowLeft,
+            onPressArrowRight: this.props.onPressArrowRight,
+          })
+          : <CalendarHeader
+            theme={this.props.theme}
+            hideArrows={this.props.hideArrows}
+            month={this.state.currentMonth}
+            addMonth={this.addMonth}
+            setMonth={this.setMonth}
```
Inbal-Tish: What is the CalendarHeader is doing with the `setMonth` prop?

Inbal-Tish:
@jfernandogt Can you please add an example in the Calendars screen so I can see how you use it?

Inbal-Tish:
Please see PR #1225

Yandamuri:
@jfernandogt I am using `renderHeader` prop to display custom header. This function receives `date` as parameter as per documentation. Problem here is multiple dates are being passed  as shown below when the calendar is first time rendered. 

      Wed, 24 Feb 2021 00:00:00 GMT
      Wed, 24 Mar 2021 00:00:00 GMT
      Sun, 24 Jan 2021 00:00:00 GMT
(Assume I open calendar on 24 Feb 2021)
From next time onwards when ever month is changed the date parameter passed to the function is always one month ahead to the current month like follows

    Sat, 24 Apr 2021 00:00:00 GMT (in this case I have changed calendar to next month that is Marh)

Can you please tell me why wrong date is being passed to custom header function?

Issue #401, opened 2018-03-25
Very small tappable area for right and left calendar arrows making it difficult to change months in iPhone


feliperaul:
Please make our job easier by filling this template out to completion. If you're requesting a feature instead of reporting a bug, please feel free to skip the Environment and Reproducible Demo sections.

## Description

Very small clickable area in the calendar right and left arrows, making it difficult to change months in iPhone. 

## Expected Behavior

Using a real device (iPhone), sometimes it takes 4-5 taps to actually hit the default arrows to change the month. 

I know the library supports a custom component for the arrows, but I didn't expect having to write one if I like the default, finding this to be a bug instead of something that I needed to customize. 

## Environment

npm ls react-native
└── react-native-calendars@1.17.4 

npm ls react-native
└── react-native@0.52.0

Also specify:

1. Phone/emulator/simulator & version: iPhone X

theohdv:
Just had a look in the code and it could be solved by adding `hitSlop` prop to the `Touchable` component (https://github.com/wix/react-native-calendars/blob/master/src/calendar/header/index.js#L79). What do you think ?

tautvilas:
solved in 1.18.2

feliperaul:
@theohdv @tautvilas thanks!

PR #551, opened 2018-07-09
Added capability to style right/left calendar arrows separately

masako-yamada:
Added styles for right/left calendar arrows that are applied in addition to the current "arrow" and "arrowImage" styles.

example usage:
```
<Calendar 
    theme={
            // for container View
            arrow: {
                // write common styles here or you can skip this entry if you inherit the default style
            },
            arrowLeft: {
                position: 'relative',
                top: 0,
                left: 0
            },
            arrowRight: {
                position: 'relative',
                top: 0,
                right: 0
            },
            // for arrow Image
            arrowImage: {
                // write common styles here or you can skip this entry if you inherit the default style
            },
            arrowImageLeft: {
                ...Platform.select({
                    ios: {
                        tintColor: '#ff0000'
                    },
                    android: {
                        tintColor: '#ff0000'
                    }
                })
            },
            arrowImageRight: {
                ...Platform.select({
                    ios: {
                        tintColor: '#00ff00'
                    },
                    android: {
                        tintColor: '#00ff00'
                    }
                })
            }
        }
    }
/>
```

### DIFF
--- a/src/calendar/day/custom/index.js
+++ b/src/calendar/day/custom/index.js
@@ -52,6 +52,7 @@ class Day extends Component {
 
     if (marking.selected) {
       containerStyle.push(this.style.selected);
+      textStyle.push(this.style.selectedText);
     } else if (isDisabled) {
       textStyle.push(this.style.disabledText);
     } else if (this.props.state === 'today') {
--- a/src/calendar/header/index.js
+++ b/src/calendar/header/index.js
@@ -82,28 +82,28 @@ class CalendarHeader extends Component {
       leftArrow = (
         <TouchableOpacity
           onPress={this.onPressLeft}
-          style={this.style.arrow}
+          style={[this.style.arrow, this.style.arrowLeft]}
           hitSlop={{left: 20, right: 20, top: 20, bottom: 20}}
         >
           {this.props.renderArrow
             ? this.props.renderArrow('left')
             : <Image
                 source={require('../img/previous.png')}
-                style={this.style.arrowImage}
+                style={[this.style.arrowImage, this.style.arrowImageLeft]}
               />}
         </TouchableOpacity>
       );
       rightArrow = (
         <TouchableOpacity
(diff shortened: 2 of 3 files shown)


tautvilas:
what about just using renderArrow property?

masako-yamada:
Hi, sorry for the late reply, I was on holiday.

Yes, I could use renderArrow property to customize the arrow view itself, however, I couldn't change the position of arrows since the arrow view is inside of TouchableOpacity which I need to style for the position.

As well, I think this change would be useful when renderArrow is overkill but need to give small different styles to right and left arrows. 😄

gmpetrov:
Hi, looks like this branch has not been merged.
@masako-yamada  do you know why ? 

Would be nice to have this feature 😀

masako-yamada:
@gmpetrov I guess probably moderators judged `renderArrow` serves enough for most cases. 
Actually, I needed this feature for my edge case (I needed to change the position of arrows).

gmpetrov:
@masako-yamada When customizing arrows's positions with renderArrow, it gets buggy, in part because of TouchableOpacity of each arrows are overlapping.

coach

masako-yamada:
I guess you need to modify `hitSlop` of TouchableOpacity and it can't be handled by style.

gmpetrov:
Currently the hitSlop property of TouchableOpacity is hardcoded, and there is no possible way to override its default value.

masako-yamada:
Here is a similar discussion, I hope he could make time for the fix 😄 
https://github.com/wix/react-native-calendars/issues/401#issuecomment-439547491

FelipeGatti:
I'm with the same exact issue, is there any workarounds for this "hitSlop" bug untill it gets fixed ?

masako-yamada:
@FelipeGatti I think no other way except to fork the repository and change the hardcoded hitSlop to your favorable value.

PR #553, opened 2018-07-09
feat(agenda) added the ability to display day as a column instead of row

cloverForbes:


chasemac:
Why hasn't this been merged yet?

src/agenda/index.js
```diff
@@ -47,6 +47,8 @@ export default class AgendaView extends Component {
     renderItem: PropTypes.func,
     // specify how each date should be rendered. day can be undefined if the item is not first in that day.
     renderDay: PropTypes.func,
+    // determine if day should be displayed above or beside item
+    dayColumn: PropTypes.bool,
```
Inbal-Tish: I don't like this prop name, it's confusing. How about changing it to something like 'columDisplay' or 'columnDayDisplay'?

magrinj:
@the-transgineer @Inbal-Tish It's actually already possible by passing the prop `theme` to the Agenda with the advanced style option like this:
```
  'stylesheet.agenda.list': {
    container: {
      flexDirection: 'column',
    },
  },
```
Adding props for everything like some style options isn't really the right choice in my opinion.

PR #554, opened 2018-07-09
Replace wrapping View with new-style React.Fragment

mikelovesrobots:
I'm not sure why, but this change eliminates animation glitches for us when we wrap the calendar with `react-native-collapsible`. `react-native-collapsible` is just a `View` with `overflow:hidden` and an animated `height` style property, so I don't know why this change would do anything but it definitely does. I feel like it's indirectly helping out the react internal Yoga layout engine somehow.

Anyway, maybe you'll merge it because React.Fragment is the new way to do views that only exist to return multiple components.

https://reactjs.org/docs/fragments.html

Thanks for looking at this!

- Mike

### DIFF
--- a/src/agenda/index.js
+++ b/src/agenda/index.js
@@ -1,20 +1,14 @@
-import React, {Component} from 'react';
-import {
-  Text,
-  View,
-  Dimensions,
-  Animated,
-  ViewPropTypes,
-} from 'react-native';
-import PropTypes from 'prop-types';
-import XDate from 'xdate';
+import React, { Component } from "react";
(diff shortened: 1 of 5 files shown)


Inbal-Tish:
@mikelovesrobots  Please remove all formatting changes so I can review relevant changes. Thanks

Inbal-Tish:
Also, you have conflicts and many lint errors from your formatting.

mikelovesrobots:
This is over a year old. Of course there’s merge conflicts and formatting issues when merged with current code.

You own it now. I don’t feel anything toward that PR.

Issue #259, opened 2017-12-04
Agenda scroll up doesn't roll back to previous week


Rc85:
When I scroll my Agenda down, it will roll over to the next week when you hit a date that is in that week. But then when I try to scroll up, it doesn't roll back to the previous week and any previous week events are not visible anymore.

Have to go into Calendar List and select the date in previous week.

ItsNoHax:
This is the reason I didn't start using this library a couple months ago, pity this isn't fixed yet. It might seem small but it complete ruins the Agenda UX feeling

tautvilas:
@Rc85 could you add a gif/video?

ItsNoHax:
It's not really a bug? But it's a crucial feature for Agenda, click on any date and try to scroll back to a day before, it wont allow you.

bertucci13:
this feature is really important for us. Please fixed asap.

abeltje1:
I'm also experiencing this problem and I've made a GIF:
![react native calendar scroll up bug](https://user-images.githubusercontent.com/11901926/34723359-8e487c74-f54a-11e7-8acb-b45189f63949.gif)

geniuscd:
Yes this is crucial feature, its not actually a bug.
maybe some workaround will be to scroll to the selected date rather than redrawing the agenda.
BTW how to scroll to specific date?

slorber:
@ItsNoHax @bertucci13 open source does not work like that. If you need a critical feature, do your job and contribute with a PR, start your own fork, create your own lib or implement this feature yourself. People don't work for you for free.

abeltje1:
@slorber You're right, but it's a little strange that the authors of this library do not acknowledge this issue anywhere - it seems right now like the Agenda component is simply unfinished, and I do not understand why that is not clearly communicated on the GH page. If it'd be clearly communicated there, maybe we can start a joint project to implement this feature. That's the reason @ItsNoHax  and @bertucci13 are not so satisfied, I assume. 

Bottom line: open source definitely is made by the community, but the authors of a library should also help the community by stating clearly which essential features are currently missing. 

Let's do that! Does everyone agree that this is an essential feature of every agenda?

tautvilas:
[...]

PR #563, opened 2018-07-13
Always display all items in Agenda component to allow scrolling up

Reidweb1:
Added a potential fix for this issue - https://github.com/wix/react-native-calendars/issues/259

This may be against the intended behavior of the Agenda Component but it fixes the issue of not being able to scroll back up to previous calendar items by keeping all of the reservations around at all times.

### DIFF
--- a/src/agenda/index.js
+++ b/src/agenda/index.js
@@ -66,6 +66,8 @@ export default class AgendaView extends Component {
     selected: PropTypes.any,
     // earliest possible day
     earliest: PropTypes.any,
+    // latest possible day
+    latest: PropTypes.any,
     // Minimum date that can be selected, dates before minDate will be grayed out. Default = undefined
     minDate: PropTypes.any,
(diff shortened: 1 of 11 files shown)


RafaelKr:
Thanks for this PR, I need this soon. I didn't test it yet, but after looking into the code I'm asking myself why you did change colors and fonts? Was this by accident as you need this for your own project?

I think for this PR it would be better to keep the defaults.

vjfuenzalida:
Is this working?

faustinobsd:
Hi @Reidweb1. I've tried your solution, but when I try to run the code, it gives me the Error: 'reservations' is read-only. 

I'm using latest version of RN and Expo. 

Thank you.

Inbal-Tish:
@Reidweb1 I don't see the issue in Agenda that you're trying to fix. Also, this is a dramatic change for a minor problem and it's not proven to work properly. Sorry, but I cannot accept it.

PR #570, opened 2018-07-23
Fix agenda wrong render when orientation changes.

jper92:
The calendar list rendered by the agenda has the initial width as the default prop,
on orientation change, a wrong width is used to render the calendar list, the fix
explicitly adds the calendarWidth prop based on agenda's `this.viewWidth`

Examples of wrong renders that are solved with this:
img_0590
img_0588
img_0592
img_0591


tautvilas:
thanks!

PR #601, opened 2018-08-21
use custom style in markingType 'multi-dot'

toonPt0473:
add custom style

### DIFF
--- a/src/calendar/day/multi-dot/index.js
+++ b/src/calendar/day/multi-dot/index.js
@@ -76,6 +76,18 @@ class Day extends Component {
       containerStyle.push(this.style.today);
       textStyle.push(this.style.todayText);
     }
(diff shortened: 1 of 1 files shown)


httpeace:
+1

ccau1:
+1

Inbal-Tish:
@toonPt0473 Hi. What is the issue you want to solve here? Why not use `markingType={'custom'}`?

Thierno2sow:
hey @Inbal-Tish quick question on the  CalendarProvider is there a way to position the todayButton on the right bottom corner of the screen and not the left (react native)

Inbal-Tish:
@Thierno2sow Use `todayButtonStyle`.

Thierno2sow:
@Inbal-Tish thanks would something like this work 
    position: "absolute",
   
    right: 20,
    bottom: 0

PR #605, opened 2018-08-23
Adds initialNumToRender in calendar list for faster rendering

### LINKED ISSUE #598: Calendar opens slower in android
## Description
Calendar is opening really slowly in android. We're resorting to wrapping it in a view with an `onLayout` handler and showing a loader for the calendar which is giving a really bad experience for our users

## Expected Behavior
Calendar should open without any delay 

## Observed Behavior
Calendar takes 1-2 seconds to open. This only happens on android devices

## Environment

└── react-native-calendars@1.19.4
* `npm ls react-native`:
└── react-native@0.55.4

Also specify:

Simulator - Nexus 5X

## Reproducible Demo
```js
<CalendarList
    disableMonthChange={true}
    pastScrollRange={0}
    futureScrollRange={6}
    scrollEnabled={false}
    horizontal={true}
    pagingEnabled={true}
    showScrollIndicator={true}
    markedDates={...}
    onDayPress={x => x}
    onDayLongPress={x => x}
  />
```

To show the loader we can wrap it in 
```js
<View onLayout={e => this.setState({ layout: e.nativeEvent.layout })}>
    {this.state.layout ? <CalendarList {...params} /> : <ActivityIndicator />}
</View>
```

debugpai:
Fixes #598

### DIFF
--- a/src/calendar-list/index.js
+++ b/src/calendar-list/index.js
@@ -219,6 +219,7 @@ class CalendarList extends Component {
         initialScrollIndex={this.state.openDate ? this.getMonthIndex(this.state.openDate) : false}
         getItemLayout={this.getItemLayout}
         scrollsToTop={this.props.scrollsToTop}
+        initialNumToRender={1}
       />
     );
   }


src/calendar-list/index.js
```diff
@@ -219,6 +219,7 @@ class CalendarList extends Component {
         initialScrollIndex={this.state.openDate ? this.getMonthIndex(this.state.openDate) : false}
         getItemLayout={this.getItemLayout}
         scrollsToTop={this.props.scrollsToTop}
+        initialNumToRender={1}
```
sibelius: can this be a prop?

debugpai: What is the usecase when `initialNumToRender` should not be 1? As far as I understand, the property is there when we are showing multiple items in the list on the page. And since in the calendar, the user can only see one month at a time, I think it should always be 1. What do you think?

tautvilas:
1. I believe you are using some navigation solution that bugs calendars. There is already optimization inside `onViewableItemsChanged` that renders placeholder 'calendars' instead of real ones if they are not visible. Please check the example project and see if it also gives you slow performance
2. initialNumToRender PR can be considered but it should be a parameter. initialNumToRender=1 means only one calendar in calendar list will be rendered. It should be not 1, but amount of calendars that fill the screen. with initialNumToRender=1 when using example project I can see that there is 'flicker'. First the first month calendar in the list is rendered then the other months appear

src/calendar-list/index.js
NickyYo: Not all time, my apps renders it more than once 
screen shot 2018-10-04 at 18 11 30

Inbal-Tish:
@debugpai Please see PR #719 for a better solution.

Issue #505, opened 2018-05-28
Add renderArrow to CalendarList


g00dnatur3:
Somebody please add this line of code

`renderArrow={this.props.renderArrow === undefined ? undefined : this.props.renderArrow}`

to file

`https://github.com/wix/react-native-calendars/blob/master/src/calendar-list/item.js`

at line number `27`

Thank You Very Much!

g00dnatur3:
Actually please add these lines

```renderArrow={this.props.renderArrow === undefined ? undefined : this.props.renderArrow}
          onPressArrowLeft={this.props.onPressArrowLeft === undefined ? undefined : this.props.onPressArrowLeft}
          onPressArrowRight={this.props.onPressArrowRight === undefined ? undefined : this.props.onPressArrowRight}```

This should do the trick and allow full customization of the left and right arrows in . the calendar list.

Many thanks for the developers for making this code well organized and easy to read.

Cheers!

olfek:
@Inbal-Tish This issue can be closed.

mensurlatifi:
How can I add arrows to Calendar List
@g00dnatur3

zagoa:
Now you can use the props `renderArrow={true}`.

mensurlatifi:
@zagoa  
    "react-native-calendars": "^1.213.0",
I have this version installed and in the CalendarList when I call renderArrow={true} it throws a warning saying it expects a function at renderArrow, not a boolean value.

huynhnhatthiep:
Now you can use renderArrow with syntax as below:
renderArrow={(direction) => direction === 'left' ? (<IconLeft>) : (< IconRight/>)}

PR #609, opened 2018-08-29
Add arrows to CalendarList #505

zagoa:
Thanks to the issue #505 I implemented and tested the code.

### DIFF
--- a/src/calendar-list/item.js
+++ b/src/calendar-list/item.js
@@ -44,6 +44,9 @@ class CalendarListItem extends Component {
           dayComponent={this.props.dayComponent}
           disabledByDefault={this.props.disabledByDefault}
           showWeekNumbers={this.props.showWeekNumbers}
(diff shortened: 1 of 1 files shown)


AdamGold:
Any news on this?

src/calendar-list/item.js
```diff
@@ -44,6 +44,9 @@ class CalendarListItem extends Component {
           dayComponent={this.props.dayComponent}
           disabledByDefault={this.props.disabledByDefault}
           showWeekNumbers={this.props.showWeekNumbers}
+          renderArrow={this.props.renderArrow === undefined ? undefined : this.props.renderArrow}
```
Inbal-Tish: The check for undefined is redundant. Just pass the prop as is

zagoa: Of course !

PR #615, opened 2018-09-03
Update index.js

Xupet:
Fix knob bug for Android

### DIFF
--- a/src/agenda/index.js
+++ b/src/agenda/index.js
@@ -16,7 +16,7 @@ import ReservationsList from './reservation-list';
 import styleConstructor from './style';
 import { VelocityTracker } from '../input';
 
-const HEADER_HEIGHT = 104;
+const HEADER_HEIGHT = 108;
 const KNOB_HEIGHT = 24;
 
 //Fallback when RN version is < 0.44


Xupet:
Fix error that made the knob on the Agenda that got cut because of the height of the container. Increased the height by 4pts and seems like it works better now.

src/agenda/index.js
```diff
@@ -16,7 +16,7 @@ import ReservationsList from './reservation-list';
 import styleConstructor from './style';
 import { VelocityTracker } from '../input';
 
-const HEADER_HEIGHT = 104;
+const HEADER_HEIGHT = 108;
```
sibelius: can't this be a prop instead?

tautvilas: probably at lerast this should be different for different platform? If IOS is fine with 104

carlosroberto555: Use like this

```jsx
import { Platform } from 'react-native'

const HEADER_HEIGHT = Platform.OS === 'ios' ? 104 : 108
```

carlosroberto555: Use Platform api to get specific value

HaswinVidanage: The issue is still there for much smaller screens (tested with 3.4")
Created the following PR as a fix https://github.com/wix/react-native-calendars/pull/953

Inbal-Tish:
@Xupet Please see PR #953

PR #620, opened 2018-09-07
Add keyboardShouldPersistTaps prop

enahum:
This PR adds the `keyboardShouldPersistTaps` prop to the CalendarList **FlatList** component being used at https://github.com/mattermost/mattermost-mobile


Inbal-Tish:
@enahum Hi. Can you describe a use-case where you need to change the 'never' default?

enahum:
@Inbal-Tish, when having an input box and using the calendar as an autocomplete option for dates.

See https://github.com/mattermost/mattermost-mobile

Inbal-Tish:
@enahum Can you please resolve conflict? Thanks

enahum:
@Inbal-Tish done

src/calendar-list/index.js
```diff
@@ -44,6 +44,10 @@ class CalendarList extends Component {
     calendarWidth: PropTypes.number,
     /** Dynamic calendar height */
     calendarHeight: PropTypes.number,
+
+    // Should Keyboard persist taps
+    keyboardShouldPersistTaps: PropTypes.oneOf(['never', 'always', 'handled']),
```
Inbal-Tish: defining the propType as `FlatList.propTypes. keyboardShouldPersistTaps` is clearer.

src/calendar-list/index.js
```diff
@@ -62,6 +66,7 @@ class CalendarList extends Component {
     scrollEnabled: true,
     scrollsToTop: false,
     removeClippedSubviews: Platform.OS === 'android' ? false : true,
+    keyboardShouldPersistTaps: 'never',
```
Inbal-Tish: This is the FlatList default value so no need to set it (if the prop will be passed as undefined this will be its value).

enahum:
@Inbal-Tish updated with your suggestions.

PR #626, opened 2018-09-10
fix(bug): get pastScrollRange and futureScrollRange from props, fix #625

### LINKED ISSUE #625: [master] initialListSize is wrong
`initialListSize={this.pastScrollRange + this.futureScrollRange + 1}`

https://github.com/wix/react-native-calendars/blob/master/src/calendar-list/index.js#L205

it should get this as props instead

sibelius:


tautvilas:
thanks

PR #627, opened 2018-09-10
feat(calendar): enable dynamic height, fix #623

### LINKED ISSUE #623: [proposal] dynamic calendar height
## Description
Each month has different number of days, so it has different number of weeks, and so different calendar heights.

Here two screenshots of the behaviours:

hideExtraDays={true), also causes some layout issues

image
image

## Proposal
Use https://github.com/slorber/react-native-scroll-into-view to enable scrolling to a View, without needing to have fixed height

## Expected Behavior

We expect that each calendar adapt the height based on months number of days

## Observed Behavior

The height is fixed per calendar (360px)

## Environment

latest version of react-native-calendars (master)
- 0.53.0

Also specify:

iOS/Android

## Reproducible Demo
Gonna try to setup soon

sibelius:
this would break scroll to month, but it can solve some issues for people that doesn't need scroll to month

we can improve this approach

### DIFF
--- a/src/calendar-list/item.js
+++ b/src/calendar-list/item.js
@@ -44,6 +44,7 @@ class CalendarListItem extends Component {
           dayComponent={this.props.dayComponent}
           disabledByDefault={this.props.disabledByDefault}
           showWeekNumbers={this.props.showWeekNumbers}
+          dynamicHeight={this.props.dynamicHeight}
         />);
     } else {
       const text = row.toString();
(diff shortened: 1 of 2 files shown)


Inbal-Tish:
@sibelius This PR seems to introduce a bug. Please open a new PR that doesn't break existing functionality. Thanks

sibelius:
which bug?

this is very old, so not sure if this makes sense anymore

Inbal-Tish:
I agree. Feel free to open a fresh one if you need it. Thanks!

PR #628, opened 2018-09-10
fix(calendar): always use flex: 1 to fix hideExtraDays={true} not wor…

### LINKED ISSUE #624: hideExtraDays={true} bug, days not aligned
## Description

Days are not aligned when hideExraDays is set to true

image
image

## Expected Behavior

It should keep alignment of days

## Observed Behavior

Days are not aligned

## Environment

latest react-native-calendars
react-native 53

Also specify:

iOS/Android

## Reproducible Demo
Gonna try to make one soon

sibelius:
…king well, fix #624

I've also used early return to avoid mutations


tautvilas:
thanks

PR #632, opened 2018-09-14
Add support for single dot in period style.

NirpE:


Inbal-Tish:
@NirpE Hi. Can you please add to Calendars screen an example for usage? Thank

Issue #532, opened 2018-06-24
renderEmptyDate should rather be renderEmptyData


gopidon:
## Description

The documentation mentions renderEmptyDate as the prop that needs to be passed when you want a custom component to be displayed for an empty date instead of the ActivityIndicator

## Expected Behavior

The documentation must mention **renderEmptyData.** 

The source code expects renderEmptyData rather than renderEmptyDate. See file src/agenda/reservation-list/index.js

> if (!this.props.reservations || !this.props.reservations[this.props.selectedDay.toString('yyyy-MM-dd')]) {
>       if (this.props.renderEmptyData) {
>         return this.props.renderEmptyData();
>       }
>       return (<ActivityIndicator style={{marginTop: 80}}/>);
>     }

PR #633, opened 2018-09-14
Fix typo that causes empty dates to only render ActivityIndicator

seanmclane:
The Agenda component does not render empty dates correctly. This is due to a typo in the render function of "reservation-list/index.js". After I fixed the typo, the Agenda component rendered the function I passed to the renderEmptyDate prop correctly.

### DIFF
--- a/src/agenda/reservation-list/index.js
+++ b/src/agenda/reservation-list/index.js
@@ -182,8 +182,8 @@ class ReactComp extends Component {
 
   render() {
     if (!this.props.reservations || !this.props.reservations[this.props.selectedDay.toString('yyyy-MM-dd')]) {
-      if (this.props.renderEmptyData) {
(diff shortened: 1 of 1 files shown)


tautvilas:
Hey `renderEmptyDate` and `renderEmptyData` are two different options. Perharps you are just using the wrong one

israelidanny:
@tautvilas I don't think he's using the wrong one. According to your examples `renderEmptyDate` should show an empty date, but there is clearly a typo here (because if you test it - it doesn't work without this fix).

This issue should be reopened.

This is also related:
https://github.com/wix/react-native-calendars/issues/532

mohameddroussi:
This is happening in my project. I used yarn to install react native calendars: "react-native-calendars": "^1.187.0",

PR #636, opened 2018-09-16
Avoid Duplicate Key warnings in AgendaView 

frusdelion:
## Description

weekDaysNames have an array of days in abbreviated letters, e.g. "S" for Sunday, and "S" for Saturday. React Native does not like that the key element have duplicates of the same key. This causes the "Encountered two children with the same key" warning. 

## Expected Behavior

No warnings appear in AgendaView.

## Observed Behavior

What actually happened when you performed the above actions?

If there's an error message, please paste the *full terminal output and error message* in this code block:

```
Encountered two children with the same key, `T`...
Encountered two children with the same key, `S`...
```

## Proposed Resolution
Added the index of the weekDaysNames array avoids duplicates, removing the warning.


tautvilas:
Thanks

PR #648, opened 2018-10-05
On the fly changing of style prop for Calendar.

nyur321:
Reason: I want to have a dynamic styling after the parent component finishes layout so I can compute the exact width I need to pass to the calendarWidth or style.width props

Changes: Added style prop to shouldComponentUpdate.

This feature is helpful for styling calendar layout every time the device orientation changes.


Inbal-Tish:
@nyur321 Hi. Why not render the Calendar only after you have its container's width? Why update the calendar width and not re-render with state update it in orientation change?

PR #649, opened 2018-10-08
Add calendar header styles for customization purposes

kirillpisarev:
Thank you for the great lib! I really love it especially because of high customization abilities. But once i faced the situation when i need to customize the calendar's header container style, but had no luck. If you think it's redundant to add styles to that View just reject this PR, there is a chance that i am the only one who encountered this problem :)

### DIFF
--- a/src/calendar/day/period/index.js
+++ b/src/calendar/day/period/index.js
@@ -129,7 +129,8 @@ class Day extends Component {
 
     if (this.props.marking) {
       containerStyle.push({
-        borderRadius: 17
+        borderRadius: 17,
+	backgroundColor: this.markingStyle.day && this.markingStyle.day.color ? this.markingStyle.day.color : this.theme.calendarBackground
       });
 
       const flags = this.markingStyle;
--- a/src/calendar/header/index.js
+++ b/src/calendar/header/index.js
@@ -119,7 +119,7 @@ class CalendarHeader extends Component {
       indicator = <ActivityIndicator />;
     }
     return (
-      <View>
+      <View style={this.style.container}>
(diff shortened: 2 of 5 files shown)


src/calendar/day/period/index.js
```diff
@@ -129,7 +129,8 @@ class Day extends Component {
 
     if (this.props.marking) {
       containerStyle.push({
-        borderRadius: 17
+        borderRadius: 17,
+	backgroundColor: this.markingStyle.day && this.markingStyle.day.color ? this.markingStyle.day.color : this.theme.calendarBackground
```
Inbal-Tish: The containerStyle's backgroundColor is treated a few rows ahead. I don't see why you add it here.

src/calendar/header/index.js
```diff
@@ -119,7 +119,7 @@ class CalendarHeader extends Component {
       indicator = <ActivityIndicator />;
     }
     return (
-      <View>
+      <View style={this.style.container}>
```
Inbal-Tish: Why pass this style here is it returns an empty object?

src/calendar/header/style.js
```diff
@@ -6,6 +6,7 @@ const STYLESHEET_ID = 'stylesheet.calendar.header';
 export default function(theme={}) {
   const appStyle = {...defaultStyle, ...theme};
   return StyleSheet.create({
+    continer: {},
```
Inbal-Tish: Seems redundant...

Inbal-Tish:
@kirillpisarev Closing this PR due to inactivity. Please open a new one if you will like me to review it. Thanks!

PR #650, opened 2018-10-09
longPress event added to the "multi-period" day component

ariverak:


Inbal-Tish:
@ariverak Please remove changes to package.json and package-lock files.

PR #653, opened 2018-10-11
Fix the indentation

HoosLab:


Inbal-Tish:
Fixed on master. Thanks!

PR #654, opened 2018-10-16
Agenda fix - hideExtraDays

rafirr:
### Before:
before

### After:
after


Inbal-Tish:
@rafirr Hi. I don't understand the fix. The extra days are hidden by default when Agenda's calendar is opened. What are you actually fixing?

PR #675, opened 2018-11-05
fix for Agenda loading previous month

### LINKED ISSUE #674: Agenda loadItemsForMonth on scroll uses previous month
## Description

When scrolling the main `Agenda` and `CalendarList` views, `loadItemsForMonth` seems to be using the previous month rather than the current month.

On initial load and on clicking into a date from `CalendarList` views, the month is correct. On scrolling to the next month in Agenda, it begins to use the previous month.

## Expected Behavior

The correct month should be used for the scroll events.

## Observed Behavior

On iOS, the initial load and clicking on a date in `CalendarList` work, while scrolling breaks the month.

On Android, the initial load seems to re-call `loadItemsForMonth` on a `componentDidMount`, triggering it twice with a second 

## Environment

* `npm ls react-native-calendars`: react-native-calendars@1.21.0
* `npm ls react-native`:

Also specify:

1. Phone/emulator/simulator & version: iPhone X + iPhone X Simulator, 12.1 - Android Emulator Pixel 

## Reproducible Demo

nov-05-2018 15-41-07

nov-05-2018 15-45-49

dlimx:
Fixes #674 

Added check for `months[1]` in case the user is at the end of viewable months, where only `months[0]` may be available.


sibelius:
@dlimx can you explain why we need this?

dlimx:
@sibelius on scrolling of both months and days in `<Agenda />`, `loadItemsForMonth` seems to always return the previous month. It was coded to take the 1st month out of 3, which based on my testing was always the previous month.

sibelius:
can you add a test using react-native-testing-library for this behavior?

epshtielsl:
I was having issues with loadItemsForMonth firing with the wrong dates (it was calling the current month instead of the next month coming into view).  After applying the changes in the PR, it started working again.

dlimx:
@sibelius my ETA on that will probably be this weekend/next - any particular tests you want to see or is showing behaviour difference enough?

sibelius:
test should fail in the current branch and pass in the current PR

the test should show when the month[1] is the correctly one to load

douglasjunior:
How about the #746 proposal?

ConnectedReasoning:
This change was not effective for me. If I select a date from the calendar, agenda would navigate to month[1]. However, this update worked for me
  onVisibleMonthsChange(months) {
    if (this.props.items && !this.state.firstResevationLoad) {
      clearTimeout(this.scrollTimeout);
      this.scrollTimeout = setTimeout(() => {
        if (this.props.loadItemsForMonth && this._isMounted) {
            this.props.loadItemsForMonth(xdateToData(this.state.selectedDay));
        }
      }, 200);
    }
  }

frankly I'm not sure why onVisibleMonthsChange exists. All loadItemsForMonth should be programatic or triggered by user interaction. This override makes agenda work how I need it to work

Inbal-Tish:
@dlimx I believe there is a bug in this implementation. Please add the test and see comment by @ConnectedReasoning. Thanks

Issue #642, opened 2018-09-24
endingDay & startingDay are not displayed correct on Android.


seco35:
## Description

endingDay and startingDay =true doesn't works properly on Android working without Expo. It works when using expo.

## Expected Behavior

Marking with rounded edges.

## Observed Behavior

Marking as a rectangle ( as if it had startingDay & endingDay = false).

<img width="368" alt="bildschirmfoto 2018-09-24 um 12 35 17" src="https://user-images.githubusercontent.com/25101244/45947936-598be300-bff6-11e8-9f84-6d6eb91e690f.png">

## Environment

Please run these commands in the project folder and fill in their results:

* `npm ls react-native-calendars`: 1.20.0
* `npm ls react-native`: 0.57.1

Emulator Nexus 5 : API 28 (Android 9)
Huawei Mediapad M5: API 26 (Android 8)

## Reproducible Demo

mDates["2018-10-05"] = { startingDay: true, color: lockColor, textColor:'black', endingDay: true, selection:false}

adimshev:
+1

adimshev:
![image](https://user-images.githubusercontent.com/33316685/48701366-61839000-ebff-11e8-8762-5f0cf5f80215.png)

wojciechver:
+1

npm ls react-native-calendars: 1.21.0
npm ls react-native: 0.57.5

Having the same issue on android (works fine on iOS). 
It's only present after a selection is made on the calendar. If a period is sent to the calendar on open, it rounds the corners properly.

on load:
<img width="358" alt="screen shot 2018-11-23 at 14 53 34" src="https://user-images.githubusercontent.com/32451034/48946943-b260f480-ef2f-11e8-95fd-40cd587e8e53.png">


after new selection:
<img width="370" alt="screen shot 2018-11-23 at 14 51 09" src="https://user-images.githubusercontent.com/32451034/48946934-a6753280-ef2f-11e8-83c6-a888a5da83cb.png">

adimshev:
You can use prop dayComponent to pass your own day component. It works for me.

kalraneeraj24550:
plz help me, calnder doesnt give rounded period corner.
![screenshot from 2018-12-18 13-29-39](https://user-images.githubusercontent.com/44640902/50139891-ffcf4800-02c8-11e9-8c3d-f30ee0ba1b12.png)
getDateString(timestamp) {
        const date = new Date(timestamp)
        const year = date.getFullYear()
[...]

PR #700, opened 2018-12-10
solve the bug #642

afresh:


Inbal-Tish:
@afresh Hi. Is this bug still relevant? I can't reproduce in RN 0.61.4...

Issue #705, opened 2018-12-20
selectedDayTextColor doesn't applied to if markingType is custom


sasurau4:
## Description

selectedDayTextColor in Calendar component doesn't applied to Day when using markingType as custom.

## Expected Behavior

The day has markedDates as true, `selectedDayTextColor` applied

## Observed Behavior

The dayTextColor hasn't changed when date as `selected: true`

## Environment

```
- react-native: 0.57.7
- react-native-calendars: 1.20.0
- iOS simulator: iPhone X
```

## Reproducible Demo

https://snack.expo.io/@sasurau4/react-native-calendar-repro

PR #706, opened 2018-12-20
fix selectedTextStyle in custom day component

sasurau4:
fixes https://github.com/wix/react-native-calendars/issues/705


Inbal-Tish:
Fix already exists in master.

PR #708, opened 2018-12-20
Add support for dots on period style

cristeahub:


felipesaruhashi:
This feature it is working? I need exactly that! How do you use both dots and period? You only need to add the dos on markedDates item?

cristeahub:
It does work, but when a date is selected you lose the dot. I bet you only need to add the same code to the other place as well for that to work, but it was outside my usecase, so I left it as it is now. Feel free to add that part on if you need it!

Inbal-Tish:
@cristeahub Seems this PR is incomplete as it presents the bug mentioned. I'm closing this PR, please open a new one if you like me to review suggested changes. Thanks

PR #717, opened 2019-01-11
Added support for passing custom styles to dots in multi-dot calendar

Elijen:
Title and diff is quite self-explanatory. Added the possibility to use custom styles (e.g. maybe we want a square instead of a dot for some days or we want a different border etc.)

### DIFF
--- a/src/calendar/day/multi-dot/index.js
+++ b/src/calendar/day/multi-dot/index.js
@@ -49,7 +49,7 @@ class Day extends Component {
       const validDots = marking.dots.filter(d => (d && d.color));
       return validDots.map((dot, index) => {
         return (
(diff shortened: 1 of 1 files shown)


src/calendar/day/multi-dot/index.js
```diff
@@ -43,13 +43,13 @@ class Day extends Component {
   }
 
   renderDots(marking) {
-    const baseDotStyle = [this.style.dot, this.style.visibleDot];
+    const baseDotStyle = [this.style.dot, this.style.visibleDot, marking.selected && this.style.selectedDot];
```
Inbal-Tish: `style.selectedDot` only hold backgroundColor from appStyle's `selectedDotColor`. That already treated in line 55. Why add it again here?

PR #736, opened 2019-02-12
Feature/text custom font weight

AdamGold:
Allow users of the package to edit ``fontWeight`` attribute of ``textDay`` and ``selectedDay``.


README.md
```diff
@@ -275,6 +275,7 @@ The loading indicator next to month name will be displayed if `<Calendar />` has
     textSectionTitleColor: '#b6c1cd',
     selectedDayBackgroundColor: '#00adf5',
     selectedDayTextColor: '#ffffff',
+    selectedDayFontWeight: 'bold',
```
Inbal-Tish: You should add these two props to src/style.js, not just to the README file

PR #737, opened 2019-02-12
#598: Fixed slow calendar opening

### LINKED ISSUE #598: Calendar opens slower in android

LimAlbert:
Fixed #598 following @tautvilas suggestion(https://github.com/wix/react-native-calendars/pull/605#issuecomment-425081913), initialNumToRender will always be equal to the number of calendar that fits the screen

### DIFF
--- a/src/calendar-list/index.js
+++ b/src/calendar-list/index.js
@@ -11,7 +11,7 @@ import dateutils from '../dateutils';
 import Calendar from '../calendar';
 import CalendarListItem from './item';
 
-const {width} = Dimensions.get('window');
+const {width, height} = Dimensions.get('window');
 
 class CalendarList extends Component {
   static propTypes = {
@@ -195,6 +195,10 @@ class CalendarList extends Component {
     return diffMonths;
   }
 
+  getNumToRender(screenSize, calendarSize) {
+    return Math.ceil(screenSize / calendarSize);
+  }
+  
   render() {
     return (
       <FlatList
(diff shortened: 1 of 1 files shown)


muhammadridho:
any update in here, i got this issue too, cant wait on this merge request.

src/calendar-list/index.js
```diff
@@ -219,6 +223,7 @@ class CalendarList extends Component {
         initialScrollIndex={this.state.openDate ? this.getMonthIndex(this.state.openDate) : false}
         getItemLayout={this.getItemLayout}
         scrollsToTop={this.props.scrollsToTop}
+        initialNumToRender={this.props.horizontal ? this.getNumToRender(width, this.props.calendarWidth) : this.getNumToRender(height, this.props.calendarHeight)}
```
Inbal-Tish: Why not call for this calculation once in the constructor and use the value here instead of calling the method on every render (scroll)?

adrhub:
this doesn't make a significant difference, it still remains slow

Inbal-Tish:
@LimAlbert Closing this PR due to inactivity. Please open a new one if you will like me to review it. Thanks!

PR #746, opened 2019-02-22
Added options to use 'onVisibleMonthsChange' directly from <Agenda />

marllonfrizzo:
This is necessary to know all months that are visible on screen.


douglasjunior:
@ethanshar any updates on this?

Today the `Agenda` only allow to know the "Next" month in the scrollview, but is useless when you need to know all months visible.

PR #774, opened 2019-03-20
Add disabledByDefault props to docs

osterbergmarcus:
If you wan't to whitelist specific dates you can use the "disabledByDefault" prop to disable all dates and then explicitly add disabled: false to the dates that should be enabled. I found this to be very useful.

<Calendar
  markedDates={{  'YYYY-MM-DD': { disabled: false }}
  disableByDefault={false}
/>

### DIFF
--- a/README.md
+++ b/README.md
@@ -444,7 +444,7 @@ An advanced agenda component that can display interactive listings for calendar 
     '2012-05-18': {disabled: true}
   }}
   // If disabledByDefaut={true} dates flaged as not disabled will be enabled. Default = false
-  disableByDefault={true}
+  disabledByDefault={true}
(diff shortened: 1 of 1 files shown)


Inbal-Tish:
This PR seems incomplete. Where is the use of the 'disableByDefault' prop in the Calendar?

README.md
```diff
@@ -443,6 +443,8 @@ An advanced agenda component that can display interactive listings for calendar
     '2012-05-17': {marked: true},
     '2012-05-18': {disabled: true}
   }}
+  // If disabledByDefaut={true} dates flaged as not disabled will be enabled. Default = false
+  disableByDefault={true}
```
Inbal-Tish: You have a typo. It should be 'disabledByDefault' (you're missing a 'd' in the end of 'disable')

PR #775, opened 2019-03-20
Add theme color for indicator

donni106:
The indicator color was not editable.
The default color of React Native ActivityIndicator is gray:
  https://facebook.github.io/react-native/docs/activityindicator#color

- added indicatorColor to the default theme styles
- added color prop for both ActivityIndicator renderings
- included indicatorColor in README and example

Bildschirmfoto 2019-03-20 um 17 19 39

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -1,4 +1,4 @@
-import {Platform} from 'react-native';
+import { Platform } from 'react-native';
 
 export const foregroundColor = '#ffffff';
 export const backgroundColor = '#f4f4f4';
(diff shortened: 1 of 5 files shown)


src/style.js
```diff
@@ -35,6 +35,7 @@ export const dotColor = textLinkColor;
 export const selectedDotColor = foregroundColor;
 export const arrowColor = textLinkColor;
 export const monthTextColor = textDefaultColor;
+export const indicatorColor = '#808080'; // gray is the default of React Native ActivityIndicator
```
Inbal-Tish: Why not just pass undefined as default so the color will be set to the RN default color?

donni106: Sounds great. I did not know, that this is working this way, when setting `undefined`.
How can I test that?
Thx, I will update my PR according to this later this week.

Inbal-Tish: Just run it with the undefined to see if the ActivityIndicator is running well.

Inbal-Tish: Please resolve the conflicts as well

donni106: Sure.

donni106: Yes, working.

Bildschirmfoto 2019-05-08 um 13 08 44

shimatai:
It's not working anymore in version 1.212.0

`markingType={'custom'}
theme={{
    indicatorColor: '#000'
}}`

PR #789, opened 2019-04-16
Added monthComponent

thousight:
Added ability to customize the month and year component display

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,8 @@ LocaleConfig.defaultLocale = 'fr';
   onDayLongPress={(day) => {console.log('selected day', day)}}
   // Month format in calendar title. Formatting values: http://arshaw.com/xdate/#Formatting
   monthFormat={'yyyy MM'}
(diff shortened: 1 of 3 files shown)


Inbal-Tish:
@thousight Hi. What is the use for the custom month component? What would you like to customize? Screenshots will be of help

thousight:
It is for adding more margins or making the month and year text size different. I left my original company already so I can't provide you screenshots.

Inbal-Tish:
@thousight Hi. What is this 'monthComponent' you want to add to the CalendarHeader? Can you send me a screenshot of a use-case for this functionality?

sykenrussel:
image
Hi I don't think I would get a reply from this, but an example screenshot would be a monthly charts like this one.

Inbal-Tish:
@sykenrussel Why would you want it to be part of the Calendar component? You can use 'renderHeader' prop to customize the header if you need.

PR #792, opened 2019-04-17
Add headingLevel prop to allow for aria-level for web rendering

DaKaZ:
We use this great library for cross platform support including web with react-native-web.  In order to provide proper Section 508 Accessibility, the elements rendered as a Heading on web need a corresponding `aria-level`.  This is PR adds supports for that functionality and makes no changes to iOS and Android or any other platform.  A user of the component can now pass in `headingLevel={3}` to make this is a 3rd level heading.


DaKaZ:
@Inbal-Tish I am wondering if you all could comment on this very small PR.

src/calendar/index.js
```diff
@@ -77,14 +77,16 @@ class Calendar extends Component {
     /** Handler which gets executed when press arrow icon left. It receive a callback can go next month */
     onPressArrowRight: PropTypes.func,
     /** Style passed to the header */
-    headerStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array])
+    headerStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array]),
+    /** Provide heading level for proper accessibility when used with web (react-native-web) */
+    headingLevel: PropTypes.number
```
Inbal-Tish: Why not call the prop `ariaLevel` to make it more explicit?

DaKaZ: @Inbal-Tish thank you for the review.  I'd gladly make the change if there was a hope for this getting merged.  Let me know.

DaKaZ: I went ahead and updated the PR and made the change since it seems like the project is active again.  Thanks!

src/calendar/header/index.js
```diff
@@ -141,12 +143,18 @@ class CalendarHeader extends Component {
...
       <View style={this.props.style}>
         <View style={this.style.header}>
           {leftArrow}
           <View style={{ flexDirection: 'row' }}>
-            <Text allowFontScaling={false} style={this.style.monthText} accessibilityTraits='header'>
+            <Text
+              allowFontScaling={false}
+              style={this.style.monthText}
+              accessibilityTraits='header'
+              {...webProps}
```
Inbal-Tish: I prefer this syntax over the spread and empty object: `aria-level={Platform.OS === 'web' ? this.props.webAriaLevel : undefined}`. Can it work?

DaKaZ: Sure, the webProps was just a clean way to encapsulate it.  But I can make that change

DaKaZ: However... you will get a react warning about this, since in native `aria-level` is not a valid prop.  That will drive most developers batty.  The current implementation avoids the warning and is the pattern we use in our app.

PR #815, opened 2019-05-10
Adds a flag to hide the left day label on agenda

felipesaruhashi:


src/agenda/reservation-list/index.js
```diff
@@ -30,13 +30,20 @@ class ReactComp extends Component {
...
+
     selectedDay: PropTypes.instanceOf(XDate),
     topDay: PropTypes.instanceOf(XDate),
     refreshControl: PropTypes.element,
     refreshing: PropTypes.bool,
     onRefresh: PropTypes.func,
   };
 
+  static defaultProps = {
+    showDate: true
```
Inbal-Tish: If you called the prop 'hideDate' you shouldn't have to set the default to true, plus it makes more sense because you want the option to hide the date...

Inbal-Tish:
@felipesaruhashi Hi. Can you please specify a scenario where you want to hide the date? It's just doesn't seem realistic to me...

felipesaruhashi:
Hi! I was in a project that I needed more space for the agenda views, and inside the custom views, I would inform the day that view belongs. That is why I added this flag.

Inbal-Tish:
@felipesaruhashi It doesn't seem like a feature that can serve other users, so I'm closing this PR. Thank you!!!

PR #631, opened 2018-09-13
Pass current month to arrow press callbacks

olfek:

PR #816, opened 2019-05-12
Fix: Navigation arrows not navigating along 'CalendarList'

olfek:
Replaces #631 

Before fix:
before_fix

After fix:
after_fix

These gifs can be explained by [this comment:](https://github.com/wix/react-native-calendars/pull/631#discussion_r279191482)

> Also, If I remember correctly, this is **needed** for navigating months in `CalendarList`.
> 
> This is because `CalendarList` renders a list of `CalendarListItem` which renders a `Calendar` which includes a `CalendarHeader`, when the left/right navigation arrows of the `CalendarHeader` are pressed, instead of navigating along the `CalendarList`, the `Calendar` is updated in place to show the month navigated to.
> 
> So for example, if `CalendarList` looked like:
> 
> ```
> -------+-------+-------+-------+
>  Index |   0   |   1   |   2   |
> -------+-------|-------|-------|
>  Month |  Jan  |  Feb  |  Mar  |
> -------+-------+-------+-------+
> ```
> 
> When `Jan` is shown, if the right arrow is pressed, instead of moving to the next `CalendarItem` in the `CalendarList` at `1`, The `CalendarItem` at `0` updates itself to show `Feb`. So now, `CalendarList` looks like:
> 
> ```
> -------+-------+-------+-------+
>  Index |   0   |   1   |   2   |
> -------+-------|-------|-------|
>  Month |  Feb  |  Feb  |  Mar  |
> -------+-------+-------+-------+
> ```
> 
> This is happening because in `CalendarHeader`:
> 
> ```
> onPressLeft() {
>     const {onPressArrowLeft} = this.props;
>     if(typeof onPressArrowLeft === 'function') {
>         return onPressArrowLeft(this.substractMonth);
>     }
>     return this.substractMonth();
> }
> 
> onPressRight() {
>     const {onPressArrowRight} = this.props;
>     if(typeof onPressArrowRight === 'function') {
>         return onPressArrowRight(this.addMonth);
>     }
>     return this.addMonth();
> }
> ```
> 
> which uses:
> 
> ```
> addMonth() {
>     this.props.addMonth(1);
> }
> 
> substractMonth() {
>     this.props.addMonth(-1);
> }
> ```
> 
> which uses `addMonth` in `Calendar`:
> 
> ```
> addMonth(count) {
>     this.updateMonth(this.state.currentMonth.clone().addMonths(count, true));
> }
> ```
> 
> which **replaces** the month displayed by `Calendar`.
> 
> This behaviour is incorrect when `Calendar` is used **inside** a `CalenderList`. When the left/right navigation arrows of `CalendarHeader` are pressed, we need to **move along** the `CalendarList` and display the previous/next `CalendarItem`, not update the month of the displayed `Calendar` item held by `CalendarList`.
> 
> To fix this, I added the following code to `CalendarList` which uses the changes in this MR:
> 
> ```
> onPressArrowLeft={(_,month)=>{
>     month = month.clone()
>     month.addMonths(-1)
>     this.refs.calendarList.scrollToMonth(month)
> }}
> onPressArrowRight={(_,month)=>{
>     month = month.clone()
>     month.addMonths(1)
>     this.refs.calendarList.scrollToMonth(month)
> }}
> ```
> 
> Where `calendarList` is a reference to the `CalendarList`.

@Inbal-Tish New MR

### DIFF
--- a/src/calendar-list/index.js
+++ b/src/calendar-list/index.js
@@ -192,7 +192,8 @@ class CalendarList extends Component {
 
   renderCalendar({item}) {
     return (
-      <CalendarListItem 
+      <CalendarListItem
+        calendarListContainer={this}
         item={item} 
         calendarHeight={this.props.calendarHeight} 
         calendarWidth={this.props.horizontal ? this.props.calendarWidth : undefined} 
--- a/src/calendar-list/index.js
+++ b/src/calendar-list/index.js
@@ -193,7 +193,7 @@ class CalendarList extends Component {
   renderCalendar({item}) {
     return (
       <CalendarListItem
-        calendarListContainer={this}
+        scrollToMonth={this.scrollToMonth.bind(this)}
         item={item} 
         calendarHeight={this.props.calendarHeight} 
         calendarWidth={this.props.horizontal ? this.props.calendarWidth : undefined} 
--- a/src/calendar-list/item.js
+++ b/src/calendar-list/item.js
@@ -48,14 +48,14 @@ class CalendarListItem extends Component {
           showWeekNumbers={this.props.showWeekNumbers}
           renderArrow={this.props.renderArrow}
           onPressArrowLeft={this.props.onPressArrowLeft ? this.props.onPressArrowLeft : (_,month)=>{
-            month = month.clone()
-            month.addMonths(-1)
-            this.props.calendarListContainer.scrollToMonth(month)
+            month = month.clone();
+            month.addMonths(-1);
+            this.props.calendarListContainer.scrollToMonth(month);
           }}
           onPressArrowRight={this.props.onPressArrowRight ? this.props.onPressArrowRight : (_,month)=>{
-            month = month.clone()
-            month.addMonths(1)
-            this.props.calendarListContainer.scrollToMonth(month)
+            month = month.clone();
+            month.addMonths(1);
+            this.props.calendarListContainer.scrollToMonth(month);
           }}
           headerStyle={this.props.headerStyle}
         />);
--- a/src/calendar-list/item.js
+++ b/src/calendar-list/item.js
@@ -23,7 +23,7 @@ class CalendarListItem extends Component {
 
   render() {
     const row = this.props.item;
-    
+
     if (row.getTime) {
       return (
         <Calendar
@@ -47,13 +47,21 @@ class CalendarListItem extends Component {
           disabledByDefault={this.props.disabledByDefault}
           showWeekNumbers={this.props.showWeekNumbers}
           renderArrow={this.props.renderArrow}
-          onPressArrowLeft={this.props.onPressArrowLeft}
-          onPressArrowRight={this.props.onPressArrowRight}
+          onPressArrowLeft={this.props.onPressArrowLeft ? this.props.onPressArrowLeft : (_,month)=>{
+            month = month.clone()
+            month.addMonths(-1)
+            this.props.calendarListContainer.scrollToMonth(month)
+          }}
+          onPressArrowRight={this.props.onPressArrowRight ? this.props.onPressArrowRight : (_,month)=>{
+            month = month.clone()
+            month.addMonths(1)
+            this.props.calendarListContainer.scrollToMonth(month)
+          }}
           headerStyle={this.props.headerStyle}
         />);
     } else {
       const text = row.toString();
-      
+
       return (
         <View style={[{height: this.props.calendarHeight, width: this.props.calendarWidth}, this.style.placeholder]}>
           <Text allowFontScaling={false} style={this.style.placeholderText}>{text}</Text>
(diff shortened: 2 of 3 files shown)


olfek:
@Inbal-Tish ...

olfek:
@Inbal-Tish Tests have now passed

olfek:
@Inbal-Tish 

The actions in the GIFs above, are as follows.

**Before fix**

1. Swipe left to go from May to June. Observe that it says May and then June.
1. Swipe right to go back from June to May.
1. Click the right navigation arrow. Observe that it now says June.
1. Swipe left to go to the next month. Observe that the next month is also June.
1. Swipe right to go back to previous month.
1. Click the left navigation arrow until it says April.
1. Swipe right to go to the preceding month. Observe that this also says April.
1. Swipe left to go back to the month we were at in the last step.
1. Click right navigation arrow to make April go back to May.
1. Swipe right again to go to the preceding month. Observe that things are now correct again.

**After fix**

1. Swipe left to go from May to June. Observe that it says May and then June.
1. Swipe right to go back from June to May.
1. Click the right navigation arrow. Observe that is now says June.
1. Swipe left to go the next month. Observe that the next month is now July.
1. Swipe right to go back to the previous month.
1. Click the left navigation arrow until it says April.
1. Swipe right to go the preceding month. Observe that this is now March.
1. Swipe left to go back to the month we were at in the last step.
1. Click right navigation arrow to make April go back to May.
1. Swipe right again to go to the preceding month. Observe that that this is now April.

olfek:
@Inbal-Tish Please can you merge this, I have been waiting 8 months!

src/calendar-list/index.js
```diff
@@ -187,7 +187,8 @@ class CalendarList extends Component {
 
   renderCalendar({item}) {
     return (
-      <CalendarListItem 
+      <CalendarListItem
+        calendarListContainer={this}
```
Inbal-Tish: Passing this/ref of the list to every item is not a good practice. You should use a callback which you then invoke in the onPress functions. For example: 
moveMonth = (month) => { 
    this.scrollToMonth(month); 
} 
and pass it to the item:  moveMonth={this.moveMonth}
and invoke it safely from the item: _.invoke(this.props, 'moveMonth');

src/calendar-list/item.js
```diff
@@ -47,8 +47,16 @@ class CalendarListItem extends Component {
           disabledByDefault={this.props.disabledByDefault}
           showWeekNumbers={this.props.showWeekNumbers}
           renderArrow={this.props.renderArrow}
-          onPressArrowLeft={this.props.onPressArrowLeft}
-          onPressArrowRight={this.props.onPressArrowRight}
+          onPressArrowLeft={this.props.onPressArrowLeft ? this.props.onPressArrowLeft : (_,month)=>{
```
Inbal-Tish: Using arrow functions in props is not advisable, you should move the functionality to separate methods. 
Also, you should check if 'this.props.calendarListContainer' is not undefined before invoking its method (if you use a callback, as I suggested in the first comment, you can instead call it safely using Lodash's 'invoke'). 
Third, you shouldn't assign to an argument, declare a new const for 'month' clone.

src/calendar-list/index.js
olfek: Ok, I will see what I can do

olfek:
@Inbal-Tish What do you think?

olfek:
Tested it too, everything still works :+1:

olfek:
@Inbal-Tish Resolved merge conflicts, tested again, still works. Please merge.

src/calendar-list/item.js
```diff
@@ -47,13 +71,13 @@ class CalendarListItem extends Component {
           disabledByDefault={this.props.disabledByDefault}
           showWeekNumbers={this.props.showWeekNumbers}
           renderArrow={this.props.renderArrow}
-          onPressArrowLeft={this.props.onPressArrowLeft}
-          onPressArrowRight={this.props.onPressArrowRight}
+          onPressArrowLeft={this.onPressArrowLeft.bind(this)}
```
Inbal-Tish: Please don't use bind in props. You can use the arrow function on the methods' signatures. For example:
`onPressArrowRight = (_, month) => {
    ...
}`

olfek: @Inbal-Tish I've made the requested changes. Please merge.

src/calendar-list/item.js
```diff
@@ -47,13 +71,13 @@ class CalendarListItem extends Component {
           disabledByDefault={this.props.disabledByDefault}
           showWeekNumbers={this.props.showWeekNumbers}
           renderArrow={this.props.renderArrow}
-          onPressArrowLeft={this.props.onPressArrowLeft}
-          onPressArrowRight={this.props.onPressArrowRight}
+          onPressArrowLeft={this.onPressArrowLeft}
```
Inbal-Tish: I would limit this fix to horizontal only, so we won't break the current behaviour of CalendatList in vertical for those who choose to add the navigation arrows for some reason. Consider something like:
`onPressArrowLeft={this.props.horizontal ? this.onPressArrowLeft : this.props.onPressArrowLeft}
  onPressArrowRight={this.props.horizontal ? this.onPressArrowRight : this.props.onPressArrowRight}`

olfek: @Inbal-Tish Okay, I've made these changes.

olfek:
@Inbal-Tish merge?

PR #827, opened 2019-05-21
Selected day should have background color

Nigh7Sh4de:
Currently days in the period markingStyle do not set a backgroundColor
on the container. I updated the `getDrawingStyle` method in
_period/index.js_ to set containerStyle.backgroundColor for when
the marking has `selected: true`


Nigh7Sh4de:
Oops didn't notice this was already done on master

PR #834, opened 2019-05-29
six rows of month, last line will not displayed

taixw2:
If six rows are occupied that month, the last line will not be displayed.

As shown, June 30 is hidden

image

### DIFF
--- a/src/calendar-list/item.js
+++ b/src/calendar-list/item.js
@@ -28,7 +28,7 @@ class CalendarListItem extends Component {
       return (
         <Calendar
           theme={this.props.theme}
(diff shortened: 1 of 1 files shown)


Inbal-Tish:
@taixw2 Hi. I don't see the issue in our example project. Did you change some style to result in this issue? Is it visible in all devices and os? See my screenshot: 
Simulator Screen Shot - iPhone X - 2019-05-30 at 15 36 44

taixw2:
Sorry, I changed the width and height of the style. That's my problem. I need to show the calendar price.

image

Issue #659, opened 2018-10-24
CalendarList does not implement `onMonthChange`


FabianMeul:
## Description
The component `README.md` describes the following;
> `<CalendarList />` is scrollable semi-infinite calendar composed of `<Calendar />` components. Currently it is possible to scroll 4 years back and 4 years to the future. All paramters that are available for `<Calendar />` are also available for this component.

## Expected Behavior

Implementing `<CalendarList onMonthChange={} />` triggers the callback when the active month is changed by either a swipe, of a press of the header buttons.

## Observed Behavior

The callback is not fired. The source code of the CalendarList also provides no reference to an onMonthCallback whatsoever.

## Environment
* `react-native-calendars@1.21.0`
* `react-native@0.57.2`

bansawbanchee:
The method to use is: `onVisibleMonthsChange={(months) => {console.log('now these months are visible', months)}}`.

Also `...Calendar.PropTypes` are a part of CalendarList. I've linked you to the line number in the source code as well below.

https://github.com/wix/react-native-calendars/blob/master/src/calendar-list/index.js#L17

victorsilent:
> 
> 
> The method to use is: `onVisibleMonthsChange={(months) => {console.log('now these months are visible', months)}}`.
> 
> Also `...Calendar.PropTypes` are a part of CalendarList. I've linked you to the line number in the source code as well below.
> 
> https://github.com/wix/react-native-calendars/blob/master/src/calendar-list/index.js#L17
> 
> Example from the docs with `onVisibleMonthChange` added.
> 
> ```
> import React, { Component } from 'react';
> 
> import {CalendarList} from 'react-native-calendars';
> import {View} from 'react-native';
> 
> export default class HorizontalCalendarList extends Component {
>   constructor(props) {
>     super(props);
>   }
> 
>   render() {
>     return (
>       <View>
>         <CalendarList
>           //current={'2012-05-16'}
>           pastScrollRange={24}
>           futureScrollRange={24}
>           horizontal
>           pagingEnabled
>           onVisibleMonthsChange={(months) => {alert('now these months are visible', months)}}
>           style={{borderBottomWidth: 1, borderBottomColor: 'black'}}
>         />
>       </View>
>     );
>   }
> }
> ```

CalendarList has the onMonthChange props, but never call it, and the onVisibleMonthsChange are a little bit trickier because sometimes you have two months or one month and you don't know which month you will do the action

FabianMeul:
[...]

PR #838, opened 2019-06-03
CalendarList does not implement `onMonthChange` #659

Jacklol:
https://github.com/wix/react-native-calendars/issues/659 solve this problem

### DIFF
--- a/src/calendar-list/index.js
+++ b/src/calendar-list/index.js
@@ -283,6 +283,9 @@ class CalendarList extends Component {
           data={this.state.rows}
           //snapToAlignment='start'
           //snapToInterval={this.calendarHeight}
+          viewabilityConfig={{
+            itemVisiblePercentThreshold: 50
+          }}
           removeClippedSubviews={this.props.removeClippedSubviews}
           pageSize={1}
           horizontal={this.props.horizontal}


Jacklol:
can merge it or  tell  what wrong?

webdevsyd:
I hope this PR can be merge immediately.

src/calendar-list/index.js
```diff
@@ -283,6 +283,9 @@ class CalendarList extends Component {
           data={this.state.rows}
           //snapToAlignment='start'
           //snapToInterval={this.calendarHeight}
+          viewabilityConfig={{
+            itemVisiblePercentThreshold: 50
+          }}
```
Inbal-Tish: Currently the viewabilityConfig sets itemVisiblePercentThreshold to 20. I don't see how that or changing it to 50 will fix the issue with invoking onMonthChange

Jacklol: https://stackoverflow.com/questions/45868284/how-to-get-currently-visible-index-in-rn-flat-list

Jacklol: Its wark i checked

Jacklol: onVisibleMonthsChange={(months) => { console.log('now these months are visible', this.setMonths(months)); }}
          pagingEnabled={true}
 result =>{year: 2019, month: 2, day: 19, timestamp: 1550534400000, dateString: "2019-02-19"}

Jacklol: u can checked too

Jacklol: we need current month , but now  if(pagingEnabled={true})  
result => #659

Inbal-Tish: First, setting itemVisiblePercentThreshold to 50 is will give you the same behaviour as the FlatList's default. Second, we already have 'this.viewabilityConfig' defined in the c'tor and you're ignoring it. And last creating a new object in the render function is bad practice. Please run the library again and check if your fix is still relevant

Jacklol: U are right in last version all wark thanks!!!  i warked width version 1 . 68

Inbal-Tish: That's great. Can you please close the PR

Issue #333, opened 2018-02-07
Agenda isCalendarOpen prop


cribcutkiran:
## Description

Is it possible to set a prop on the Agenda component to specify whether it should be in the calendar or agenda view? The use case for me is that when the calendar has been opened and the user navigates to a different screen and then back, I want to be able to reset the Agenda to not have the calendar opened.

I am currently doing this using a ref to the agenda component:
```
onNavigatorEvent(event) {
  if(event.type === 'ScreenChangedEvent' && event.id === 'didDisappear') {
    this._agenda.chooseDay(this.agenda.state.selectedDay, false);
  }
}
```
Please let me know if there is a cleaner way to do this. Otherwise, I think having a prop would be a nice enhancement.

## Expected Behavior

isCalendarOpen=true should show the Calendar opened, isCalendarOpen=false should show the agenda view.

## Observed Behavior

Not implemented (asfaik), as of version 1.17.0

tautvilas:
Hi, currently I think your solution is the only way to do this

Issue #277, opened 2017-12-18
Agenda with calendar view


loic-lopez:
## Description

Hello, I would like to create a view with agenda component.
In my case I would like to display the calendar first instead of the day list:

Example, when the user come to the view, he got the calendar, and, on press, got the agenda view:

![simulator screen shot - iphone 6 - 2017-12-18 at 17 06 37](https://user-images.githubusercontent.com/17724398/34115475-0ce08414-e416-11e7-8dc2-e479d93ca023.png)

PR #227, opened 2017-11-07
Agenda View toggled opened/closed callback

dborstelmann:
We had a need to render content based off whether the calendar was opened or not.  This just adds a prop to have a callback function fire with the state of the calendar when it's toggled opened or closed.  Would love to get this merged asap because it's so small!  Thanks.

PR #854, opened 2019-06-16
shouldOpenCalendar prop for Agenda

shuzootani:
### added `shouldOpenCalendar` prop for `<Agenda />`
this allows you to control a flag to show CalendarView by passing a prop to `<Agenda />`

it makes more sense to toggle CalendarView depending on `shouldOpenCalendar` prop rather than calling `chooseDay()`.

related to #333 #277 #227 

## use cases
- for the simple case like #277 
if you wanna show calendar view only at initial render,
you can just pass `shouldOpenCalendar: true` to `<Agenda />`

- for the case like #333 
suppose you're using [react-native-navigation](https://github.com/wix/react-native-navigation) 
and you wanna show CalendarView not just when the component is mounted, but also whenever the user comes back to the screen where `<Agenda />` is rendered.

``` js
import { Navigation } from 'react-native-navigation'

constructor() {
    this.state = { shouldOpenCalendar: false }
    Navigation.events().bindComponent(this);
}

componentDidAppear() {
    this.setState({ shouldOpenCalendar: true })
}

<Agenda shouldOpenCalendar={this.state.shouldOpenCalendar} />

```

### DIFF
--- a/src/agenda/index.js
+++ b/src/agenda/index.js
@@ -89,6 +89,8 @@ export default class AgendaView extends Component {
     refreshing: PropTypes.bool,
     // Display loading indicador. Default = false
     displayLoadingIndicator: PropTypes.bool,
+    // If true, open calendar
+    shouldOpenCalendar: PropTypes.bool,
   };
 
   constructor(props) {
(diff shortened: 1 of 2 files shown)


src/agenda/index.js
```diff
@@ -135,7 +137,9 @@ export default class AgendaView extends Component {
     // When user touches knob, the actual component that receives touch events is a ScrollView.
     // It needs to be scrolled to the bottom, so that when user moves finger downwards,
     // scroll position actually changes (it would stay at 0, when scrolled to the top).
-    this.setScrollPadPosition(this.initialScrollPadPosition(), false);
+    if (!this.props.shouldOpenCalendar) {
```
shuzootani: prevent an animation conflict on initial render so that it either sets initial position or open the calendar

shuzootani:
for naming, which do you prefer `isCalendarOpen` or `shouldOpenCalendar`?

raen79:
👍

shuzootani:
@raen79 Did you test my branch?
This is just my PR for this feature, which is not merged yet!

Inbal-Tish:
@shuzootani Hi. Thank you for your contribution! 
I like the idea behind this PR, but I would like to suggest a modification to 'initialPosition' (see ExapandableCalendar) instead of 'shouldOpenCalendar' to control the calendar's position (open/closed). First, since it is clearer to the user and, second, for keeping the API consistent. 
Hope you can make the changes. Thanks!

PR #889, opened 2019-07-25
Fix/updating dependencies

Inbal-Tish:


AleksandrZhukov:
That's not a good changes, importing lodash as `_` will inject whole lodash library to the bundle. Much better to import only functions that you need, eg `import get from 'lodash/get'` or you can add babel plugin `babel-plugin-lodash` and then you can do `import { get } from 'lodash'`

PR #894, opened 2019-07-29
Added accessibility props to days

mymattcarroll:


src/calendar-list/item.js
```diff
@@ -67,6 +67,7 @@ class CalendarListItem extends Component {
           maxDate={this.props.maxDate}
           firstDay={this.props.firstDay}
           monthFormat={this.props.monthFormat}
+          dayAccessibilityHint={this.props.dayAccessibilityHint}
```
Inbal-Tish: This one should be passed from the CalendarList component. No?

src/interface.js
```diff
@@ -14,7 +14,8 @@ function xdateToData(xdate) {
     month: xdate.getMonth() + 1,
     day: xdate.getDate(),
     timestamp: XDate(dateString, true).getTime(),
-    dateString: dateString
+    dateString: dateString,
+    friendly: xdate.toString('dddd MMMM d')
```
Inbal-Tish: 'friendly' is not a clear prop name. Also, will be better to allow the user to send the format he wants instead of setting it hard coded.

src/calendar/day/multi-period/index.js
```diff
@@ -7,7 +7,7 @@ import styleConstructor from './style';
 
 class Day extends Component {
   static displayName = 'IGNORE';
-  
+
```
Inbal-Tish: Day/period.js is missing from this PR

src/expandableCalendar/week.js
```diff
@@ -132,6 +132,7 @@ class Week extends Component {
       <View style={{flex: 1, alignItems: 'center'}} key={id}>
         <DayComp
           testID={`${SELECT_DATE_SLOT}-${dateAsObject.dateString}`}
+          accessibilityHint={this.props.dayAccessibilityHint}
```
Inbal-Tish: This should be passed from the ExpandableCalendar component, no?

Inbal-Tish:
@mymattcarroll Please see version 1.261.0. Thank you!

mymattcarroll:
@Inbal-Tish, Thank you for considering my Pull Request :)

In regards to 1.261.0, I don't see any see documentation on how to make the calendar accessible. Is that coming in a future update?

Inbal-Tish:
@mymattcarroll Hi. The calendar is accessible by default, just turn on the VoiceOver/TalkBack and check it out.

PR #902, opened 2019-08-06
Adding 'buttonTopPosition' prop to control button's y position

Inbal-Tish:


src/expandableCalendar/calendarProvider.js
```diff
@@ -31,6 +32,8 @@ class CalendarProvider extends Component {
     onMonthChange: PropTypes.func,
     /** whether to show the today button */
     showTodayButton: PropTypes.bool,
+    /** The button top position */
+    buttonTopPosition: PropTypes.number,
```
ethanshar: shouldn't it be bottomPosition? 
also I would name the prop `todayButton...Position`

src/expandableCalendar/calendarProvider.js
```diff
@@ -119,7 +122,7 @@ class CalendarProvider extends Component {
       const isToday = today === date;
       
       Animated.spring(this.state.buttonY, {
-        toValue: isToday ? 65 : -65,
+        toValue: isToday ? TOP_POSITION : -this.props.buttonTopPosition || -TOP_POSITION,
```
ethanshar: What happens when `isToday` is false that you need a different value?

Inbal-Tish: I need it to get out of the frame. It won't with every value so I'm keeping my own value

src/expandableCalendar/calendarProvider.js
Inbal-Tish: It the button's Y position. You think that's bottom?

PR #906, opened 2019-08-18
ExpandableCalendar - adjustments for landscape orientation

Inbal-Tish:


src/expandableCalendar/index.js
```diff
@@ -168,7 +168,7 @@ class ExpandableCalendar extends Component {
   /** Utils */
   getOpenHeight() {
     if (!this.props.horizontal) {
-      return commons.screenHeight;
+      return commons.screenHeight > commons.screenWidth ? commons.screenHeight : commons.screenWidth;
```
ethanshar: use Math.max here

Inbal-Tish: Done.

PR #913, opened 2019-08-29
Makes font weight dynamic

adamhinckley:


src/agenda/reservation-list/style.js
```diff
@@ -12,6 +12,7 @@ export default function styleConstructor(theme = {}) {
     dayNum: {
       fontSize: 28,
       fontWeight: '200',
```
adamhinckley: Line 14 needs to be deleted

Inbal-Tish: You should submit the PR with all the changes you suggest and set the default to the existing value.

Inbal-Tish:
@adamhinckley Hi. Please fix the duplicate key 'fontWeight' issue.

PR #914, opened 2019-08-29
Makes font weight dynamic

adamhinckley:


src/agenda/reservation-list/style.js
```diff
@@ -11,7 +11,7 @@ export default function styleConstructor(theme = {}) {
     },
     dayNum: {
       fontSize: 28,
-      fontWeight: '200',
+      fontWeight: appStyle.agendaDayFontWeight,
```
Inbal-Tish: Please set the default to 200

PR #921, opened 2019-09-03
Added agenda selectedDay state update on selected prop change.

jesuscc1993:


jesuscc1993:
@Inbal-Tish sorry to mention you but could someone please check this PR?
It's been a while since we opened it and it currently blocks one of our requirements.

src/agenda/index.js
```diff
@@ -216,13 +216,21 @@ export default class AgendaView extends Component {
...
-      this.setState({
-        firstResevationLoad: false
-      });
+  componentWillReceiveProps(nextProps) {
+    const nextState = {};
+    if (nextProps.items) {
+      nextState.firstResevationLoad = false;
+    }
+    if (
+      nextProps.selected &&
```
Inbal-Tish: What if it changed from being selected to not selected? No update?

jesuscc1993: It's been a while since I opened this PR so I'm not really in context. Unless I'm mistaken:

Changes in the optional ``selected`` trigger a change on the internal ``selectedDay`` (well, they actually don't and that's what the PR is trying to fix). This internal date is not optional so even if the prop changes to undefined we can't set the internal date to undefined (hence the check).

Inbal-Tish: What I mean is that you pass the condition only if `selected === true`. But what if it was true but now it's false?

jesuscc1993: ``selected``is **not** a boolean (I know it sounds like it is, I certainly wouldn't had named it like that). ``selected``is a date. What I am checking is that it has a value, as we cannot assign undefined or null to ``selectedDay``; we cannot unselect the date on the component (we get an exception otherwise).

jesuscc1993:
It seems this PR got closed as a side effect of deleting a dirty master branch on our repo so I will reopen it as soon as possible when I have time to do it. Apologies for the inconvenience.

Inbal-Tish:
@jesuscc1993 No worries, looking forward to reviewing your new PR.

PR #940, opened 2019-09-26
Move prop spreading to the end of SectionList to enable overriding it's props

debugginator:
I ran into a situation where I wanted to modify the default section header without hard copying the whole agendaList component to my project and was unable to override the method handler for it (renderSectionHeader).

### DIFF
--- a/src/expandableCalendar/agendaList.js
+++ b/src/expandableCalendar/agendaList.js
@@ -135,7 +135,6 @@ class AgendaList extends Component {
   render() {
     return (
       <SectionList
-        {...this.props}
         ref={this.list}
         keyExtractor={this.keyExtractor}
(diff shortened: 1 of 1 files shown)


src/expandableCalendar/agendaList.js
```diff
@@ -148,6 +147,7 @@ class AgendaList extends Component {
         onMomentumScrollEnd={this.onMomentumScrollEnd}
         // onScrollToIndexFailed={(info) => { console.warn('onScrollToIndexFailed info: ', info); }}
         // getItemLayout={this.getItemLayout} // onViewableItemsChanged is not updated when list scrolls!!!
+        {...this.props}
```
Inbal-Tish: This is risky. Not all props should be overridden by the user.

Issue #944, opened 2019-09-30
Disabled and DisableTouchEvent are not working as expected


jnrepo:
# Description

I'm using the Calendar component. Trying to disable dates using markedDates prop and `disabled: true` and `disableTouchEvent: true`. Was there any intention on making the disabled days not trigger? Looking at the code it doesn't seem like there are any conditionals there https://github.com/wix/react-native-calendars/blob/d26dfd4e8a735e8c691a65caf55b115915115dd4/src/calendar/day/period/index.js#L40-L42

## Expected Behavior

Should not be able to trigger onDayPressed for disabled days.

## Observed Behavior

onDayPressed gets triggered on days with the disabled prop.

## Environment

Please run these commands in the project folder and fill in their results:

* `npm ls react-native-calendars`:1.212.0
* `npm ls react-native`: 0.60.3

Also specify:

1. Phone/emulator/simulator & version: iPhoneX

## Reproducible Demo

```
                <Calendar
                  onDayPress={() => { console.log('SHOULD NOT WORK')}
                  monthFormat={'MMM yyyy'}
                  hideExtraDays={true}
                  firstDay={1}
                  hideDayNames={true}
                  markedDates={{
                    '2019-10-03': { disabled: true, disableTouchEvent: true }
                  }}
                  markingType={'period'}
                  />
```

jnrepo:
wrote a PR: https://github.com/wix/react-native-calendars/pull/945

Stas-Buzunko:
this issue was back in 1.11.0, i thought updating to latest 1.214 would fix it. BUT IT'S NOT.
weird. such a basic and mandatory functionality is missing.

Inbal-Tish:
The 'disable' marking is only to apply the 'textDisabledColor' theme style on the day text while the 'disableTouchEvent' marking is passed to the Day component's TouchableOpacity, meaning `disabled={marking.disableTouchEvent}`, to disable all interactions for this component. 
This snippet works for me:
`
<Calendar
          onDayPress={() => console.warn('onDayPress')}
          hideExtraDays={true}
          markedDates={{
            '2020-01-19': {disabled: true, disableTouchEvent: true}
          }}
          theme={{
            textDisabledColor: 'red'
          }}
        />
`

PR #945, opened 2019-09-30
use disableTouchEvent to prevent onDayPress

jnrepo:
https://github.com/wix/react-native-calendars/issues/944


Inbal-Tish:
@jnrepo Hi. I don't see the use of this condition on the onDayPress() since the 'disableTouchEvent` marking is passed to the Day's touchableOpacity component to disable all interactions. 
Please see my comment on issue #944 .

PR #946, opened 2019-10-01
Changed Calendar Offset

GunaShekar02:
The calendar offset which was 90 - (this.viewHeight / 2) previously is causing a lot of issues to many people by hiding the lower portion of the calendar's dates in week view. This problem will be eliminated by changing the offset to 100 - (this.viewHeight / 2).


Inbal-Tish:
@GunaShekar02 Hi. Please see version 1.230.0 with fix. Thanks

Issue #938, opened 2019-09-26
How to set style of all Saturday and Sunday days in the Calendar ?


MustaphaGhlissi:
Hi , 

How can I set the style of month weekends in the Calendar (All saturday and Sunday days  )

Regards !

Inbal-Tish:
@MustaphaGhlissi You can follow this PR #965

PR #949, opened 2019-10-04
Added support to mark saturdays and sundays.

singhpratik17:
Issue #938

Added props:
    `markSunday: PropTypes.bool`
    `markSaturday: PropTypes.bool`

Styling can be passed using:
`markedDates={{weekendStyle: {marked: true, dotColor: 'red'}, ...}`


Inbal-Tish:
@singhpratik17 Hi. Thanks for the PR. I would like you to take a look at this PR #965.
I think @Moreno97 has a more generic approach to the same feature that allows you to mark other days of the week and not just Saturday and Sunday. See if that can work for you.

PR #954, opened 2019-10-08
Update incorrectly named prop scrollingEnabled

lanechr:
Agenda is passing prop scrollingEnabled to CalendarList, which expects scrollEnabled. This fixes that


Inbal-Tish:
@lanechr Hi. This bug was fixed. Please see master. Thanks

PR #955, opened 2019-10-08
Fixed CalendarList not rendering when flex:1 is used

IjzerenHein:
Hi! This PR fixes a problem with the `<CalendarList>` component which does not layout correctly when a style like `flex: 1` is used. This is because the used parent View does not propagate the flex style correctly. It has been changed to a `React.Fragment`, which fixes the problem.

So in short, it fixes this:

```jsx
<CalendarList style={{flex: 1}} .../>
```

Interestingly, this problem did not show its head in <= RN59, but only after upgrading to RN61. Which may indicate that Yoga has become a bit stricter in RN60+.

cheers!

### DIFF
--- a/src/calendar-list/index.js
+++ b/src/calendar-list/index.js
@@ -286,7 +286,7 @@ class CalendarList extends Component {
 
   render() {
     return (
-      <View>
+      <React.Fragment>
         <FlatList
           onLayout={this.onLayout}
           ref={(c) => this.listView = c}
@@ -312,7 +312,7 @@ class CalendarList extends Component {
           scrollsToTop={this.props.scrollsToTop}
         />
         {this.renderStaticHeader()}
-      </View>
+      </React.Fragment>
     );
   }
 }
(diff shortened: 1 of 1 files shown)


Inbal-Tish:
@IjzerenHein Hi. What is the `style={{flex: 1}}` for? Also, please fix the lint issue (unused View import).

IjzerenHein:
Hi, I've fixed the unused import.

The `style={{flex: 1}}` style demonstrates the case that did not work. If you wanted to create a CalendarList that has the full height/width of the parent then you needed to pass flex:1 in. However this style was set on the inner View and not on the outer view, causing the CalendarList to have zero height.

Inbal-Tish:
I understand that CalendarList doesn't render with `style={{flex: 1}}`, but that's why there is `calendarWidth` and `calendarHeight` props where you can adjust the Calendar and CalendarList size. 
I really don't see the use for this fix at the current API.
Thanks!

RuairiOliver:
This is great thanks. Definitely useful rather than having to calculate what the height should be to fit in your layout.

Issue #900, opened 2019-08-02
Unable to resolve module `hoist-non-react-statics`...


mgfjxxiexiaolong:
![Simulator Screen Shot - iPhone Xʀ - 2019-08-02 at 08 32 32](https://user-images.githubusercontent.com/13596540/62335960-ad156a00-b500-11e9-871a-1ff7c8c6934e.png)

I run `npm install --save hoist-non-react-statics` to solve my problem. Just note that.

iiagodias:
thanks, it helped me.

Inbal-Tish:
Fix on version 1.225.0

PR #963, opened 2019-10-18
Fix error with missing dependency

Moreno97:
This PR fixes error on #900 
It also removes `react-native-navigation` dev dependency from the `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,8 @@
   "dependencies": {
     "lodash": "^4.0.0",
     "prop-types": "^15.5.10",
-    "xdate": "^0.8.0"
+    "xdate": "^0.8.0",
+    "hoist-non-react-statics": "3.3.0"
   },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -24,7 +24,8 @@
   "dependencies": {
     "lodash": "^4.0.0",
     "prop-types": "^15.5.10",
-    "xdate": "^0.8.0"
+    "xdate": "^0.8.0",
+    "hoist-non-react-statics": "3.3.0"
```
AlinDev: You have to deploy this fast! New installs fail without this dependency.

Inbal-Tish:
@Moreno97 Hi. This PR is no longer relevant since we moved to RN 61. I fixed the issue #900 in version 1.225.0 and I'm closing all related issues and PRs. Thank you!!!

PR #771, opened 2019-03-18
Agenda closed by default

IronTony:
Added a props on the Agenda to load it closed.
Based on this comment and code: https://github.com/wix/react-native-calendars/issues/309#issuecomment-375109475

PR #974, opened 2019-10-30
Added isDefaultViewCalendar prop.

puneetp0110:


Inbal-Tish:
@puneetp0110 Hi. Please explain the need for this prop. What is the use-case in hand?

Inbal-Tish:
Is this a duplication of PR #771 ?

PR #980, opened 2019-11-05
Added custom styles for left and right arrows

TchernyavskyDaniil:
Hi, after searching some custom styles for right or left icon, I understand that there is no such possibility. Am I mistaken? 

So, I would like to suggest this method for adding custom styles for left and right arrow components

for example:

```
    arrow: {
      width: 100,
    },
    rightArrow: {
      some styles,
      alignItems: 'flex-end',
    },
    leftArrow: {
      some styles,
      alignItems: 'flex-start',
    }
  },
```

@ethanshar ?

### DIFF
--- a/src/calendar/header/index.js
+++ b/src/calendar/header/index.js
@@ -10,7 +10,7 @@ import {CHANGE_MONTH_LEFT_ARROW, CHANGE_MONTH_RIGHT_ARROW} from '../../testIDs';
 
 class CalendarHeader extends Component {
   static displayName = 'IGNORE';
-  
+
   static propTypes = {
     theme: PropTypes.object,
     hideArrows: PropTypes.bool,
@@ -94,7 +94,7 @@ class CalendarHeader extends Component {
       leftArrow = (
         <TouchableOpacity
           onPress={this.onPressLeft}
-          style={this.style.arrow}
+          style={{...this.style.arrow, ...this.style.leftArrow}}
           hitSlop={{left: 20, right: 20, top: 20, bottom: 20}}
           testID={testID ? `${CHANGE_MONTH_LEFT_ARROW}-${testID}`: CHANGE_MONTH_LEFT_ARROW}
         >
@@ -109,7 +109,7 @@ class CalendarHeader extends Component {
       rightArrow = (
         <TouchableOpacity
           onPress={this.onPressRight}
-          style={this.style.arrow}
+          style={{...this.style.arrow, ...this.style.rightArrow}}
           hitSlop={{left: 20, right: 20, top: 20, bottom: 20}}
           testID={testID ? `${CHANGE_MONTH_RIGHT_ARROW}-${testID}`: CHANGE_MONTH_RIGHT_ARROW}
         >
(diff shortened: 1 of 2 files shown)


src/calendar/header/index.js
```diff
@@ -94,7 +94,7 @@ class CalendarHeader extends Component {
       leftArrow = (
         <TouchableOpacity
           onPress={this.onPressLeft}
-          style={this.style.arrow}
+          style={{...this.style.arrow, ...this.style.leftArrow}}
```
Inbal-Tish: Why are you spreading the style objects?

TchernyavskyDaniil: Because we have two objects. So, we need to create one object for styles. Am i doing that wrong?

Inbal-Tish: You don't need to spread the objects.

TchernyavskyDaniil: Uh, why? Can you give me example, please.

Inbal-Tish:
@TchernyavskyDaniil Please try using 'arrowStyle' in the `theme` prop object to adjust the style of the arrows and let me know if that is working out for you. I feel that there is no need to set separate styles to the left and right arrow but to adjust them as a unit.

src/calendar/header/index.js
Inbal-Tish: Passing styles as an array: `style={[this.style.arrow, this.style.leftArrow]}`

TchernyavskyDaniil:
@Inbal-Tish But how? I coudn't find example to set seperate styles to the left and right arrow from theme prop :(
Calendar have renderArrow prop, but it's just a child for TouchableOpacity where I need use unique style for the left or right arrow.
Am i some prop or setting missing from doc?

Inbal-Tish:
There is no separate style for left/right arrows, there is an `arrowStyle` prop. You pass it in the `theme` prop object. Example:
`
theme={{arrowStyle: {padding: 30}}}
`
This is the `appStyle.arrowStyle` spread in the `stylesheet.calendar.header.arrow` object.

Inbal-Tish:
Look at 1src/style.js` to see other styles you can pass in the `theme` prop object.

TchernyavskyDaniil:
> Look at 1src/style.js`to see other styles you can pass in the`theme` prop object.

Yes, you are right. So, I decided do uniques styles for the left and right arrows, but not required.
Ok, i fixed my mistake (spread). So, what next steps?

Inbal-Tish:
I think your fix is unnecessary so I'm inclined to close this pr unless you present a real case where you must use separate styles for left/right arrows that can't be treated with `arrowStyle` nor `renderArrow` props.

TchernyavskyDaniil:
@Inbal-Tish ```arrowStyle```  but this prop for the left and right arrow, not for some one of this.
```renderArrow``` for a child TouchableOpacity.


In my app, by design, need set width for arrows like 100px
``` 
arrow: {
      width: 100,
    },
```

ok, them I need to add ```alignItems: flex-end or flex-start```, because without this styles
The arrows are not around the edges
So, like this:
```
 arrow: {
      width: 100,
    },
    rightArrow: {
      paddingRight: 5,
      alignItems: 'flex-end',
    },
    leftArrow: {
      paddingLeft: 5,
      alignItems: 'flex-start',
    },
```

Inbal-Tish:
You can do that in the `arrowStyle`, for example instead of using 'paddingLeft' and 'paddingRight' in each style, use 'paddingHorizontal' for the arrow style. Please adjust your code accordingly, adding additional props/styles is not the right way to tackle this issue, as it creates additional complications and duplications of the code.

TchernyavskyDaniil:
@Inbal-Tish But what about ```alignItems: flex-end or flex-start``` for the left or right arrow ? I can create left arrow and right arrow prop and passing to header style.

Inbal-Tish:
Please review layout props: https://facebook.github.io/react-native/docs/layout-props

TchernyavskyDaniil:
@Inbal-Tish Thanks, but i needed this styles for TouchableOpacity.  Now Calendar didn't provides that

PR #981, opened 2019-11-07
Fixed: getDateMarking returns empty array even if no marking specified

andrey-tsaplin:
Fixed: getDateMarking returns empty array even if no marking specified for day, causing day/period to have borderRadius style set even on not marked days, i.e. on all days.

Empty array treated as true here:
https://github.com/wix/react-native-calendars/blob/master/src/calendar/day/period/index.js#L131
makes all days to be rounded event if not selected. It is not visible at all but it is a root cause of another bug in "period" marking type on android: when initially deselected day becomes selected it is not rounded:

Initially unmarked and marked after render
Screenshot from 2019-11-07 21-09-39

Initially marked
Screenshot from 2019-11-07 21-34-40

Now rounding works if marker was initially unmarked.

### DIFF
--- a/src/calendar/index.js
+++ b/src/calendar/index.js
@@ -18,7 +18,6 @@ import {SELECT_DATE_SLOT} from '../testIDs';
 
 //Fallback when RN version is < 0.44
 const viewPropTypes = ViewPropTypes || View.propTypes;
-const EmptyArray = [];
 
 /**
  * @description: Calendar component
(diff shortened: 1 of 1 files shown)


src/calendar/index.js
```diff
@@ -214,11 +213,14 @@ class Calendar extends Component {
       return false;
     }
 
-    const dates = this.props.markedDates[day.toString('yyyy-MM-dd')] || EmptyArray;
-    if (dates.length || dates) {
```
andrey-tsaplin: This expression was always true

Inbal-Tish:
@andrey-tsaplin Hi. Can you please send me a code snippet for this use case so I can see the issue? Is it an Android only issue?

PR #990, opened 2019-11-13
Add function to close Agenda calendar manually

RZulfikri:
Hi, I'm facing issues once I navigate the screen. In my case, I put `Agenda` in the bottom tab navigation screen menu.
1. once calendar opened -> then I select another tab menu -> calendar still opened
2. once calendar opened -> then I navigate to another screen (not root tab menu) -> calendar closed, knob hide and cannot scroll/open calendar.

how to fix:
1. I adding `setRef` props to get `Agenda` ref.
2. I adding `closeCalendar()` function to close calendar manually from ref

how to use :
```javascript
class AgendaScreen extends Component {
  ....
  agendaRef = undefined

  closeCalendar = () => {
     if (this.agendaRef) {
        this.agendaRef.closeCalendar()
     }
  }

  render() {
     return (
        <View>
        <Agenda
         setRef={calendarRef => this.agendaRef = calendarRef}
          ....
        />
        </View>
     )
  }
}
```


src/agenda/index.js
```diff
@@ -11,8 +11,8 @@ import styleConstructor from './style';
 import {VelocityTracker} from '../input';
 
 
-const HEADER_HEIGHT = 104;
-const KNOB_HEIGHT = 24;
+const HEADER_HEIGHT = 115;
+const KNOB_HEIGHT = 15;
```
Inbal-Tish: Why are you changing these values?

src/agenda/index.js
```diff
@@ -115,6 +116,7 @@ export default class AgendaView extends Component {
     this.generateMarkings = this.generateMarkings.bind(this);
     this.knobTracker = new VelocityTracker();
     this.state.scrollY.addListener(({value}) => this.knobTracker.add(value));
+    this.closeCalendar = this.closeCalendar.bind(this)
```
Inbal-Tish: missing semicolon

src/agenda/index.js
```diff
@@ -207,6 +209,14 @@ export default class AgendaView extends Component {
     }
   }
 
+  componentDidMount() {
+    const {setRef} = this.props
+    /** Use to pass ref */
+    if (typeof setRef === 'function') {
+      setRef(this)
```
Inbal-Tish: missing semicolon

Inbal-Tish:
@RZulfikri Please resolve conflicts and fix lint errors so I can review.

src/agenda/index.js
```diff
@@ -1,14 +1,14 @@
-import React, {Component} from 'react';
-import {Text, View, Dimensions, Animated, ViewPropTypes} from 'react-native';
+import React, { Component } from 'react';
```
Inbal-Tish: Please remove all formatting changes

RZulfikri:
@Inbal-Tish done.

src/agenda/index.js
```diff
@@ -27,6 +27,7 @@ export default class AgendaView extends Component {
   static displayName = 'Agenda';
 
   static propTypes = {
+    setRef: PropTypes.func,
```
Inbal-Tish: I don't see the use of this 'setRef' prop. Why not pass ref to the Agenda component and use it to invoke the closeCalendar() method?

src/agenda/reservation-list/index.js
```diff
@@ -190,6 +190,8 @@ class ReactComp extends Component {
       return (
         <ActivityIndicator style={{marginTop: 80}} color={this.props.theme && this.props.theme.indicatorColor} />
       );
+    } else if (this.props.reservations[this.props.selectedDay.toString('yyyy-MM-dd')] && this.props.reservations[this.props.selectedDay.toString('yyyy-MM-dd')].length === 0) {
```
Inbal-Tish: Please explain. How is this has to do with the closeCalendar() functionality?

src/agenda/index.js
RZulfikri: `setRef` used in `componentDidmount`. I pass `this` to setRef so parent class can access `closeCallendar` method of this class.

src/agenda/reservation-list/index.js
RZulfikri: whoops, this line seems not related to closeCalendar method. Actually I'm adding this for my case, to add more validation for date items that have empty data. So if selected date is empty, it render empty screen instead of render list with empty items.

src/agenda/index.js
Inbal-Tish: Again, why not use the ref? pass `ref={calendarRef => this.agendaRef = calendarRef}` to Agenda component and use it the same way you do with `setRef`.

RZulfikri: because ref not working.

RZulfikri: my bad. to be honest, I forgot why I put setRef instead of ref. 😅
But, I think with setRef we can set which methods that we want to pass to the parent instead of pass all methods (but, I haven't added this to the code also).

Inbal-Tish: And that's one of the reasons I don't like the `setRef`...

src/agenda/reservation-list/index.js
Inbal-Tish: Then remove it please

Inbal-Tish:
@RZulfikri Please see PR #854 as well.

PR #996, opened 2019-11-23
Support option; onlySelectedDataList

Alghost:
When I use your calendar as agenda, I need the list to show entries about the selected date only.

So I add the option onlySelectedDataList to Agenda.

I'd like to merge it.

if I write the codes using wrong way or you don't like it,
Could you fix this?

Thanks in advance!

### DIFF
--- a/src/agenda/index.js
+++ b/src/agenda/index.js
@@ -82,7 +82,9 @@ export default class AgendaView extends Component {
     /** Set this true while waiting for new data from a refresh. */
     refreshing: PropTypes.bool,
     /** Display loading indicador. Default = false */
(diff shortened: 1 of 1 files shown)


src/agenda/index.js
```diff
@@ -280,6 +282,18 @@ export default class AgendaView extends Component {
   }
 
   renderReservations() {
+    let items = this.props.items;
```
Inbal-Tish: Items are not loading at all...

Vasault:
this should be a MUST, i've been struggling to get this working, just realize it wasn't supported in the first place

PR #997, opened 2019-11-25
Remove pods from source control

friederbluemle:
This PR addresses the following issues:

- Updates `.gitignore` to match the default one from React Native
- Adds the missing dependency lock files (they [should be added](https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/))
- Removes the entire `ios/Pods` directory

A note on `ios/Pods`: The folder was already added to `.gitignore` in 62d9576d12e052da3fefb1f03a286214332f1700), however many Pod files were committed to source control in ef621a555d55dc30351b484c08b378d72f932fe9 and dc9c2d8c4cb06f37f6f08ff1e61443c6be2fc9eb, which unfortunately increased the repository size by >15MB (permanently).


friederbluemle:
@Inbal-Tish - Please review when you have a minute! Thanks 🙏

friederbluemle:
A commit authored `Dec 15` with identical content to the original `Remove pods` commit was pushed to `master` (670be3de9d98ac311f2b5a36059b39ddafb0b0a3). The rebased remainder of this PR now only contains a few minor cleanups to `.gitignore` and adds a `yarn.lock` file.

.gitignore
```diff
@@ -1,11 +1,3 @@
-node_modules
```
Inbal-Tish: Why are you removing all these from the ignore file?

.gitignore
```diff
@@ -28,7 +20,6 @@ DerivedData
 *.hmap
 *.ipa
 *.xcuserstate
-# project.xcworkspace
```
Inbal-Tish: this one should be added

example/.gitignore
```diff
@@ -20,7 +20,6 @@ DerivedData
 *.hmap
 *.ipa
 *.xcuserstate
-project.xcworkspace
```
Inbal-Tish: why remove this?

Inbal-Tish: `yarn-lock` should be added

.gitignore
friederbluemle: If you check the rest of the `.gitignore` file, you'll see that `node_modules/` is _already_ listed (in the default location for `.gitignore`). There is no need for this duplicate.

.gitignore
friederbluemle: What do you mean by this? It's a comment that I removed. The file is more in line with the React Native default. Any specific requirements for this project?

example/.gitignore
friederbluemle: With `yarn-lock` you mean `yarn.lock`? `yarn.lock` records a _known_ dependency tree for this example project (exact dependency versions that the project is guaranteed to work with). Please check this article for more info:
https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/

friederbluemle:
Rebased onto latest `master` branch.

@Inbal-Tish - Could you please have another look at this?

PR #1001, opened 2019-12-01
Replace of componentWillReceiveProps method on Agenda, Calendar, Calendar-List

AgustinJimenez:
According to react warning https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

### DIFF
--- a/src/agenda/index.js
+++ b/src/agenda/index.js
@@ -216,14 +216,14 @@ export default class AgendaView extends Component {
     this._isMounted = false;
   }
 
-  componentWillReceiveProps(props) {
+  static getDerivedStateFromProps(props, state) {
     if (props.items) {
-      this.setState({
-        firstResevationLoad: false
-      });
+      state.firstResevationLoad = false;
     } else {
       this.loadReservations(props);
     }
+
+    return state;
   }
 
   enableCalendarScrolling() {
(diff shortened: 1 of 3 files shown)


src/calendar-list/index.js
```diff
@@ -139,7 +139,7 @@ class CalendarList extends Component {
     this.listView.scrollToOffset({offset: scrollAmount, animated: false});
   }
 
-  componentWillReceiveProps(props) {
+  UNSAFE_componentWillReceiveProps(props) {
```
oliver-ni: probably not a good idea

AgustinJimenez: why?

src/agenda/index.js
```diff
@@ -216,14 +216,14 @@ export default class AgendaView extends Component {
...
 
-  componentWillReceiveProps(props) {
+  static getDerivedStateFromProps(props, state) {
     if (props.items) {
-      this.setState({
-        firstResevationLoad: false
-      });
+      state.firstResevationLoad = false;
     } else {
       this.loadReservations(props);
```
Inbal-Tish, line 223: There is no access to 'this' in a static method

src/calendar-list/index.js
Inbal-Tish: This method should be replaced as it's being deprecated soon

Inbal-Tish:
@AgustinJimenez Thank you for the PR. Please check out PR #1356. Thanks!

Issue #80, opened 2017-07-04
Agenda: allow to scroll past the initial selected day?


slorber:
Hi,

Thanks for this great project

Let's take the [example agenda](https://github.com/wix/react-native-calendars/blob/master/example/src/screens/agenda.js) that is [published on expo](https://expo.io/@community/react-native-calendars-example)

The initial day is `selected={'2017-05-16'}`

The loading function is:

```javascript
  loadItems(day) {
    setTimeout(() => {
      for (let i = -15; i < 85; i++) {
        const time = day.timestamp + i * 24 * 60 * 60 * 1000;
        // load items for that time
        // ...
      }
    }
```

To my understanding, the purpose of this function is to load new items data as the user is changing date or scrolling. It is also triggered initially on initial day.

In this demo, we can see the purpose of loading next 85 days, and previous 15 days.

Yet, even if 15 days before the initial day `selected={'2017-05-16'}` have been loaded in state, the UI does not allow to scroll back in time. The only way to browse items of a past day is to open the knob and select a date. In my app I don't want the knob so it means there's no easy way to access the past items.

I don't know what could be a good solution to this problem. I guess rendering items lazily at the top after mount would mess-up with scroll position. For me a good enough solution would be that every items provided on mount could be rendered initially, and for otherwise keep current behavior.

What do you think?

tautvilas:
Hi, ideally I wanted to implement full scrolling backwards and forward, however there are RN limitations with such solution. There were some thoughts to make available past scrolling for preloaded period, but in order of consistency scrolling always stops on selected calendar date. Maybe this could be improved.

dongawli:
I am too facing this issue. I've already loaded more than 500 items and still user needs to select date using knob to see the previous items. Any timeline for this enhancement? 
Thank you.

dborstelmann:
Would really love this enhancement, please let us know if this is in the pipeline!

francescjimenez:
Yes, we would also be very useful for our application

tautvilas:
Sorry, but there are no short-term plans to solve this. There is some chance that it could be fixed this year though.

fescate:
+1

smkhalsa:
+1

mirrorshades:
+1

asantos00:
+1

asantos00:
Guys, I'm thinking about actually implementing this improvement, that's why I'm starting this discussion.

[...]

PR #1003, opened 2019-12-03
Implement Infinite Past Scrolling for Agenda

CHOIMINSEOK:
Implement Infinite Scrolling for Past with enabling props in Agenda. 

It works like charms. 

Refer to  @saigyouyou 's thread, #80

ezgif com-resize


CHOIMINSEOK:
Seems to improve the feature independently for a while. At least until the answer comes from maintainers.

PR #1015, opened 2019-12-22
Feat/expandable week scroll android fix

ethanshar:
I added the fix. please take a look.
 
In general there's still a very minor issue in both platforms, when u scroll really fast you might get stuck because the FlatList didn't got a chance to scroll back to the middle.
(on Android it's a little more noticeable) 

Another issue I noticed, not sure if it was there before, when I scroll a week, the first day doesn't always get marked. (only the edge pages). Possible I created this issue, I assume it's fixable. 

anyway, please go over and let me know if there're other issues, it would never be perfect on Android, just because it's Android, but at least it doesn't have the flicker.

### DIFF
--- a/src/expandableCalendar/weekCalendar.js
+++ b/src/expandableCalendar/weekCalendar.js
@@ -8,16 +8,15 @@ import styleConstructor from './style';
 import CalendarList from '../calendar-list';
 import Week from '../expandableCalendar/week';
 
-
 const commons = require('./commons');
 const UPDATE_SOURCES = commons.UPDATE_SOURCES;
-const NUMBER_OF_PAGES = 1; // must be a positive number
+const NUMBER_OF_PAGES = 2; // must be a positive number
 
 class WeekCalendar extends Component {
   static propTypes = {
     ...CalendarList.propTypes,
     // the current date
-    current: PropTypes.any
+    current: PropTypes.any,
   };
 
   constructor(props) {
@@ -25,19 +24,24 @@ class WeekCalendar extends Component {
     this.style = styleConstructor(props.theme);
 
     this.viewabilityConfig = {
-      itemVisiblePercentThreshold: 90
+      itemVisiblePercentThreshold: 90,
     };
     this.visibleItem = undefined;
-    this.items = this.getDatesArray();
+    // this.items = this.getDatesArray();
     this.list = React.createRef();
+    this.page = NUMBER_OF_PAGES;
+
+    this.state = {
+      items: this.getDatesArray(),
+    };
   }
 
-  componentDidUpdate() {
-    if (this.props.context.updateSource === UPDATE_SOURCES.WEEK_SCROLL) {
-      // to avoid new items render from changing the visible week
-      this.list.current.scrollToIndex({animated: false, index: NUMBER_OF_PAGES});
-    }
-  }
+  // componentDidUpdate() {
+  //   if (this.props.context.updateSource === UPDATE_SOURCES.WEEK_SCROLL) {
+  //     // to avoid new items render from changing the visible week
+  //     // this.list.current.scrollToIndex({animated: false, index: NUMBER_OF_PAGES});
+  //   }
+  // }
 
   getDatesArray() {
     const array = [];
@@ -48,7 +52,7 @@ class WeekCalendar extends Component {
     return array;
   }
 
-  getDate(index) {
+  getDate(weekIndex) {
     const {current, firstDay} = this.props;
     const d = XDate(current);
     // get the first day of the week as date (for the on scroll mark)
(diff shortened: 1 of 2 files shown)


src/expandableCalendar/weekCalendar.js
```diff
@@ -8,37 +8,41 @@ import styleConstructor from './style';
...
     // the current date
-    current: PropTypes.any
+    current: PropTypes.any,
   };
 
   constructor(props) {
     super(props);
     this.style = styleConstructor(props.theme);
 
     this.viewabilityConfig = {
```
Inbal-Tish: This one should be removed as well

src/expandableCalendar/weekCalendar.js
```diff
@@ -8,37 +8,41 @@ import styleConstructor from './style';
...
     this.list = React.createRef();
-  }
+    this.page = NUMBER_OF_PAGES;
 
-  componentDidUpdate() {
-    if (this.props.context.updateSource === UPDATE_SOURCES.WEEK_SCROLL) {
-      // to avoid new items render from changing the visible week
-      this.list.current.scrollToIndex({animated: false, index: NUMBER_OF_PAGES});
-    }
+    this.state = {
```
Inbal-Tish: You have a c'tor, why not place the state in it?

src/expandableCalendar/weekCalendar.js
```diff
@@ -57,62 +61,133 @@ class WeekCalendar extends Component {
...
   }
 
-  onViewableItemsChanged = ({viewableItems}) => {
-    // NOTE: gets called only on week scroll (so might be un-synced with displayed week) !!!
-    if (viewableItems.length > 0) {
-      const viweableItem = _.first(viewableItems);
-      if (viweableItem.isViewable && this.visibleItem !== viweableItem.item) {
-        if (this.visibleItem) { // to avoid invoke on first init
-          _.invoke(this.props.context, 'setDate', viweableItem.item, UPDATE_SOURCES.WEEK_SCROLL); 
+  onScroll = ({
```
Inbal-Tish: You prettier is breaking the lines where it shouldn't...

src/expandableCalendar/weekCalendar.js
```diff
@@ -57,62 +61,133 @@ class WeekCalendar extends Component {
...
+    nativeEvent: {
+      contentOffset: {x},
+    },
+  }) => {
+    const newPage = Math.round(x / commons.screenWidth);
+    if (this.page !== newPage) {
+      const {items} = this.state;
+      this.page = newPage;
+
+      _.invoke(
```
Inbal-Tish: Also unnecessary line break

src/expandableCalendar/weekCalendar.js
```diff
@@ -57,62 +61,133 @@ class WeekCalendar extends Component {
...
+        'setDate',
+        items[this.page],
+        UPDATE_SOURCES.WEEK_SCROLL,
+      );
+
+      if (this.page === items.length - 1) {
+        for (let i = 0; i <= NUMBER_OF_PAGES; i++) {
+          items[i] = items[i + NUMBER_OF_PAGES];
+        }
+        this.setState({items: [...items]});
```
Inbal-Tish: If you take out this setState() from the conditions it will fix the bug and mark every week's first day.

src/expandableCalendar/weekCalendar.js
```diff
@@ -57,62 +61,133 @@ class WeekCalendar extends Component {
...
+      } else if (this.page === 0) {
+
+        for (let i = NUMBER_OF_PAGES; i < items.length; i++) {
+          items[i] = items[i - NUMBER_OF_PAGES];
         }
-        this.visibleItem = viweableItem.item;
+
+        this.setState({items: [...items]});
```
Inbal-Tish: This too

src/expandableCalendar/weekCalendar.js
```diff
@@ -57,62 +61,133 @@ class WeekCalendar extends Component {
...
-  }
+  };
+
+  onMomentumScrollEnd = () => {
+    const {items} = this.state;
+    const isFirstPage = this.page === 0;
+    const isLastPage = this.page === items.length - 1;
+
+    if (isFirstPage || isLastPage) {
+      this.list.current.scrollToIndex({
```
Inbal-Tish: Also unnecessary line break

src/expandableCalendar/weekCalendar.js
```diff
@@ -57,62 +61,133 @@ class WeekCalendar extends Component {
...
+        animated: false,
+        index: NUMBER_OF_PAGES,
+      });
+      this.page = NUMBER_OF_PAGES;
+      const newWeekArray = this.getDatesArray();
+
+      if (isLastPage) {
+
```
Inbal-Tish: Unnecessary empty line

src/expandableCalendar/weekCalendar.js
```diff
@@ -8,37 +8,41 @@ import styleConstructor from './style';
...
     this.viewabilityConfig = {
-      itemVisiblePercentThreshold: 90
+      itemVisiblePercentThreshold: 90,
     };
     this.visibleItem = undefined;
-    this.items = this.getDatesArray();
+    // this.items = this.getDatesArray();
     this.list = React.createRef();
-  }
+    this.page = NUMBER_OF_PAGES;
```
Inbal-Tish: I will place it in the c'tor as well

src/expandableCalendar/weekCalendar.js
ethanshar: used `extraData` instead.

src/expandableCalendar/weekCalendar.js
ethanshar: used `extraData` instead.

src/expandableCalendar/weekCalendar.js
ethanshar: fixed

src/expandableCalendar/weekCalendar.js
ethanshar: fixed

src/expandableCalendar/weekCalendar.js
ethanshar: removed

src/expandableCalendar/weekCalendar.js
ethanshar: anyway, fixed

src/expandableCalendar/weekCalendar.js
Inbal-Tish: Sorry, I probably missed it

src/expandableCalendar/weekCalendar.js
Inbal-Tish: Cool

src/expandableCalendar/weekCalendar.js
Inbal-Tish: I don't. Never got around to it

src/expandableCalendar/weekCalendar.js
Inbal-Tish: Yes. I'll add it to the guild week tasks

Inbal-Tish:
@ethanshar There is still one issue - the arrows should scroll the week calendar as well. Do you want me to merge this PR and keep on working on it?

PR #1023, opened 2020-01-06
Test id for calendar knob

BfutureP:


src/agenda/index.js
```diff
@@ -118,7 +120,7 @@ export default class AgendaView extends Component {
   }
 
   calendarOffset() {
-    return 90 - (this.viewHeight / 2);
+    return 96 - (this.viewHeight / 2);
```
Inbal-Tish: Why did you change this? It's not part of testID issue...

PR #1024, opened 2020-01-06
added feature to have custom header, custom week names, custom markin…

trickyc0d3r:
…g of custom week days


example/android/.settings/org.eclipse.buildship.core.prefs
```diff
@@ -0,0 +1,2 @@
+connection.project.dir=
+eclipse.preferences.version=1
```
friederbluemle: Please remove this file (as well as `example/android/.project`) - They should not be in source control.

trickyc0d3r: oh, my bad. ok. sorry, I don't know why they appear every-time

src/calendar/index.js
```diff
@@ -272,6 +272,7 @@ class Calendar extends Component {
           weekNumbers={this.props.showWeekNumbers}
           onPressArrowLeft={this.props.onPressArrowLeft}
           onPressArrowRight={this.props.onPressArrowRight}
+          {...this.props}
```
Inbal-Tish: This is dangerous as you let the user override all props sent to CalendarHeader.

src/calendar/header/index.js
```diff
@@ -130,7 +130,7 @@ class CalendarHeader extends Component {
 
     return (
       <View style={this.props.style}>
-        <View style={this.style.header}>
+        {this.props.customHeader ? this.props.customHeader() : <View style={this.style.header}>
```
Inbal-Tish: I would prefer a check if 'customHeader' is even a function, or safely invoke using lodash

src/calendar/header/index.js
```diff
@@ -87,7 +87,7 @@ class CalendarHeader extends Component {
   render() {
     let leftArrow = <View />;
     let rightArrow = <View />;
-    let weekDaysNames = weekDayNames(this.props.firstDay);
+    let weekDaysNames = this.props.customDays ? this.props.customDays: weekDayNames(this.props.firstDay);
```
Inbal-Tish: The prop 'customDays' is not declared in the propTypes

example/android/.settings/org.eclipse.buildship.core.prefs
friederbluemle: The reason they (re-)appear is a buggy Java plugin for VS Code. Unless you heavily rely on it, the recommendation is to uninstall it.
Also see: https://github.com/redhat-developer/vscode-java/issues/618

Issue #874, opened 2019-07-15
recieveTouches RCTEventEmitter height NaN


iOSUser110:
## Description

First time on tap of that expandable line tap , giving error : Error calling RCTEventEmitter.recieveTouches 
Unhandled Js Exceptions 1097 : RCTView {height:"<<NaN>>} is not usable as a nativemethod argument 

## Expected Behavior

On tap of that hosrizontal bar for expansion it should expand normally , it is happening only first time 

## Observed Behavior

screen come with the below error for the first time 

```
Error calling RCTEventEmitter.recieveTouches 
Unhandled Js Exceptions 1097 : RCTView {height:"<<NaN>>} is not usable as a nativemethod argument 
```

## Environment
xcode - 10.2.1
react native - 0.60.0
react-native-calendars - 1.198.0

Phone/emulator/simulator & version:
iPhone 7
iOS 11.4.1

habibiazmi123:
I also have a problem, someone can help ?

iOSUser110:
do we need to pass height explicitly ?? is there any workaround ?
Please anybody ?

iOSUser110:
Invariant Violation: Invariant Violation: Invariant Violation: [1097,"RCTView",{"height":"<<NaN>>"}] is not usable as a native method argument


This error is located at:
    in RCTView (at View.js:45)
    in View (at createAnimatedComponent.js:151)
    in AnimatedComponent (at expandableCalendar/index.js:465)
    in RCTView (at View.js:45)
    in View (at expandableCalendar/index.js:464)

mtieltjes:
I boiled it down to this:

In the constructor the [`_wrapperStyles`](https://github.com/wix/react-native-calendars/blob/375fd756d664da5e84fdcf5d9b18126c5f26924f/src/expandableCalendar/index.js#L84) gets initialized with an empty style object.

Next, in [`handlePanResponderEnd`](https://github.com/wix/react-native-calendars/blob/375fd756d664da5e84fdcf5d9b18126c5f26924f/src/expandableCalendar/index.js#L268) the height gets set to the height in the `wrapperStyles.style`, begin `undefined` at that point.

[...]

PR #1044, opened 2020-01-21
Fix undefined height in expandableCalendar

mtieltjes:
See my comment here: https://github.com/wix/react-native-calendars/issues/874#issuecomment-576696278

Whether it is the right fix, I don't know. But due to an (unexpected?) order in panResponder callbacks the height will be set to NaN, causing native issues.

if 'panResponderEnd' gets called before 'handlePanResponderMove', `this._height` will be set to `undefined`, which will lead to NaN's coming out of Math.[min|max] calls.

### DIFF
--- a/src/expandableCalendar/index.js
+++ b/src/expandableCalendar/index.js
@@ -81,7 +81,7 @@ class ExpandableCalendar extends Component {
     
     const startHeight = props.initialPosition === POSITIONS.CLOSED ? this.closedHeight : this.openHeight;
     this._height = startHeight;
-    this._wrapperStyles = {style: {}};
+    this._wrapperStyles = {style: {height: startHeight}};
     this._headerStyles = {style: {top: this.props.initialPosition === POSITIONS.CLOSED ? 0 : -HEADER_HEIGHT}};
     this._weekCalendarStyles = {style: {}};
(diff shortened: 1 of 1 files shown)


Inbal-Tish:
@mtieltjes Did you came across this scenario (that 'panResponderEnd' gets called before 'handlePanResponderMove')? Can you tell me how to reproduce it (like which platform / device / os / usage)?

mtieltjes:
I was not able to reproduce it with the example project or a new React Native project.

I am encountering this issue on Android only with the following setup:
Device: Pixel 3a Emulator
Platform: Android API 29
React Native setup (not Expo)

Using React Navigation as our navigation stack, with this code in a screen
```javascript
const Screen = () => {
  return (
    <CalendarProvider {...} >
      <ExpandableCalendar {...} />
      <OurCustomCalendarConsumer {...} />
    </CalendarProvider>
  );
}
```

We dont interact with the CalendarContext in our consumer, so that _should_ not be related. I am not sure if ReactNavigation might be interfering with things.
I'll try to gather some more information in the next few days.

Inbal-Tish:
@mtieltjes Keep me posted if you have more details. Thanks!

PR #865, opened 2019-07-03
Passing testID to static header

Inbal-Tish:
to defer from calendar header arrows' testIDs

PR #1058, opened 2020-02-11
[FIX] Adding test id in Calendar Component

michael-harley:
So I found out that **HeaderCalendar** Component is using `testID` from `this.props`

But... PropTypes didn't ask for a `testID` and **Calendar** Component didn't pass `testID` to **HeaderCalendar** Component.

So I made this PR to fix that.

Related PR: #865 #882

### DIFF
--- a/src/calendar/header/index.js
+++ b/src/calendar/header/index.js
@@ -25,7 +25,8 @@ class CalendarHeader extends Component {
     onPressArrowRight: PropTypes.func,
     disableArrowLeft: PropTypes.bool,
     disableArrowRight: PropTypes.bool,
-    webAriaLevel: PropTypes.number
+    webAriaLevel: PropTypes.number,
+    testID: PropTypes.string
   };
 
   static defaultProps = {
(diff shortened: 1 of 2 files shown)


src/calendar/header/index.js
```diff
@@ -25,7 +25,8 @@ class CalendarHeader extends Component {
     onPressArrowRight: PropTypes.func,
     disableArrowLeft: PropTypes.bool,
     disableArrowRight: PropTypes.bool,
-    webAriaLevel: PropTypes.number
+    webAriaLevel: PropTypes.number,
+    testID: PropTypes.string
```
Inbal-Tish: `testID` is an RN prop, no need to declare it

src/calendar/index.js
```diff
@@ -84,7 +84,9 @@ class Calendar extends Component {
     /** Style passed to the header */
     headerStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array]),
     /** Provide aria-level for calendar heading for proper accessibility when used with web (react-native-web) */
-    webAriaLevel: PropTypes.number
+    webAriaLevel: PropTypes.number,
+    /** Provide Calendar Test ID. Default = undefined */
+    testID: PropTypes.string
```
Inbal-Tish: `testID` is an RN prop, no need to declare it

michael-harley:
I have removed the testID 🙇

PR #1063, opened 2020-02-17
Adding scroll events to the agenda list

HenryGorner:
Added the ability to catch scroll events of the agenda flatlist

Tested on iOS and Android.


guanw:
Shouldn't there be any unit test for this?

PR #1067, opened 2020-02-20
adjust weekday width

guanw:
before change:
image

after change: 
image

### DIFF
--- a/src/agenda/style.js
+++ b/src/agenda/style.js
@@ -30,7 +30,7 @@ export default function styleConstructor(theme = {}) {
       backgroundColor: appStyle.calendarBackground
     },
     weekday: {
-      width: 32,
+      width: 35,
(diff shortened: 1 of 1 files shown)


guanw:
@Inbal-Tish The calendar weekday doesn't seem to fit all days on single row without omitting some chars. 

After discussed with our design people, we have two ways to do this:
1. replace all weekday with single char (e.g Tue -> T)
2. style change to make it fit

I did the second one with this for now. Let me know what you think?

Inbal-Tish:
@guanw Can you please specify how to reproduce the bug? Is device/os-related? is it happens when you set `appStyle.textDayHeaderFontSize` or other font-related styles?

guanw:
You are right. Our code seems to update `textDayHeaderFontSize`. I will revert that and close this comment

Issue #1074, opened 2020-03-01
Header and montView not align when showWeekNumbers


TomCorvus:
Hi ✌️ 

## Description

When you set showWeekNumbers to true, the header and monthView are not aligned.

## Environment

Please run these commands in the project folder and fill in their results:

* `npm ls react-native-calendars`: react-native-calendars@1.263.0
* `npm ls react-native`: react-native@0.61.4

## Reproducible Demo

https://snack.expo.io/@tom.corvus/react-native-calendars-demo-3

PR #1077, opened 2020-03-02
fix week numbers alignment with header

TomCorvus:
Fix week numbers alignment in month view when showWeekNumbers is true. https://github.com/wix/react-native-calendars/issues/1074


chenei:
Thanks

PR #1079, opened 2020-03-03
Added optional formatMonth and formatNumber functions

jesuscc1993:
Added optional formatMonth and formatNumber functions to be used for adding support for Arabic numbers. There is no native way to do this conversion so this was the only solution I could think of.

Preview:
preview

### DIFF
--- a/src/calendar/day/basic/index.js
+++ b/src/calendar/day/basic/index.js
@@ -94,9 +94,7 @@ class Day extends Component {
         accessibilityRole={isDisabled ? undefined : 'button'}
         accessibilityLabel={this.props.accessibilityLabel}
       >
-        <Text allowFontScaling={false} style={textStyle}>
-          {String(this.props.formatNumber ? this.props.formatNumber(this.props.children) : this.props.children)}
-        </Text>
+        <Text allowFontScaling={false} style={textStyle}>{String(this.props.formatNumber ? this.props.formatNumber(this.props.children) : this.props.children)}</Text>
         {dot}
       </TouchableOpacity>
     );
--- a/src/calendar/day/basic/index.js
+++ b/src/calendar/day/basic/index.js
@@ -8,10 +8,12 @@ import styleConstructor from './style';
 
 class Day extends Component {
   static displayName = 'IGNORE';
-  
+
   static propTypes = {
     // TODO: disabled props should be removed
     state: PropTypes.oneOf(['disabled', 'today', '']),
+    // Optional function to format displayed numbers
+    formatNumber: PropTypes.func,
     // Specify theme properties to override specific styles for calendar parts. Default = {}
     theme: PropTypes.object,
     marking: PropTypes.any,
@@ -44,16 +46,16 @@ class Day extends Component {
     const containerStyle = [this.style.base];
     const textStyle = [this.style.text];
     const dotStyle = [this.style.dot];
-    
+
     let marking = this.props.marking || {};
     if (marking && marking.constructor === Array && marking.length) {
       marking = {
         marking: true
       };
     }
-    
+
     const isDisabled = typeof marking.disabled !== 'undefined' ? marking.disabled : this.props.state === 'disabled';
-    
+
     let dot;
     if (marking.marked) {
       dotStyle.push(this.style.visibleDot);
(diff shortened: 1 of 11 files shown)


jesuscc1993:
@Inbal-Tish what about this PR?

src/calendar/day/basic/index.js
```diff
@@ -92,7 +94,7 @@ class Day extends Component {
         accessibilityRole={isDisabled ? undefined : 'button'}
         accessibilityLabel={this.props.accessibilityLabel}
       >
-        <Text allowFontScaling={false} style={textStyle}>{String(this.props.children)}</Text>
+        <Text allowFontScaling={false} style={textStyle}>{String(this.props.formatNumber ? this.props.formatNumber(this.props.children) : this.props.children)}</Text>
```
Inbal-Tish, line 97: Why duplicate this line through all the Day components instead of doing it once where you send the text in Calendar to the <DayComp> element?

Inbal-Tish:
@jesuscc1993 Hi. Thank you for this PR. This is a very important feature. The thing is that I don't see why you chose "formatting" the dates/titles instead of adding it to the Localization support. I believe that what you're trying to achieve for Arabic could be great for many other languages such as Japanese, thus it should be a part of the LocalConfig and not passed as a formatting prop.

jesuscc1993:
> @jesuscc1993 Hi. Thank you for this PR. This is a very important feature. The thing is that I don't see why you chose "formatting" the dates/titles instead of adding it to the Localization support. I believe that what you're trying to achieve for Arabic could be great for many other languages such as Japanese, thus it should be a part of the LocalConfig and not passed as a formatting prop.

What do you have in mind? Something like passing the numbers and let the component handle the formatting? Like this:
```
LocaleConfig.locales['ar'] = {
  ...
  numbers: ["٠", "١", "٢", "٣", "٤", "٥", "٦", "٧", "٨", "٩"]
};
```

Inbal-Tish:
Yes. Passing the numbers' translations

jesuscc1993:
> Yes. Passing the numbers' translations

LocaleConfig comes from XDate and is typed like this:
```
{
    monthNames? : string [];
    monthNamesShort?: string [];
    dayNames?: string[];
    dayNamesShort?: string [];
    amDesignator?: string;
    pmDesignator?: string;
}
```
...so we can't add numbers to LocaleConfig. I'm looking at the XDate module to see if it natively supports number formatting but it seems like it doesn't. Looks like we will probably need to pass an additional `localeNumbers` prop (or similarly called) instead.

Inbal-Tish:
Yes. Look what we did with 'today' string.

jesuscc1993:
> Yes. Look what we did with 'today' string.

I'm starting to think we can't achieve this without a formatting function, as the XDate library seems pretty limited on its localization. For instance, even if you setup the FR locale, if you run `date.toString('MMMM ddS yyyy', 'fr)` it will output `Décembre 10th 2020`, as the `st|nd|rd|th` string is *not* part of XDate's locale config. You can see it yourself on [this snippet](http://jsfiddle.net/jesuscc1993_work/7po1jdas/14/).

Inbal-Tish:
What if we work with English formats but only translate the displayed day number? Where we set the 'String(this.props.children)' to the Text component.

jesuscc1993:
I have the month and dates but I can't locate what renders the month placeholders when scrolling:
image

I looked up the reservations but those seem to only include the days. @Inbal-Tish do you have any tips on where to look for these?

Inbal-Tish:
I'm not sure what you mean... but you should translate the year as well, in Header/index.js 'renderHeader()' method.

jesuscc1993:
> I'm not sure what you mean... but you should translate the year as well, in Header/index.js 'renderHeader()' method.

I have those, @Inbal-Tish. But there's a different render when scrolling, as you can see in [this recording](https://i.imgur.com/DLXElAi.mp4)

Inbal-Tish:
That's in CalendatListItem line#101

jesuscc1993:
> That's in CalendatListItem line#101

Thanks 🙂 .
I was able to get everything done; I created a separate PR for it ([#1369](https://github.com/wix/react-native-calendars/pull/1369)). I referenced this one there but I will now close it.

PR #1089, opened 2020-03-11
fix: Correctly handle case where viewPropTypes is not exported from RN

nathsimpson:
When running build on a docz site, I came across this error....
```
Module not found: Can't resolve 'react-native-web/dist/exports/ViewPropTypes'
```
(full stack trace below)

viewPropTypes was removed in a recent version of React Native (0.6 maybe). After some investigation, it appears RN Calendars has some logic to fallback if the viewPropTypes isn't exported, but it seams that logic was not correct.

This PR ensures that RN Calendars correctly handles the case where viewPropTypes is not exported.

Full stack trace of error
```
✖  fatal     Error: ./node_modules/react-native-calendars/src/agenda/index.js 
Module not found: Can't resolve 'react-native-web/dist/exports/ViewPropTypes' in '/XXXXXX/node_modules/react-native-calendars/src/agenda'

./node_modules/react-native-calendars/src/calendar/index.js
Module not found: Can't resolve 'react-native-web/dist/exports/ViewPropTypes' in '/XXXXXXX/node_modules/react-native-calendars/src/calendar'
    at Promise (/XXXXX/node_modules/docz-core/dist/index.js:1306:21)
```

### DIFF
--- a/src/agenda/index.js
+++ b/src/agenda/index.js
@@ -1,5 +1,5 @@
 import React, {Component} from 'react';
-import {Text, View, Dimensions, Animated, ViewPropTypes} from 'react-native';
+import * as ReactNative from 'react-native';
 import PropTypes from 'prop-types';
 import XDate from 'xdate';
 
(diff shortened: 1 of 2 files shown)


nathsimpson:
@Inbal-Tish can we merge this PR in?

PR #1090, opened 2020-03-12
Detox setup (iOS)

Inbal-Tish:


ethanshar:
@Inbal-Tish 
How to I run the tests? 
I tried a command but got an error..

Inbal-Tish:
@ethanshar `detox test` didn't work?

PR #1110, opened 2020-04-08
withFootNote component is added

Gokaysim:
Screen Shot 2020-04-08 at 21 08 06
a new day component is added to achive above design

note: 'Dolu' means 'soldout' in Turkish :)

### DIFF
--- a/src/calendar/day/withFootNote/index.js
+++ b/src/calendar/day/withFootNote/index.js
@@ -67,8 +67,6 @@ class Day extends Component {
       textStyle.push(this.style.todayText);
     }
 
-    console.log(this.props.marking);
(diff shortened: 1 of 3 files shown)


chenei:
Hi @Gokaysim ,
Thanks for your PR!
Why do we need an all-new component just for that?
I saw that you copy-paste from another day component. can we please think about an API that can add this footNote?
Why can't we pass it through `markedDates`?

emilisb:
Hi @Gokaysim 

Any plans to finish with this?

Gokaysim:
Hi @emilisb 
At the moment I do not have time to change it. If I am not mistaken it was working

PR #1118, opened 2020-04-17
Updateing index.js

zameerascentic:
Removing getNode() to Support RN 62.0. This fixes warning "Calling `getNode()` on the ref of an Animated component is no longer necessary. You can now directly use the ref instead"


chenei:
Thanks! Already fixed.

PR #1123, opened 2020-04-23
Change border radius of calendar

samiede:
I honestly don't remember the changes I made because it's been so long ago, but it was only a few lines of code.

### DIFF
--- a/src/calendar-list/index.js
+++ b/src/calendar-list/index.js
@@ -47,7 +47,9 @@ class CalendarList extends Component {
     /** Style for the List item (the calendar) */
     calendarStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array]),
     /** Whether to use static header that will not scroll with the list (horizontal only) */
-    staticHeader: PropTypes.bool
+    staticHeader: PropTypes.bool,
+    /** Whether to scroll to selected dates on change */
+    scrollToDate: PropTypes.bool
   }
 
   static defaultProps = {
@@ -59,7 +61,8 @@ class CalendarList extends Component {
     showScrollIndicator: false,
     scrollEnabled: true,
     scrollsToTop: false,
-    removeClippedSubviews: Platform.OS === 'android' ? false : true
+    removeClippedSubviews: Platform.OS === 'android' ? false : true,
+    scrollToDate: true
   }
 
   constructor(props) {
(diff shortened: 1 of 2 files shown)


chenei:
Can you please explain more what it does?
Why we can’t we change the border radius today using theme prop?

samiede:
I'm not sure, at the time I wrote this small PR, it was not possible to change the radius via any customisation as it was hardcoded

src/calendar-list/index.js
```diff
@@ -59,22 +61,23 @@ class CalendarList extends Component {
     showScrollIndicator: false,
     scrollEnabled: true,
     scrollsToTop: false,
-    removeClippedSubviews: Platform.OS === 'android' ? false : true
+    removeClippedSubviews: Platform.OS === 'android' ? false : true,
+    scrollToDate: true
```
Inbal-Tish, line 65: What is 'scrollToDate' prop has to do with fixing borderRadius?

src/calendar/day/period/index.js
```diff
@@ -130,7 +130,7 @@ class Day extends Component {
 
     if (this.props.marking) {
       containerStyle.push({
-        borderRadius: 17
+        borderRadius: this.style.base.borderRadius
```
Inbal-Tish, line 133: This is the actual fix of this PR, right? Where is the 'style.base.borderRadius' is coming from? Why you're not using the 'theme' object and 'appStyle' to pass styles to the component?

emilisb:
Hi @samiede,

Any plans to finish with this?

src/calendar-list/index.js
samiede: I think this was a different issue where the calendar would scroll to the date automatically, even if you don't necessarily want it to.

src/calendar/day/period/index.js
samiede: If I remember correctly, this is what I wanted to do, but the border radius was hardcoded at the time, i.e. this part of the styling could not be overwritten by the appStyle object.
I apologise of this is obsolete by now, but I don't have access to the companies code anymore, so I'm not 100% sure in what situation I needed this.

samiede:
Hey @emilisb, I haven't worked on this since April, so I'm not quite sure anymore, but I have commented under the remarks with everything I remember!

PR #1124, opened 2020-04-24
Fix agenda scroll issue

narender2031:


azuxx:
Hello, thank you for the fix. Can you add also useNativeDriver fix in CalendarProvider please?

```
animateOpacity(disabled) {
    const {disabledOpacity} = this.props;
    if (disabledOpacity) {
      Animated.timing(this.state.opacity, {
        toValue: disabled ? disabledOpacity : 1,
        duration: 500,
        useNativeDriver: true //add this
      }).start();
    }
  }
```

narender2031:
@azuxx I didn't get this. Can you explain which file or function requires this?

azuxx:
> @azuxx I didn't get this. Can you explain which file or function requires this?

@narender2031 yes sure, it is located here: /src/expandableCalenda/calendarProvider.js, function name is animateOpacity.

useNativeDriver property is missing and adding it in this part of the code fixed the performance issue and warning message.

There are also other parts in the library where useNativeDriver is missing but maybe someone else will do the whole support to RN 0.62.x afterwards.

Thank you very much

narender2031:
@azuxx Attribute is added.

android/.project
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
```
Inbal-Tish, line 2: the android modified files should not be a part of this PR

android/.settings/org.eclipse.buildship.core.prefs
```diff
@@ -0,0 +1,13 @@
+arguments=
+auto.sync=false
```
Inbal-Tish, line 2: the android modified files should not be a part of this PR

src/calendar-list/index.js
```diff
@@ -130,8 +132,15 @@ class CalendarList extends Component {
           break;
         }
       }
+      if (scrollAmount < 0) {
+        this.setState({addMarginTop: Math.abs(scrollAmount)});
```
Inbal-Tish, line 136: You're invoking render while scrolling the list... Isn't there a better solution?

emilisb:
Hi @narender2031 ,

Any plans to finish this?

narender2031:
Sorry guys. I was a bit busy with other works. Now I'm back and will update the PR soon 
@emilisb @Inbal-Tish

francesco-clementi-92:
Any news?

PR #956, opened 2019-10-09
Remove deprecated lifecycle API

Moreno97:
This PR removes the old deprecated `componentWillReceiveProps` lifecycle API from the `calendar` component.
Now, it uses the last recent implementation with `getDerivedStateFromProps`.

PR #942, opened 2019-09-29
refactor life cycles deprecated

avrcoelho:
Refactor life cycles componentWillMount and componentWillReceiveProps

PR #770, opened 2019-03-16
Added missed calendarHeight props to Agenda

IronTony:
Added missed calendarHeight props to Agenda so we can use:
`
<Agenda
    calendarHeight={480}
/>
`
to customize the height of the Calendar

PR #719, opened 2019-01-16
Add initialNumToRender prop to CalendarList 

NedimMuhamedagic:
Not sure if this is actually needed, just passing the `initialNumToRender` prop to the `FlatList` component so we can control how many months are pre-rendered.

PR #1149, opened 2020-05-22
New lifecycle / initialPosition Agenda / calendarHeigh / initialNumToRender

magrinj:
Hi, here what this PR do:

- [x] Update class old lifecycle to new one (#956, #942)
       (Some PR already exist, but fix it just partially, or don't do it the right way)
       For this part, `ReservationList` Component was the tricky one, I use a global variable to avoid unwanted update, but `selectedDay` can be moved into state.

- [x] Add initialPosition prop for Agenda (open/closed) (#974, #990)

- [x] Fix missed calendarHeight prop for Agenda (#770)

- [x] Allow to specify initialNumToRender for ReservationList and CalendarList (#719)
       (This is useful if you want better performance, and it depend on your component height, so it must be accessible from outside)

Some PR already exist for some point, but suggested changes has not been made.
You can cherry pick the changes you want, or I can split this PR if needed.

I'm open to any suggestion/modification !


magrinj:
@Inbal-Tish Hey ! Any news on this ?

magrinj:
@Inbal-Tish I saw some PR merged on similar subject but partially fixing the old lifecycles.
Are my first fix still relevant ?
Maybe you can just cherry pick the other ones

Issue #1128, opened 2020-04-29
Multi mark and multi markingType


muslimmuda15:
Hi,
Can I make my calendar like this :
![calendar](https://user-images.githubusercontent.com/26782958/80560496-52fff180-8a0b-11ea-8cf0-fcb7d74475d3.png)

how do I make my calendar have different of `date marking style`?

chenei:
Hi,
Please check the last **NEW!** section here:
https://github.com/wix/react-native-calendars#date-marking

PR #1157, opened 2020-06-01
Adding the option to combine between Dot marking and Period marking.

chenei:


chenei:
Resolves #1128

ethanshar:
@chenei 
I see this PR is based on the code of this PR 
https://github.com/wix/react-native-calendars/pull/1158
right?

chenei:
> @chenei
> I see this PR is based on the code of this PR
> #1158
> right?

Yeah sorry, I see that there is some coupling.

Issue #1095, opened 2020-03-26
set all dates "marked" default and then change as per requirement


TapsaleDipa:
Please make our job easier by filling this template out to completion. If you're requesting a feature instead of reporting a bug, please feel free to skip the Environment and Reproducible Demo sections.

I need to set all dates default as marked(example blue) and then changes colour for few dates to yellow.

1-2 sentences describing the problem you're having or the feature you'd like to request

I need to set all dates default as marked(example blue) and then changes colour for few dates to yellow.

What action did you perform, and what did you expect to happen?

I have observed that, I need to create object of all the dates and assign dot colour and assign this to markedDate.....

chenei:
I am not sure if we need such a feature.
I think it's a special case and can be solved by using `markedDates`.

TapsaleDipa:
With markedDates we can add dot colour to particular date for that I need some date, in my case I need it for whole calendar and then I will update the colour as I am adding item to calendar.

ethanshar:
@chenei 
What about having a `defaultDayMark` prop that will be merged with specific `markDates` that are being passed for specific dates

chenei:
Yeah, we can do that.
I'll take it.

PR #1166, opened 2020-06-04
Adding defaultDayMark

chenei:


chenei:
I also refactored the example file so we can comment easily examples.

chenei:
resolves #1095

ethanshar:
Before I dive in into this one. 
I meant that the `defaultDayMark` prop will be an **object** specifying the default marking options for all days. 
But maybe I misunderstood the requirement :/

Inbal-Tish:
@chenei Can you please add a description.

emilisb:
@chenei any plans to finish with this?

PR #1167, opened 2020-06-05
Custom calendar header v2

chenei:
Improve custom calendar header options by splitting year and month into separate components.
Allow rendering single letter weekday.

How the example looks:
https://ibb.co/q9nbnRV


ethanshar:
@chenei 
I think this solution is still pretty limited. 
It will allow you to accomplish the solution in the reference but customization never ends. 

As a thumb rule I usually provide users the common UI with basic ways to customize it (by passing styles and various props) 
But eventually for advanced cases I provide a **render prop**. 
In this case I would add a new prop called `renderCustomHeader` that users can pass so they control completely what is being rendered inside the header. 
They can control alignment, add icons and so on..

For us it's good because we don't have too many props (which makes the API more difficult to understand) 
And also, we allow users do whatever they want.

Just notice that you do need to provide some information for that function (`renderCustomHeader`), in our case it's the year and month so the user have what to render.

chenei:
@ethanshar Cool, ill implement.
What about this: `allowWeekDaysFirstLetterOnly`.
How can we make it more flexible?

PR #1177, opened 2020-06-07
Custom Header with props (Updated)

jacm80:
The component was adjusted to allow a custom header, some css were opened, and day / custom was modified to render icon, added accessibility format fix for the blind people


jacm80:
calendar-custom

PR #1212, opened 2020-06-25
Support fews Flatlist callback

anhtuank7c:
Added `onScrollBeginDrag, onScrollEndDrag, onMomentumScrollBegin, onMomentumScrollEnd` to give dev have more toys to interactive with CalendarList


emilisb:
Could you add these props to prop types as well?

PR #1213, opened 2020-06-30
Fix CalendarList scroll bug

tomjohn1028:
#### The Issue
https://www.loom.com/share/156836f07c804eadb59bef856b4b3516
Note: Each time the mouse is moving I'm clicking and trying to scroll. 

The video above details an issue trying to scroll the _CalendarList_ component. I eventually found out it was related to [react-native-swipe-gestures](https://github.com/glepur/react-native-swipe-gestures) returning a null value when swiping and causing the _Calendar_ component to hijack the gesture handling (see [null issue](https://github.com/glepur/react-native-swipe-gestures/issues/40) )

#### Proposed Solution

Below is a proposed patch that adds a new _disableCalendarGestureRecognizer_ property to the _Calendar_component. When using a "infinite" vertically scrolled _CalendarList_ there is no need for the _Calendar_ items to handle or recognize any gestures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,8 @@ LocaleConfig.defaultLocale = 'fr';
   disableAllTouchEventsForDisabledDays={true}
   /** Replace default month and year title with custom one. the function receive a date as parameter. */
   renderHeader={(date) => {/*Return JSX*/}}
+  // (Advanced) Disable calendar swipe gesture handling. Default: false
+  disableCalendarGestureRecognizer={false}
 />
 ```
 
--- a/src/calendar-list/index.js
+++ b/src/calendar-list/index.js
@@ -69,9 +69,9 @@ class CalendarList extends Component {
 
   constructor(props) {
     super(props);
-    
+
     this.style = styleConstructor(props.theme);
-    
+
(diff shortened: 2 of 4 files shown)


santiagomartinez:
This solves #1224

chenei:
Hi @tomjohn1028 and @santiagomartinez ,

Sorry for the performance issue. you right, we shouldn't detect a left/right gesture on vertical scrolling.

Thanks for contributing @tomjohn1028. I just want to let you know that we already have a prop called `enableSwipeMonths` which disabled by default.
To take advantage of this prop, I submit this PR:
https://github.com/wix/react-native-calendars/pull/1248
can you please review? it achieves the same result using less code.

emilisb:
Hi @tomjohn1028 @chenei , 

I believe this issue was solved by #1248 , wasn't it? If so, can we close this PR?

tomjohn1028:
@chenei @emilisb My apologies I haven't been able to work on my app in quite some time and these comments must have been lost in the shuffle. 

https://github.com/wix/react-native-calendars/pull/1248 looks like a good fix. I'll close this out

PR #1215, opened 2020-06-30
Allow props override in agendaList

belgamo:
## Motivation
I wasn't able to customize the prop `renderSectionHeader`. 

## Solution
Put spread of props as the last prop of `<SectionList />`

### DIFF
--- a/src/expandableCalendar/agendaList.js
+++ b/src/expandableCalendar/agendaList.js
@@ -7,7 +7,6 @@ import XDate from 'xdate';
 import styleConstructor from './style';
 import asCalendarConsumer from './asCalendarConsumer';
 
-
 const commons = require('./commons');
 const UPDATE_SOURCES = commons.UPDATE_SOURCES;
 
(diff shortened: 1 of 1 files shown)


belgamo:
> Maybe a spec?
Maybe an units specs would be great to check if props overriding is working, but this project just do e2e tests.

src/expandableCalendar/agendaList.js
```diff
@@ -158,7 +177,10 @@ class AgendaList extends Component {
         onScroll={this.onScroll}
         onMomentumScrollBegin={this.onMomentumScrollBegin}
         onMomentumScrollEnd={this.onMomentumScrollEnd}
-        onScrollToIndexFailed={(info) => { console.warn('onScrollToIndexFailed info: ', info); }}
+        onScrollToIndexFailed={(info) => {
+          console.warn('onScrollToIndexFailed info: ', info);
+        }}
+        {...this.props}
```
Inbal-Tish, line 183: This is risky and should be located here. Please specify the need to override in the description.

src/expandableCalendar/agendaList.js
```diff
@@ -25,13 +24,17 @@ class AgendaList extends Component {
     /** day format in section title. Formatting values: http://arshaw.com/xdate/#Formatting */
     dayFormat: PropTypes.string,
     /** style passed to the section view */
-    sectionStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array])
-  }
+    sectionStyle: PropTypes.oneOfType([
```
Inbal-Tish, line 27: Please refrain from formatting the file using your own tools and preferences, we are trying to keep the code unified.

src/expandableCalendar/agendaList.js
Inbal-Tish: If you need 'renderSectionHeader' why no exposing a prop?

PR #1225, opened 2020-07-09
Implement a custom header prop

bradens:
Allows you to completely override the header.

Similar to #546 but it fixes the problems.

### DIFF
--- a/example/src/screens/calendars.js
+++ b/example/src/screens/calendars.js
@@ -371,7 +371,7 @@ const CalendarsScreen = () => {
               borderBottomColor: 'lightgrey'
             }
           ]}
-          customHeader={() => <Text>Custom Header</Text>}
+          customHeader={React.forwardRef((props, ref) => <Text ref={ref}>Custom Header</Text>)}
         />
       </Fragment>
     );
--- a/example/src/screens/calendars.js
+++ b/example/src/screens/calendars.js
@@ -357,6 +357,25 @@ const CalendarsScreen = () => {
     );
   };
 
+  const renderCalendarWithCustomHeader = () => {
+    return (
+      <Fragment>
+        <Text style={styles.text}>Calendar with custom header component</Text>
+        <Calendar
+          testID={testIDs.calendars.LAST}
+          style={[
+            styles.calendar,
+            {
+              height: 250,
+              borderBottomWidth: 1,
+              borderBottomColor: 'lightgrey'
+            }
+          ]}
+          customHeader={() => <Text>Custom Header</Text>}
+        />
+      </Fragment>
+    );
+  };
   return (
     <ScrollView showsVerticalScrollIndicator={false} testID={testIDs.calendars.CONTAINER}>
       {renderCalendarWithSelectableDate()}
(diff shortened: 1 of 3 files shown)


bradens:
@Inbal-Tish does this fix the issues with #546?

bradens:
@Inbal-Tish Does this PR need anything else before it can be merged.

example/src/screens/calendars.js
```diff
@@ -357,6 +357,25 @@ const CalendarsScreen = () => {
     );
   };
 
+  const renderCalendarWithCustomHeader = () => {
+    return (
+      <Fragment>
+        <Text style={styles.text}>Calendar with custom header component</Text>
+        <Calendar
+          testID={testIDs.calendars.LAST}
```
Inbal-Tish, line 365: 'testID' must be unique string

example/src/screens/calendars.js
```diff
@@ -357,6 +357,25 @@ const CalendarsScreen = () => {
...
+          style={[
+            styles.calendar,
+            {
+              height: 250,
+              borderBottomWidth: 1,
+              borderBottomColor: 'lightgrey'
+            }
+          ]}
+          customHeader={React.forwardRef((props, ref) => <Text ref={ref}>Custom Header</Text>)}
```
Inbal-Tish, line 374: Would love if you can style the header a bit

Inbal-Tish:
@bradens Hi. Thank you for your contribution. I love the idea of the custom header and I do think we need something of this sort. My main issue with this implementation is that once you pass 'customHeader' you do not use the Calendar's header, which means you have to implement the arrows functionality all by yourself. Otherwise, there is no way to change months in the calendar, unless it is a CalendarList component where you have the scroll.

bradens:
> @bradens Hi. Thank you for your contribution. I love the idea of the custom header and I do think we need something of this sort. My main issue with this implementation is that once you pass 'customHeader' you do not use the Calendar's header, which means you have to implement the arrows functionality all by yourself. Otherwise, there is no way to change months in the calendar, unless it is a CalendarList component where you have the scroll.

I think many applications (including my own) require full control over what the header is, not just the ability to insert nodes into the default header (which we cannot change the optimizations on, making updates to the header impossible without changing keys -- probably the source of many issues on here). 

I get your concern about the arrows, but you already provide ways to disable them (`hideArrows` prop), and ways to set the current date (`current` prop), so I don't think we are causing any problems by providing a totally custom header here.

example/src/screens/calendars.js
```diff
@@ -357,6 +357,40 @@ const CalendarsScreen = () => {
...
+          <TouchableOpacity onPress={() => console.log("Tapped!")}><Text>Tap Me</Text></TouchableOpacity>
+        </View>
+      );
+    });
+
+    return (
+      <Fragment>
+        <Text style={styles.text}>Calendar with custom header component</Text>
+        <Calendar
+          testID={`custom-header-calendar`}
```
Inbal-Tish, line 380: If you don't have tests for this calendar example you can remove 'testID' prop as there is no need for it, otherwise, add it to `testIDs.calendar' constants and add a test in calendars.spec.js

bradens: Done in 00e09ad

Inbal-Tish: Sorry if I wasn't clear, there is a test that scrolls the calendars and checks if the last one is visible. That means that the last calendar example should have `testID={testIDs.calendars.LAST}`. So, you need to move your example to be one before last or move the 'testID' from the previous last calendar example to yours. Please run the tests before you submit to make sure it all checks out. Thanks! And sorry for the inconvenience.

Inbal-Tish (approved):

Issue #52, opened 2017-06-06
Agenda expand month range.


Symous:
Hi.awesome lib!
Is there any way to enable us to define the range of the agenda when it expand ? just like only showing one year. Thanks. :-)

tautvilas:
Hi there are `pastScrollRange` and `futureScrollRange` paramas available to `CalendarList`. Would behaviour of these parameters work for you if they were also available for Agenda?

Symous:
@tautvilas , these 2 props pastScrollRange & futureScrollRange are not worked with Agenda.

tautvilas:
Hey, you're right I need to add these options to agenda

Symous:
@tautvilas PR #54

PR #1233, opened 2020-07-15
Update index.js

nanavinay:
Changes in renderPeriods adding periods.name in Text component

### DIFF
--- a/src/calendar/day/multi-period/index.js
+++ b/src/calendar/day/multi-period/index.js
@@ -70,7 +70,13 @@ class Day extends Component {
             marginRight: 4
           });
         }
-        return <View key={index} style={style}/>;
+        return (
+           <View key={index} style={style}>
+              validPeriods.map((period, index) => {
+              <Text style={{fontWeight:'600', fontSize:12}}>{period.name}</Text>
+              })
+           </View>
+        );
       });
     }
     return;


nanavinay:
in multi-period  should have prop called name should be nice to render an item in the calendar

src/calendar/day/multi-period/index.js
```diff
@@ -70,7 +70,13 @@ class Day extends Component {
             marginRight: 4
           });
         }
-        return <View key={index} style={style}/>;
+        return (
+           <View key={index} style={style}>
+              validPeriods.map((period, index) => {
```
Inbal-Tish, line 75: Why a map inside a map (line #52)?

src/calendar/day/multi-period/index.js
```diff
@@ -70,7 +70,13 @@ class Day extends Component {
...
+              <Text style={{fontWeight:'600', fontSize:12}}>{period.name}</Text>
```
Inbal-Tish, line 76: If you add a 'name' you should declare it, otherwise, no one will know that there is such an option.

Inbal-Tish:
@nanavinay Please provide screenshots and a detailed description of your feature.

src/calendar/day/multi-period/index.js
Inbal-Tish: Also, inline style is bad for performance and you should provide a way to override it.

PR #1235, opened 2020-07-17
Add additional e2e sanity tests & upgrade detox version

ejakavicius:


emilisb:
I wonder if it is possible to run public PR builds on Mac agents. It would be great to have Detox running on CI.

Issue #105, opened 2017-07-19
Different calendar sizes on Agenda


matwii:
My Agenda is working perfectly on my Android phone, but on my iPad the calendar view over the agenda is really big. Is there any way I can shrink this? I'm using flexbox so I don't understand how it gets bigger than my flexbox. Here are two images, one from the android phone where it works perfectly and one from the ipad where the calendar view of the weeks are to big. I want to show one week at the time on both phone and ipad.
![screenshot_20170719-162604](https://user-images.githubusercontent.com/8501876/28372322-afd90ec8-6c9f-11e7-8494-55a333139464.png)
![img_0001](https://user-images.githubusercontent.com/8501876/28372330-b304c59c-6c9f-11e7-9fb1-1d8705f21600.PNG)

tautvilas:
looks like it's related to #97 bug @overengineered

overengineered:
Here is a simple problem, will fix it tomorrow. #97 I don't understand, probably will remain when this is fixed.

matwii:
@overengineered You know how to fix this?

overengineered:
Yes, `overflow:'hidden'` is missing in style in one of the subviews in Agenda. Adding it to Agenda style could be a workaround.

matwii:
@overengineered Yeah adding overflow: 'hidden' in the style object for Agenda worked perfectly. I think that either you need to add it in the documentation or fix it :)

tautvilas:
@matwii your PR with fix also welcome :)

PR #1240, opened 2020-07-24
Render custom week days interface

HyeonjuPark:
I wanted to change my calendar's day of week style one by one. 

For example, if I wanted to give opacity to only Friday, I could not do that because there is only an option to change style which change whole day of week texts. 

So, I added renderWeekDays props to Calendar, CalendarList and CalendarItem.
And I also added this props into README with comment.


src/calendar/header/index.js
```diff
@@ -99,14 +100,18 @@ class CalendarHeader extends Component {
...
-    const {disabledDaysIndexes} = this.props;
+    const {disabledDaysIndexes, renderWeekDays} = this.props;
     return weekDaysNames.map((day, idx) => {
       const dayStyle = [this.style.dayHeader];
 
       if (_.includes(disabledDaysIndexes, idx)) {
         dayStyle.push(this.style.disabledDayHeader);
       }
 
+      if (renderWeekDays) {
```
Inbal-Tish, line 111: If you're rendering the weekDays Text component yourself, this should be at the top (before line #105). Since you're not using 'dayStyle' const there is no need to compute it.

Inbal-Tish:
@HyeonjuPark Hi. Please resolve conflicts

PR #1246, opened 2020-07-30
renderArrow correction

mk1020:
added the ability to add 2 or more arrows on one side (for example, to switch the year). The this.props.addMonth method was added to the renderArrow property parameters

### DIFF
--- a/src/calendar/header/index.js
+++ b/src/calendar/header/index.js
@@ -1,6 +1,5 @@
 import React, {Component, Fragment} from 'react';
-import {ActivityIndicator, Platform} from 'react-native';
-import {View, Text, TouchableOpacity, Image} from 'react-native';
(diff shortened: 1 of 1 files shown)


src/calendar/header/index.js
```diff
@@ -151,36 +150,40 @@ class CalendarHeader extends Component {
...
-        >
-          {this.props.renderArrow
-            ? this.props.renderArrow('left')
-            : <Image
-              source={require('../img/previous.png')}
-              style={this.props.disableArrowLeft ? this.style.disabledArrowImage : this.style.arrowImage}
-            />}
-        </TouchableOpacity>
+        this.props.renderArrow
+          ? this.props.renderArrow('left', this.props.addMonth)
```
Inbal-Tish, line 154: This is a breaking change. You're moving the 'renderArrow' out of the Touchable area.

PR #1266, opened 2020-08-14
Disable code signing for debug example app

emilisb:
This will no longer require to change bundle identifier and development team in order to run debug app.

### DIFF
--- a/ios/CalendarsExample.xcodeproj/project.pbxproj
+++ b/ios/CalendarsExample.xcodeproj/project.pbxproj
@@ -686,8 +686,7 @@
 						TestTargetID = 13B07F861A680F5B00A75B9A;
 					};
 					13B07F861A680F5B00A75B9A = {
-						DevelopmentTeam = 78G36M9BB8;
-						ProvisioningStyle = Automatic;
+						ProvisioningStyle = Manual;
 					};
 					2D02E47A1E0B4A5D006451C7 = {
 						CreatedOnToolsVersion = 8.2.1;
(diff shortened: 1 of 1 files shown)


Inbal-Tish:
@emilisb Hi. I don't really like this solution of changing the code signing to manual. It's ok that you'll need to choose team to run on a debug device, but I do agree we need to find a solution for the bundle id (in rnuilib we're not changing every time we want to run on a debug device).

emilisb:
@Inbal-Tish What regressions does this approach introduce? Using this method you don't have to select a development team at all. Code signing should not be required for open source library debug builds. I tested this change on a different machine and the project ran without any modifications in team.

Inbal-Tish:
It's not just manual for debug builds, it's debug and release. I don't really want to start managing Provisioning profiles and certificates.

Inbal-Tish:
oh sorry, I didn't see you change it just for debug...

Inbal-Tish:
It still requires a provisioning profile and now you must manage it manually....

emilisb:
I tested the project on a different machine where this project hasn't been run before and it worked for me. I can test it on my personal machine when I get home with a non-Wix developer account.

Inbal-Tish:
I couldn't run it on a device while on this branch.

emilisb:
It won't run on a device, but it will run on a simulator. For a real device you have to tick "Automatically manage signing". You cannot run an iOS app on a real device without provisioning profile, and no matter what you do, people will have to change the bundle identifier and the certificate in order to run it on a real device.

I don't think many people run demo apps on real devices in any case.

Inbal-Tish:
That is exactly my point. And that's why I don't like to change it to manual. Why is there a problem running on the simulator?

emilisb:
Xcode attemps to sign a debug app with Wix development team and bundle identifier. It happens to me too, even though I have a Wix development certificate, not to mention all the 3rd party contributors.

Issue #1191, opened 2020-06-10
Add Arrows to Agenda Header


okechukwu0127:
Please make our job easier by filling this template out to completion. If you're requesting a feature instead of reporting a bug, please feel free to skip the Environment and Reproducible Demo sections.

## **Description**

I am looking at adding the arrows (next/prev) on the Agenda header when it is closed. This would allow the user to go next and previous week/days without clicking on the knob
Realized that i can only have access to this using the Calender

**1-2 sentences describing the problem you're having or the feature you'd like to request**

Access to hideArrows={true} and  renderArrow={(direction) => (<Arrow/>)} using Agenda

## **Expected Behavior**

Show and Hide the Arrows on the Agenda option vis Props

**What action did you perform, and what did you expect to happen?**
hideArrows={true}
But did not show up the next.prev arrows on the Agenda header

## **Observed Behavior**
props hideArrows={true}
does not work for Agenda

**What actually happened when you performed the above actions?**
nothing

If there's an error message, please paste the *full terminal output and error message* in this code block:
No error
```
Error text goes here!
```

## **Environment**
Android
react-native-calendars@1.217.0

Please run these commands in the project folder and fill in their results:

* `npm ls react-native-calendars`:
* `npm ls react-native`:

Also specify:

1. **Phone/emulator/simulator & version:**
Infinix Zero5 Pro

## Reproducible Demo

![image](https://user-images.githubusercontent.com/23473673/84323616-09103c80-ab6f-11ea-8ea7-d7deeb8a0228.png)
Something like this image but with a right and left arrows

Please provide a minimized reproducible demonstration of the problem you're reporting.

Issues that come with minimal repro's are resolved much more quickly than issues where a maintainer has to reproduce themselves.

FatemeMirzaeii:
I need this feature, too.
isn't there any way to swipe left and right to other weeks right now?

belgamo:
Great! I'm working on it. Soon I will send the PR.

NooruddinLakhani:
Please make it soon, I'm waiting for it

okechukwu0127:
@belgamo I await the new release with access to previous and next week date data.
Regards

quangnguyen3499:
I'm waiting too

rabiaah:
Is it available?!

pgeimer:
Any news on this? Is there no other way to get the arrows to ``<Agenda/>``?

rabiaah:
Seems no one cares about this feature :)

PR #1270, opened 2020-08-16
Arrow component

belgamo:
## Motivation
I noticed that the arrow code is repeated in some places and the logic to render an arrow as well. I decided to create a component to abstract the logic and make easier the maintainability. Another reason is this [feature request](https://github.com/wix/react-native-calendars/issues/1191) that probably will need of an arrow component. I'm already working on it as well.

## Component props
| Prop  | Type | Description
| --- | --- | --- |
| type * | string  | The arrow direction (left or right)
| onPress * | func  | A callback triggered when on press arrow
| disableArrow | boolean  | An option to disable arrow on press
| renderArrow | func  | A method to render the arrow. This method should receive the type of arrow.
| style | object | A style object to customize arrow button
| disabledArrowImageStyle | object | A style to the disabled arrow image
| arrowImageStyle | object | A style to the arrow image
| testID | string | A string to be used in testing.

## Testing strategy
For this component, I think would be fine to write unit tests with [react-native-testing-library](https://github.com/callstack/react-native-testing-library). If you like, I can open another PR to set up jest and testing-library.

### DIFF
--- a/src/calendar/header/index.js
+++ b/src/calendar/header/index.js
@@ -1,13 +1,14 @@
 import React, {Component, Fragment} from 'react';
 import {ActivityIndicator, Platform} from 'react-native';
-import {View, Text, TouchableOpacity, Image} from 'react-native';
+import {View, Text} from 'react-native';
 import XDate from 'xdate';
 import PropTypes from 'prop-types';
 import styleConstructor from './style';
 import {weekDayNames} from '../../dateutils';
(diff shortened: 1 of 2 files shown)


abnersajr:
@belgamo there is one test failing. Can you take a look at that, if it's something in CI only, or does it exist?

belgamo:
> @belgamo there is one test failing. Can you take a look at that, if it's something in CI only, or does it exist?

Yeah I think so, it's only a CI problem.

rabiaah:
Anyone is fixing it? Or it been fixed already?!

PR #1273, opened 2020-08-20
removed old fallbacks and hacks for Calendar style PropTypes

zloyrusskiy:
There was a wrong PropType fallback for react-native-web, because of bad PropType **PropTypes.shape({style: PropTypes.object})**
And there is no sense in other fallbacks, because of react-native version 0.62.2 at deps (it has typescript type checks now, not PropTypes).

P.S. fixing warnings
_index.js:1 Warning: Failed prop type: Calendar: prop type `style` is invalid; it must be a function, usually from the `prop-types` package, but received `undefined`._
and
_Attempted import error: 'ViewPropTypes' is not exported from 'react-native-web/dist/index' (imported as 'ReactNative')._

### DIFF
--- a/src/agenda/index.js
+++ b/src/agenda/index.js
@@ -11,6 +11,7 @@ import styleConstructor from './style';
 import {VelocityTracker} from '../input';
 import {AGENDA_CALENDAR_KNOB} from '../testIDs';
 
+
 const HEADER_HEIGHT = 104;
 const KNOB_HEIGHT = 24;
 
@@ -68,8 +69,8 @@ export default class AgendaView extends Component {
     /** Collection of dates that have to be marked. Default = items */
     markedDates: PropTypes.object,
     /** Optional marking type if custom markedDates are provided */
-    markingType: PropTypes.string /*
-    /** Hide knob button. Default = false */,
+    markingType: PropTypes.string,/*
+    /** Hide knob button. Default = false */
     hideKnob: PropTypes.bool,
     /** Month format in calendar title. Formatting values: http://arshaw.com/xdate/#Formatting */
     monthFormat: PropTypes.string,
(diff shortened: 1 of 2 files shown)


Inbal-Tish:
@zloyrusskiy Hi. Please remove all formatting so I can review only the relevant code. Thanks

zloyrusskiy:
formatting was needed because of your
image

src/agenda/index.js
```diff
@@ -35,7 +29,7 @@ export default class AgendaView extends Component {
     /** Specify theme properties to override specific styles for calendar parts. Default = {} */
     theme: PropTypes.object,
     /** agenda container style */
-    style: viewPropTypes.style,
+    style: PropTypes.object,
```
Inbal-Tish, line 32: 'style' prop type should be PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array]) as you can pass style inline as an object or array and you can pass it from a StyleSheet where it's just a number.

src/agenda/index.js
```diff
@@ -14,12 +14,6 @@ import {AGENDA_CALENDAR_KNOB} from '../testIDs';
 
 const HEADER_HEIGHT = 104;
 const KNOB_HEIGHT = 24;
-//Fallback for react-native-web or when RN version is < 0.44
```
Inbal-Tish, line 17: This is not a breaking change? How can I test it?

honia19:
@zloyrusskiy should you fix comments?

src/agenda/index.js
zloyrusskiy: ok, i have returned `View.propTypes.style` there

src/agenda/index.js
zloyrusskiy: this calendar version has dependency on "react-native": "0.62.2" (it is located in devDependencies, it's wrong)
from this version till master branch there is no changes with `View.propTypes.style`

Inbal-Tish:
@zloyrusskiy Please see merged PR #1424

Issue #569, opened 2018-07-23
Display dots in custom markingType


slecorvaisier:
## Description

in order to add a border to some dates, I switched `markingType` to `custom`. However, I still need the dots to be displayed when the day has `marked: true`. 

I tried to bring them back by modifying the day `customStyle` or the calendar theme via `stylesheet.day.single`, but nothing seems to do the job.

Is there any way to display the dots while working with a custom style ?

<img width="108" alt="screen shot 2018-07-23 at 15 58 03" src="https://user-images.githubusercontent.com/16785211/43081505-516233f4-8e92-11e8-8150-b7d88e23ae8e.png">

```jsx
'2018-07-16': {
  customStyles: {
    container: {
      borderWidth: 2,
      borderColor: 'black',
      backgroundColor: 'green',
      justifyContent: 'center',
    },
    text: {
      marginTop: 0,
    },
  },
  selected: true,
  marked: true,
 },
 '2018-07-17': { selected: true, selectedColor: 'green', marked: true },
```

vanessaleo:
Hi, have you got any news about it?

ODINA101:
same :(

slecorvaisier:
Hi all, it's been a while but here is what I found:
Unfortunatelly the dots cannot be displayed in the `custom` mode. There is simply no code to handle it in its file:
https://github.com/wix/react-native-calendars/blob/master/src/calendar/day/custom/index.js

For anyone willing to do his own implementation, a good start can be found in the `basic` mode:
https://github.com/wix/react-native-calendars/blob/55df6f1bde62cb25327648c25d6cd5d0e228684c/src/calendar/day/basic/index.js#L55

quanru:
+1

Maddumage:
Same issue here...

geetesh-gupta:
Same issue here... Please add this ability asap.

monspo1:
Hi guys, 

I am currently working on displaying dots in the custom mode (unoptimized yet, but shows dots at least). I plan to push to my forked repo within weeks.

![Screen Shot 2019-12-28 at 12 17 38 AM](https://user-images.githubusercontent.com/10152090/71539248-83be0600-2907-11ea-9dc3-96a44037ea15.png)

[...]

PR #1289, opened 2020-09-16
Add basic dots implementation when using custom markingType

awilliamsra:
Ensure basic dots can be displayed even when using custom markingType - PR based upon https://github.com/wix/react-native-calendars/issues/569


Inbal-Tish:
@awilliamsra Hi. Thank you for the PR. I think that if you're using a custom day component you should implement the marking as well unless you can suggest a way for the user to choose the marking type (dot, multi-dot, period, multi-period) and not just dot. That said, I do plan a major refactoring that might allow adding that kind of flexibility when Marking will become a component used by the Day component, or something of that sort...

PR #1290, opened 2020-09-16
Add custom footer to calendar

awilliamsra:
Enable a custom footer to be added to the bottom of the calendar.

### DIFF
--- a/src/calendar/index.js
+++ b/src/calendar/index.js
@@ -99,6 +99,8 @@ class Calendar extends Component {
     renderHeader: PropTypes.any,
     /** Allow rendering of a totally custom header */
     customHeader: PropTypes.any,
+    /** Allow rendering of a totally custom footer */
+    customFooter: PropTypes.any,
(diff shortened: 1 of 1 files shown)


src/calendar/index.js
```diff
@@ -400,6 +403,10 @@ class Calendar extends Component {
             : <CalendarHeader {...headerProps}/>
           }
           <View style={this.style.monthView}>{weeks}</View>
+          { CustomFooter
+            ? <CustomFooter {...headerProps}/>
+            : null
```
Inbal-Tish, line 408: Why passing the 'headerProps' to the footer? Also, you don't need a trinary here, a simple {CustomFooter && < CustomFooter/>} should do the trick.

Inbal-Tish: Why not use a custom render function? I mean, a 'renderFooter' prop that will be called here and return the element, instead of passing the element?

PR #1298, opened 2020-09-30
Add theme object to weekcalendar

sfal:
Add support for the theme prop to WeekCalendar week view to allow for customization.

### DIFF
--- a/src/expandableCalendar/weekCalendar.js
+++ b/src/expandableCalendar/weekCalendar.js
@@ -30,7 +30,8 @@ class WeekCalendar extends Component {
     /** whether to have shadow/elevation for the calendar */
     allowShadow: PropTypes.bool,
     /** whether to hide the names of the week days */
-    hideDayNames: PropTypes.bool
+    hideDayNames: PropTypes.bool,
+    theme: PropTypes.object,
   };
 
   static defaultProps = {
@@ -158,7 +159,8 @@ class WeekCalendar extends Component {
   }
 
   renderItem = ({item}) => {
-    const {calendarWidth, style, onDayPress, ...others} = this.props;
+    const {calendarWidth, style, onDayPress, theme, ...others} = this.props;
+    const themeObject = Object.assign(theme);
 
     return (
       <Week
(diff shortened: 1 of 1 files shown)


src/expandableCalendar/weekCalendar.js
```diff
@@ -30,7 +30,8 @@ class WeekCalendar extends Component {
     /** whether to have shadow/elevation for the calendar */
     allowShadow: PropTypes.bool,
     /** whether to hide the names of the week days */
-    hideDayNames: PropTypes.bool
+    hideDayNames: PropTypes.bool,
+    theme: PropTypes.object,
```
Inbal-Tish, line 34: No need to declare 'theme' prop, it is declared in Calendar prop types. CalendarList gets all Calendar props and WeekCalendar gets all CalendarList props.

src/expandableCalendar/weekCalendar.js
```diff
@@ -158,7 +159,8 @@ class WeekCalendar extends Component {
   }
 
   renderItem = ({item}) => {
-    const {calendarWidth, style, onDayPress, ...others} = this.props;
+    const {calendarWidth, style, onDayPress, theme, ...others} = this.props;
+    const themeObject = Object.assign(theme);
```
Inbal-Tish, line 163: Why is 'Object.assign' necessary here?

src/expandableCalendar/weekCalendar.js
```diff
@@ -168,6 +170,7 @@ class WeekCalendar extends Component {
         style={[{width: calendarWidth || this.containerWidth}, style]}
         markedDates={this.getMarkedDates()}
         onDayPress={onDayPress || this.onDayPress}
+        theme={themeObject}
```
Inbal-Tish, line 173: Isn't 'theme' prop already passed in {...others}?

Inbal-Tish:
@sfal Hi. The 'theme' prop is already passed to WeekCalendar. I don't understand the need for this change, can you please elaborate?

PR #1299, opened 2020-10-04
Unblock rerendering when no items for agenda

AdamWier:
I use this library in a project of mine and I have buttons that trigger overlays that render in the day view. These buttons are present even when there are no items in the agenda for that day. Because of a function that attempts to optimize the rerenders, these overlays could not be displayed. In this pull request, there are two commits. The first removes the condition, but then I thought it might be good to make this toggleable and the second commit does that. If this change is taken, we would need to update the @types library with it.


Inbal-Tish:
@AdamWier I'm not sure I understand the need. Can you add some screenshots to better explain the scenario?

PR #1300, opened 2020-10-04
Fixed startingDate and endingDate style

AdamWier:
This would fix a common problem with starting and ending date in period selection not being rounded.

### DIFF
--- a/src/calendar/day/period/index.js
+++ b/src/calendar/day/period/index.js
@@ -127,7 +127,8 @@ class Day extends Component {
 
     if (this.props.marking) {
       containerStyle.push({
-        borderRadius: 17
+        borderRadius: 17,
+        overflow: 'hidden'
       });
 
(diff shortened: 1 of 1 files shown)


Ollynov (approved):

Inbal-Tish:
@AdamWier Hi. Can you please elaborate on the bug and add screenshots? Why is this change only relevant to Day component type **period** and not the others?

darron1217:
@Inbal-Tish I'm experiencing same bug here.

Here's Screenshot
Before
image

After
image

This PR modifies only `src/calendar/day/period/index.js` file.
So I guess other types would not be affected by this change.

In most cases, `borderRadius` should be applied with `overflow: hidden`.
I think this PR is fixing the problem in right way :)

Inbal-Tish (approved):

PR #1322, opened 2020-11-06
Current time indicator and scrolling to current time

planeer:
Adds current time indicator on current date.
image

Scrolls to current time when calendar is opened.
demo

Also `npm run test` doesn't work because of `Agenda should move to previous month` so I couldn't get through that.

### DIFF
--- a/src/timeline/Timeline.js
+++ b/src/timeline/Timeline.js
@@ -32,7 +32,8 @@ export default class Timeline extends React.PureComponent {
       })
     ).isRequired,
     scrollToNow: PropTypes.bool,
-    currentDateString: PropTypes.string
+    currentDateString: PropTypes.string,
+    updateCurrentTimeIndicatorEveryMinute: PropTypes.bool,
   };
 
   static defaultProps = {
@@ -74,10 +75,30 @@ export default class Timeline extends React.PureComponent {
   }
 
   componentDidMount() {
+    if (this.isCurrentDateStringForTimeIndicatorSet()) {
+      this.setState({
+        currentTimeIndicatorTopCoordinate: this.currentTimeOffset()
+      });
+
+      if (this.props.updateCurrentTimeIndicatorEveryMinute) {
+        this.interval = setInterval(() => {
+          this.setState({
+            currentTimeIndicatorTopCoordinate: this.currentTimeOffset()
+          });
+        }, 60000);
+      }
+    }
+
     if (this.props.scrollToFirst) {
       this.scrollToFirst();
     } else if (this.props.scrollToNow) {
       this.scrollToNow();
+    }
+  }
+
+  componentWillUnmount() {
+    if (this.props.updateCurrentTimeIndicatorEveryMinute) {
+      clearInterval(this.interval);
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/timeline/Timeline.js
```diff
@@ -72,7 +72,23 @@ export default class Timeline extends React.PureComponent {
   }
 
   componentDidMount() {
-    this.props.scrollToFirst && this.scrollToFirst();
+    if (this.props.scrollToFirst) {
+      this.scrollToFirst();
+    } else if (this.props.scrollToNow) {
```
Inbal-Tish, line 77: Where is 'scrollToNow' prop is declared?

Inbal-Tish:
@planeer Can you please add an example for this use-case?

emilisb:
Good job, I think it looks really cool! You should also add these things:
1. Add the new props to PropTypes
2. Add the new props to Readme
3. Add an example to demo app

src/timeline/Timeline.js
```diff
@@ -87,6 +103,38 @@ export default class Timeline extends React.PureComponent {
...
+    return offset * (timeNowHour - start) + (offset * timeNowMin) / 60;
+  }
+
+  _renderCurrentTimeIndicator() {
+    // currentDateForTimeIndicator format YYYY-MM-DD, e.g. 2020-11-06
+    if (typeof this.props.currentDateForTimeIndicator === 'undefined') {
+      return;
+    }
+
+    if (this.props.currentDateForTimeIndicator === moment().format('YYYY-MM-DD')) {
```
emilisb, line 121: `currentDateForTimeIndicator` is a long variable name, you may consider renaming this

planeer:
Two questions:
* Where in the Readme should I put this, because there is nothing about Timeline calendar there yet?
* I hope this was what you meant by adding it to the demo app. Actual demo would be a bit harder to produce now, because of hardcoded events from 2017. Or should I make an actual example somehow? :)

src/timeline/Timeline.js
```diff
@@ -87,6 +105,38 @@ export default class Timeline extends React.PureComponent {
...
+      return;
+    }
+
+    if (this.props.currentDateString === moment().format('YYYY-MM-DD')) {
+      // Time indicator should be displayed only on the current date
+      return (
+        <View
+          key={'timeNow'}
+          style={[
+            this.styles.lineNow,
```
amazingmarvin, line 129: Should be `this.style.lineNow` with the current version.

amazingmarvin:
@planeer A squash and a rebase would help with reviewing this! It's currently broken since `this.styles` has been renamed to `this.style`, but otherwise this seems to be good 👍.

src/timeline/Timeline.js
Inbal-Tish, line 129:
```suggestion
            this.style.lineNow,
```

Inbal-Tish:
@planeer Hi. Looks really great. All test pass on master branch so I suggest you'll check your changes to see what fails the test

planeer:
Okay I did a squash and a rebase, so everything is up to date and all tests pass now. Tomorrow I will also try to tackle re-rendering once per minute :)

src/timeline/Timeline.js
```diff
@@ -30,7 +30,10 @@ export default class Timeline extends React.PureComponent {
         summary: PropTypes.string.isRequired,
         color: PropTypes.string
       })
-    ).isRequired
+    ).isRequired,
+    scrollToNow: PropTypes.bool,
+    currentDateString: PropTypes.string,
+    updateCurrentTimeIndicatorEveryMinute: PropTypes.bool,
```
Inbal-Tish, line 36: I think it's better to let the user decide how often to update by passing the number of seconds/minutes, like 'updateInterval' or something like that (where default is 0 - no updates).

src/timeline/Timeline.js
```diff
@@ -30,7 +30,10 @@ export default class Timeline extends React.PureComponent {
         summary: PropTypes.string.isRequired,
         color: PropTypes.string
       })
-    ).isRequired
+    ).isRequired,
+    scrollToNow: PropTypes.bool,
+    currentDateString: PropTypes.string,
```
Inbal-Tish, line 35: Let the user pass a Date instead of a string with a specific format so you can validate and format it yourself internally.

src/timeline/Timeline.js
```diff
@@ -87,14 +126,47 @@ export default class Timeline extends React.PureComponent {
     }, 1);
   }
 
+  currentTimeOffset() {
```
Inbal-Tish, line 129: A function should be named as action, like 'getCurrentTimeOffset()'

fwidtmann:
holy shit i need this feature! how can i get it?

droplessjake:
I just raised an issue about exactly this before I saw it https://github.com/wix/react-native-calendars/issues/1620
I need this! Can we expect an update?

goxr3plus:
Any news on this ? It looks amazing :)

Inbal-Tish:
@goxr3plus These feature already exit on Timeline component. Check our `showNowIndicator` and `scrollToNow` props.

PR #1327, opened 2020-11-11
Incorporate react-native-calendars into nb-mobile project (CU-fpyv71)

derick-bai:
### Description
Added `.watchmanconfig` file so that when using [wml](https://github.com/wix/wml) unnecessary files/folders won't be copied over.
Also modified a few default styles that cannot be overwritten through the props provided by the `<CalendarList>` component.

### Motivation and Context
To implement per our UI design, we need a calendar component more powerful and customizable than the one comes with React Native/Expo.

### Types of Change
- Source code
-  Configuration

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,16 @@
+### Description
+<!--- Describe your changes in detail -->
+
+### Motivation and Context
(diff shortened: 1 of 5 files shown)


derick-bai:
Commit was pushed to the wrong repository.

PR #1334, opened 2020-11-27
added callback on toggling expandable calendar

maeschli:


Inbal-Tish:
@maeschli Hi. Can you please remove the formatting so I can review the relevant code?

maeschli:
@Inbal-Tish I'm not able to do that. Formatting is automatically applies while I'm commiting the file

src/expandableCalendar/index.js
```diff
@@ -301,9 +295,13 @@ class ExpandableCalendar extends Component {
         toValue: this._height,
         speed: SPEED,
         bounciness: BOUNCINESS,
-        useNativeDriver: false
+        useNativeDriver: false,
       }).start(this.onAnimatedFinished);
 
+      if (this.props.onCalendarToggled) {
```
Inbal-Tish, line 301: Why not use _.invoke instead?

Inbal-Tish: Also, why not invoke it on 'onAnimatedFinished' instead?

src/expandableCalendar/index.js
```diff
@@ -345,20 +343,26 @@ class ExpandableCalendar extends Component {
...
-  }
+  };
 
-  onDayPress = (value) => { // {year: 2019, month: 4, day: 22, timestamp: 1555977600000, dateString: "2019-04-23"}
+  onDayPress = (value) => {
+    // {year: 2019, month: 4, day: 22, timestamp: 1555977600000, dateString: "2019-04-23"}
     _.invoke(this.props.context, 'setDate', value.dateString, UPDATE_SOURCES.DAY_PRESS);
 
-    setTimeout(() => { // to allows setDate to be completed
+    if (this.props.onCalendarToggled) {
```
Inbal-Tish, line 355: I'm not sure that's the right place for it

src/expandableCalendar/index.js
maeschli: that's a simple callback, so I prefer to use it natively instead of lodash, that's easier to debug
could be changed also to `this.props.onCalendarToggled?.(isOpen)` wdyt?

I need to pass there position is it opened or closed and it counts directly at this place. So that's why it's called there.

src/expandableCalendar/index.js
maeschli: duplicated logic, will be removed

src/expandableCalendar/index.js
Inbal-Tish: Yes. Better.

Inbal-Tish (approved):

PR #1336, opened 2020-11-29
fixed #1304

### LINKED ISSUE #1304: Can't scroll weekCalendar to next/prev month
I cant scroll pre/next month in this week-calendar of react-native-calendar , It seem to back pre week instead of go to next month
```
<WeekCalendar
                            hideDayNames={true}
                            firstDay={1}
                            dayComponent={this.renderDate}
                            theme={calendarThemes}
                            allowShadow={false}
                            markedDates={{
                                ...this.state.markedDates,
                                [this.state.selectedDate.dateString]: {
                                    selected: true,
                                },
                            }}
                            pastScrollRange={(new Date().getFullYear() - 1900) * 12}
                            futureScrollRange={(2099 - new Date().getFullYear()) * 12}
                            current={this.state.selectedDate}
                            onDayPress={(date) => this.changeDate(date.dateString)} />
```
https://i.stack.imgur.com/3a27g.gif

tamago3keran:
This PR fixes issue on #1304. 

When we scroll `WeekCalendar` to the end, it calls `onMomentumScrollEnd`.
And `getDatesArray` is also called in the `onMomentumScrollEnd`.

Maybe it have to change newDateArray on scroll to the end.
But it loads same dateArray.
So I changed code to resolve issue.

|Before|After|
|:--:|:--:|
|||

Nov-29-2020 19-32-06

### DIFF
--- a/src/expandableCalendar/weekCalendar.js
+++ b/src/expandableCalendar/weekCalendar.js
@@ -45,6 +45,7 @@ class WeekCalendar extends Component {
 
     this.list = React.createRef();
     this.page = NUMBER_OF_PAGES;
+    this.gettingDatesMedianIndex = 0;
 
     this.state = {
       items: this.getDatesArray()
@@ -67,7 +68,7 @@ class WeekCalendar extends Component {
   getDatesArray() {
     const array = [];
     for (let index = -NUMBER_OF_PAGES; index <= NUMBER_OF_PAGES; index++) {
-      const d = this.getDate(index);
+      const d = this.getDate(index + this.gettingDatesMedianIndex);
       array.push(d);
     }
     return array;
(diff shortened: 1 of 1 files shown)


emilisb (changes requested):
Hi, thanks for your contribution!

This issue happens because you don't wrap `WeekCalendar` with `CalendarProvider.` `WeekCalendar` is not supposed to be used as a standalone component. We will add docs for this component and we will mention this. Please take a look at an example here: https://github.com/wix/react-native-calendars/blob/0f7037bdcaaa2ecce3b2d992c55c2b0b8d33fb51/example/src/screens/expandableCalendar.js#L175-L189

Unfortunately, we cannot take your PR because it causes issues when `WeekCalendar` is wrapped with `CalendarProvider`. We could take a PR that would allow to use this component standalone, but it should not break behaviour when used with provider.

You can check the correct usage in our demo app `Week calendar` section.

DonOzOn:
@emilisb i try wrap WeekCalendar with CalendarProvider but it not solve the problem .
i still try solution of @tamago3keran, there are bug with scroll fast and select date but it already generate a new week when scroll to end.Thank all for 
your support

emilisb:
@DonOzOn please look at the example I sent you, it works in the example app.

emilisb:
@tamago3keran feel free to open a new Pull Request if you decide to implement standalone support for `CalendarHeader`. I'm closing this PR due to the reasons I mentioned above. Thank you though!

PR #1351, opened 2020-12-02
Set up component tests + Add Calendar tests

emilisb:
This PR introduces infra for component testing and 38 tests for `Calendar` component.


src/calendar/day/basic/index.js
```diff
@@ -95,14 +85,16 @@ class Day extends Component {
       <TouchableOpacity
         testID={this.props.testID}
         style={containerStyle}
-        onPress={this.onDayPress}
+        onPress={!shouldDisableTouchEvent ? this.onDayPress : undefined}
```
emilisb, line 88: Usually I don't like changing production code to satisfy tests, but it looked like the easiest option in this case.

Inbal-Tish: I don't mind that. Then the same should go to 'onLongPress'

Inbal-Tish:
Amazing work! This is a huge step forward for this repo!!!

src/calendar/day/basic/index.js
emilisb: Done

Inbal-Tish (approved):

PR #1352, opened 2020-12-03
Infra/CalendarHeader props

Inbal-Tish:
sorting prop duplications

### DIFF
--- a/example/src/screens/calendars.js
+++ b/example/src/screens/calendars.js
@@ -107,7 +107,8 @@ const CalendarsScreen = () => {
             // textDisabledColor: 'red',
             'stylesheet.calendar.header': {
               week: {
-                marginTop: 5,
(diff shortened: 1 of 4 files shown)


example/src/screens/calendars.js
```diff
@@ -107,7 +107,8 @@ const CalendarsScreen = () => {
             // textDisabledColor: 'red',
             'stylesheet.calendar.header': {
               week: {
-                marginTop: 5,
+                marginTop: 30,
+                marginHorizontal: 12,
```
emilisb, line 111: What is this for?

emilisb (approved):

example/src/screens/calendars.js
Inbal-Tish: Just a fix for the example, it looked really off

PR #1366, opened 2020-12-08
Fix Agenda's "renderEmptyData" not rendering and missing theme propType

dtsolis:
This PR fixes the following things

1. The Agenda's **renderEmptyData**, when provided, is called but not renders anything
2. The Agenda's reservation component, is missing the **theme** propType and some theme overrides are not working


emilisb (approved):
Looks good to me, thank you!

Issue #94, opened 2017-07-12
bug when scrolling fast


sibelius:
When I scroll very fast up or down, and then stop.

it pass a wrong date on callback `loadItemsForMonth`

I'll try to come up with more information soon

jkomyno:
I guess we're talking about the same thing. This is the visual result I have after scrolling fast up or down.
![](http://i.imgur.com/dJnAeWf.png)

tautvilas:
@jkomyno does this happen also if agenda is full-screen? not in a modal?

tautvilas:
@sibelius do you have more info about this?

sibelius:
my agenda component
```
<Agenda
          theme={calendarTheme}
          items={items}
          loadItemsForMonth={this.loadItemsForMonth}
          renderItem={this.renderItem}
          renderEmptyDate={this.renderEmptyDate}
          rowHasChanged={this.rowHasChanged}
          onDayPress={this.onDayPress}
          onDayChange={this.onDayChange}
        />
```

```
loadItemsForMonth = (date: XDate) => {
    const month = new Date(date.year, date.month - 1, 1); // first day of the month
    const start = startOfMonth(month);
    const end = addMonths(start, 2); // fetch 2 months

    this.fetchItems(start, end);
  };
```

I've noticed that somehow the loadItemsForMonth get lost if I scroll to fast.

Sometimes date param o loadItemsForMonth is not the month that is been shown in the screen

flaviotobi:
I have the problem of the data coming from loadItemsForMonth is usually a month behind.
I'm not using the Agenda component in a modal, is "Full Screen" with marginTop:64 (this 64 is the space for the navbar).
Sometimes the loadItemsForMonth doesn't get called, another times it get's two times in a row.
This happen both in fast scroll and slow scroll.

react-native-calendars: v1.5.6

flaviotobi:
Guys a kick fix that worked for me it solves (Wrong Data Coming From loadItemsForMonth ).
Just add in the Agenda Component:
      
```
DayChange={(day)=>{activityData=day.dateString;}}
onDayPress={(day)=>{activityData=day.dateString;}}
```

And create a "Global Variavel"

```
var activityData = moment().format('YYYY-MM-DD');
export default class CalendarView extends Component {
```
And after this instead of use the variavel that comes from the component "Day"
 ```
 loadItemsForMonth={this.loadItems.bind(this)}
loadItems(day) {}
```
[...]

Issue #23, opened 2017-05-22
How to call scrollToDay() method to set the today's day in Agenda View?


dongawli:
Hi, 

I am very new to React native and working on Calendar App. I really like this calendar Agenda view.
I want to set the today's date on button click in Agenda view or move to a different month - 
 https://hub.mangoapps.com/sf/ODcxNzZfMTMyNzE0NQ .

Is it possible to set the date?

tautvilas:
Try this code:

```javascript
<Agenda ref={(agenda) => { this.agenda = agenda; }} />
```

Then in your 'today' handler method:

```javascript
this.agenda.chooseDay(date)
```

Tell me if this works to you. I probably will add `chooseDay` to documentation soon

dongawli:
Hi Tautvilas,

Thank you for prompt response. 

I've tried above solution but getting error "undefined is not an object (evaluating 'this.agenda.chooseDay')" - https://dongawli.tinytake.com/sf/MTYyNzk4NV81NTA4Mzgz

Do I need to import any component/class/file from React Native Calendars package ? 
Currently I've just imported {Agenda} from 'react-native-calendars'.

tautvilas:
Are you sure you got the reference code right?

https://facebook.github.io/react/docs/refs-and-the-dom.html

Also references are only accesible after component is mounted

dongawli:
Yes, I am using the reference code right. 
Even I am not able to use the 'this.state.items', timeToString() method in newly added handler methods. 
Same methods or state variables I am able to access from loadItems() method. There is something not right in my project may be. I am finding out. 

Please let me know is it necessary to register the component 
Navigation.registerComponent('Agenda', () => AgendaScreen); as I am not using registerScreens() in my project and directly accessing agenda.js

tautvilas:
This is example how you should achieve what you want. There was also one bug that I fixed just now to make `chooseDay` properly work.

```javascript
import React, {Component} from 'react';
import {
  Text,
  TouchableOpacity,
  View,
  Dimensions,
  Animated
} from 'react-native';
import XDate from 'xdate';

import {parseDate, xdateToData} from '../interface';
import dateutils from '../dateutils';
import CalendarList from '../calendar-list';
import ReservationsList from './reservation-list';
import styleConstructor from './style';

const CALENDAR_OFFSET = 38;

export default class AgendaView extends Component {

  constructor(props) {
    super(props);
    this.styles = styleConstructor(props.theme);
[...]

PR #1369, opened 2020-12-10
Number localization

jesuscc1993:
Replacement for [#1079](https://github.com/wix/react-native-calendars/pull/1079), following the suggested alternative.

The way this works is by passing a `numbers` prop to the calendar components (e.g.: `["٠", "١", "٢", "٣", "٤", "٥", "٦", "٧", "٨", "٩"]`). It was discussed adding it to the LocaleConfig instead but this is defined by the XDate dependency so we can't change it.


jguix (approved):
Reviewed code and passed tests. I think this pull request is useful, well organized and follow conventions.

jguix (approved):
Reviewed code and passed tests. I think this pull request is useful, well organized and follow conventions.

Inbal-Tish:
@jesuscc1993 Hi. I don't understand why we can't use LocaleConfig for that instead of holding a prop for the numbers array?

jesuscc1993:
@Inbal-Tish LocaleConfig is a type declared by XDate, so we can't just add to it.
It would require extending it and replacing the type definition for all places that use LocaleConfig .

Inbal-Tish:
@jesuscc1993 Look at 'app.js' where we define the 'LocalConfig' object in `LocaleConfig.locales['en']` line #23. Here we can add keys to the locale configuration on default 'en' and use it in the proper place (in this case, the Day component's Text). When the user wants to localize his app he will pass the other locales objects, as shown in the 'fr' example. In that object, he will pass the localized numbers array. See how I do it with the 'today' key.

jesuscc1993:
@Inbal-Tish this is what I mean:
image

Inbal-Tish:
@jesuscc1993 Do you get this error even when you defined the default 'en' local 'numbers' key in app.js? like so: 
`LocaleConfig.locales['en'] = {
  formatAccessibilityLabel: "dddd d 'of' MMMM 'of' yyyy",
  monthNames: [
    'January',
    'February',
    'March',
    'April',
    'May',
    'June',
    'July',
    'August',
    'September',
    'October',
    'November',
    'December'
  ],
  monthNamesShort: ['jan', 'feb', 'mar', 'apr', 'may', 'jun', 'jul', 'aug', 'sep', 'oct', 'nov', 'dec'],
  dayNames: ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'],
  dayNamesShort: ['S', 'M', 'T', 'W', 'T', 'F', 'S'],
  **numbers: ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9']**
};
`

Inbal-Tish:
@jesuscc1993 By the way, I will do the formatting in the Day/index.js line #94 instead of having duplications in all Day components.

jesuscc1993:
I see that the type definitions actually come from [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-native-calendars) and not this library, so they shouldn't be relevant and your request should be doable.

I don't know when I'll be able to work on this PR, though. It's unlikely that it'll be soon.

I don't understand what you mean by this:
> I will do the formatting in the Day/index.js line #94 instead of having duplications in all Day components.

Inbal-Tish:
@jesuscc1993 I meant that if you format the text in the Day/index component instead on the Day/basic/index component, we'll have in the Day/period/index as well

jesuscc1993:
The PR has been reworked to meet your requests. Please, review at your earliest convenience.
@Inbal-Tish

Inbal-Tish (approved):

Inbal-Tish:
@jesuscc1993 Greate job! Thank you so much!!

PR #1375, opened 2020-12-17
Avoid auto-select for disabled days

m1ky91:
I was about to open an issue but opened this PR through Github without any major context. Feel free to contribute to this PR or take it as an issue.

Example code:

```
<ExpandableCalendar
          disabledByDefault
 />
```

Unexpected behaviour:

gif

### DIFF
--- a/src/expandableCalendar/index.js
+++ b/src/expandableCalendar/index.js
@@ -207,13 +207,13 @@ class ExpandableCalendar extends Component {
   }
 
   getMarkedDates() {
-    const {context, markedDates} = this.props;
+    const {context, markedDates, disabledByDefault} = this.props;
 
     if (markedDates) {
       const marked = _.cloneDeep(markedDates);
       if (marked[context.date]) {
         marked[context.date].selected = true;
-      } else {
+      } else if (!disabledByDefault) {
         marked[context.date] = {selected: true};
       }
       return marked;


m1ky91:
@Inbal-Tish could you please tell me if this is addressed somewhere or could be achieved in another combination of props? Many thanks

src/expandableCalendar/index.js
```diff
@@ -207,13 +207,13 @@ class ExpandableCalendar extends Component {
   }
 
   getMarkedDates() {
-    const {context, markedDates} = this.props;
+    const {context, markedDates, disabledByDefault} = this.props;
```
Inbal-Tish, line 210: You didn't declare the prop and described it. The prop's name is unclear.

src/expandableCalendar/index.js
```diff
@@ -207,13 +207,13 @@ class ExpandableCalendar extends Component {
...
 
     if (markedDates) {
       const marked = _.cloneDeep(markedDates);
       if (marked[context.date]) {
         marked[context.date].selected = true;
-      } else {
+      } else if (!disabledByDefault) {
```
Inbal-Tish, line 216: I understand you don't want the month scroll to autoselect a disabled date, but what it should select instead? Or you mean to leave the calendar with no day selected?

m1ky91: @Inbal-Tish I do think that if we set the prop disabledByDefault and the user changes the month in the calendar, the first day of the month (context.date) should NOT be marked if it's not part of the markedDates. We should leave the month with no day selected so the user knows there are no selectable days in that month. Hope this makes sense

src/expandableCalendar/index.js
Inbal-Tish: Can we check if the `marked[context.date]` is disabled before we apply this prop?

m1ky91: Sorry I don't get you. This prop is inherited from Calendar that should be false by default. If that prop is true we don't want to go in the else block down in the code. If that prop is somehow undefined the behaviour remain as is. If you think something could be improved could you please include suggestions code or contribute to this branch? Thabks

Inbal-Tish: Yes, you're right, but that prop is doing something else and you are adding an extra behaviour on it. It's a breaking change for those who use this prop and don't want your extra behaviuor. Also, I run the code and it doesn't do anything, the date is still being automatically selected when you move to the previous/next month. What am 
I missing?

m1ky91: I see. Thanks for the explanation. However I did this months ago and I am not able to contribute at the moment so I don't know if something has changed meanwhile. I thought this was a bug but since a new prop is required this feature would be much appreciated. Many thanks

Inbal-Tish:
@m1ky91 Ok I understand. So I'll close this PR and you're welcome to open us an issue with a feature request. Thank you!!!

PR #1401, opened 2021-01-29
set dynamically size of knob

dcdaniiel:


Inbal-Tish:
@dcdaniiel Thank you for the PR. I was wondering why not use 'hitSlop' to enlarge the click area? Please provide a description of the use case for this feature.

Issue #1033, opened 2020-01-13
Web (react-native-web) will not support ViewPropTypes API it is upcoming release


vladp:
## Description

React-native-calendars is a very portable package. Some folks, like us, have been using it on web browsers with React-native-web.

With React-native-web release 0.12 and above,  RNW will remove ViewPropTypes export.

This API, unfortunately, is being used by this package, therefore, unless changes are made this will break the web-platform usage.

(tested with most recent release of react-native-calendars 1.221.0  and the upcoming release candidate for RNW -- 0.12.0-rc.1)

jpaas:
Just hit this problem on my upgrade to react-native-web 0.12

nathsimpson:
I have a PR for this, but it's still waiting for review :( https://github.com/wix/react-native-calendars/pull/1089

vladp:
@Inbal-Tish  is that possible to incorporate @nathsimpson PR in?
Or are there wider context issues that prevent the PR from going in?

thx

vladp:
it seems that it was merged yesterday, but not part of the npm release yet

vladp:
@nathsimpson 
The fix is not working for us.
Still getting

```
Creating an optimized production build...
Failed to compile.

./node_modules/react-native-calendars/src/calendar/index.js
Attempted import error: 'ViewPropTypes' is not exported from 'react-native' (imported as 'ReactNative').
```

If I am understanding this correctly, your update checks if ViewPropTypes is defined, and if it is not defined, it aliases to something else.

However webpack (we are using CRA toolchain with config overrides),
seems to 'detect' that ViewPropTypes are still being imported
may be by this line calendars/index.js

`const {View, ViewPropTypes} = ReactNative;
`
Because, if I remove ViewPropTypes from the above line, in this file and in agenda/idnex.js -- the build goes through.

Are you somehow able to tell webpack to ignore the non-existing import?

thank you in advance for any hints

nandorojo:
This is the solution:

```js
// agenda/index.js
import {Text, View, Dimensions, Animated, Platform}  from 'react-native';

...

const viewPropTypes =
  typeof document !== 'undefined' || Platform.OS === 'web'
    ? PropTypes.shape({style: PropTypes.object})
    : require('react-native').ViewPropTypes || View.propTypes;
```

We'd do the same in `calendar/index.js`. This way, we inline require `ViewPropTypes`, only on platforms where it still exists.

[...]

PR #1424, opened 2021-02-25
remove usage of ViewPropTypes

bananer:
`ViewPropTypes` are deprecated in `react-native` ( https://github.com/react-native-community/discussions-and-proposals/issues/29 , https://github.com/react-native-community/discussions-and-proposals/issues/29 ) and were already removed from `react-native-web`.

Supposedly #1236 should fix building with `react-native-web` 12+, however that did not work for me and many others (see #1033). I'm still getting the error `"Attempted import error: 'ViewPropTypes' is not exported from 'react-native' (imported as 'ReactNative')."` 

It seems to me that the usage of `ViewPropTypes` does not add a lot of value anyways, the generic types like `PropTypes.object` from the `prop-types` package are used for most type definitions in `Calendar` and `Agenda`.

Therefore, i propose removing the reliance on `ViewPropTypes` for the only two cases where it was used. This should make the package compatible with `react-native-web` today and a future version of `react-native` with `ViewPropTypes` removed.

### DIFF
--- a/src/agenda/index.js
+++ b/src/agenda/index.js
@@ -2,7 +2,7 @@ import _ from 'lodash';
 import PropTypes from 'prop-types';
 import XDate from 'xdate';
 import React, {Component} from 'react';
-import * as ReactNative from 'react-native';
+import {Text, View, Dimensions, Animated} from 'react-native';
 import {extractComponentProps} from '../component-updater';
 import {parseDate, xdateToData} from '../interface';
 import dateutils from '../dateutils';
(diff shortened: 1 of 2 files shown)


viniciuspetrachin (approved):

stackhousesrl (approved):

src/agenda/index.js
```diff
@@ -34,7 +29,7 @@ export default class AgendaView extends Component {
     ...CalendarList.propTypes,
     ...ReservationList.propTypes,
     /** agenda container style */
-    style: viewPropTypes.style,
+    style: PropTypes.object,
```
Inbal-Tish, line 32: What if I pass style as an array? or StyleSheet ref?

viniciuspetrachin (approved):

src/agenda/index.js
bananer: Thanks for lookong into this. I forgot about arrays of style objects, those would fail the checks. Not sure about style refs, thought they were objects.

Tricky stuff, and not a lot to gain anyways. Would you be ok with `PropTypes.any`?

Inbal-Tish: I have no probblem with 'any' here but isn't it better to limit to `PropTypes.oneOfType([PropTypes.object, PropTypes.array, PropTypes.number])`?

bananer: I updated the PR with your suggestion.

Inbal-Tish (approved):

Issue #457, opened 2018-04-20
Feature: slide the knob up to close the calendars view on agenda


slorber:
Currently it's possible to slide down the knob to open the calendar view.

I think it would be good UX to allow the opposite gesture, with a knob that remains visible at the bottom of the screen even after the calendars view is displayed, so that it can be dragged up.

It would be good also in the knob view to inject a boolean representing this calendars view toggled state. In my case the knob renders an arrow inside it, so if the knob is at the bottom of the screen I want to flip that arrow.

slorber:
It's pretty easy to let the knob visible at the end of the screen visible, and inject the `this.state.calendarScrollable` boolean prop in the renderKnob method. But it seems the knob does not fire "touchStart" when it's at the end of the screen, any idea why?

overengineered:
Animated.ScrollView is a small area on the screen (shown below in red). In current implementation it stays fixed and as user interacts with screen it receives scrolling events.

![calendar-agenda](https://user-images.githubusercontent.com/2990253/39867657-9a822490-545e-11e8-8969-e2905911e400.gif)

You should to change position of it to the bottom of the agenda view for it to handle touches.

ahtesham-shah:
> Animated.ScrollView is a small area on the screen (shown below in red). In current implementation it stays fixed and as user interacts with screen it receives scrolling events.
> 
> ![calendar-agenda](https://user-images.githubusercontent.com/2990253/39867657-9a822490-545e-11e8-8969-e2905911e400.gif)
> 
> You should to change position of it to the bottom of the agenda view for it to handle touches when knob is on the bottom.

@overengineered Sorry but I am not able to do that can you explain in brief.

ahtesham-shah:
@overengineered please help me

overengineered:
src/agenda/index.js has Animated.ScrollView component, which is drawn as a red area in animation above. As you can see it stays fixed in place as user drags knob. The place on the screen for this red area is controlled by scrollPadStyle variable. You want to have 2 options for this style:
 * when calendars is collapsed, draw it in it's current place
 * when calendars is expanded, draw it at the bottom of the agenda

overengineered:
It's not an easy implementation, so I chose to try to explain how it works, instead of attempting to implement it. My hope is that the explanation is clear enough so that someone who really needs this can build it.

angeldua:
Thanks for your reply. What I understood from your comment is the styles
should be different for the scrollPad when in expanded and collapsed state.
[...]

Issue #937, opened 2019-09-24
Agenda toggle knob with a function


elliscwc:
This might related to https://github.com/wix/react-native-calendars/issues/309

I want to show the calendar list with a call of the function in the agenda component, is there a way to do it correctly? Someone has suggested this way but is no longer working
```
function Agenda(props) {
  const agenda = useRef(null);

  useEffect(() => {
    // doesn't work anymore
    agenda.current.setScrollPadPosition(0, true);
    agenda.current.enableCalendarScrolling();
    
    // suggestion?
    const animated = false;
    agenda.current.toggleKnob(animated);
  }, []);

  return (
    <Agenda
      ref={agenda}
      items={[]}
      renderItem={() => <View />}
      renderEmptyDate={() => <View />}
      renderEmptyData={() => <View />}
      rowHasChanged={(r1, r2) => r1.text !== r2.text}
      refreshing={false}
      style={styles.container}
    />
  );
}
```
![3](https://user-images.githubusercontent.com/29953232/65513705-562f7e00-df0e-11e9-8085-c9f24141951c.png)

jraleman:
This will be a cool feature to have!

Inbal-Tish:
@elliscwc Do you mean open/close Agenda's calendar programmatically?

elliscwc:
Yes, @Inbal-Tish. it would be helpful when it can be activated via other buttons on the screen!

flobady:
Hello, I was able to solve my problem with this:
` <Button 
title="knob"
onPress={() => {
          let initPos = refAgenda.current.initialScrollPadPosition()
          refAgenda.current.setScrollPadPosition(initPos, true)}
        }>`

waynedunkley:
For those that find this post and are looking for a solution when using the `ExpanableCalendar` within a `CalendarProvider` as such:
```
<CalendarProvider date={current} onDateChanged={onDateChanged}>
  <ExpandableCalendar
    ref={ref}
    {...props}
  />
</CalendarProvider>
```
My solution was to use the expandableCalendar ref and mock a dayPress
`calendarRef.current.contentRef.onDayPress({ dateString: '2021-03-20' })`

Brayden-kwak:
> const agenda = useRef(null);

could you show me full code that I can fully understand? 
I don't understand where the 'agendaRef' came from. 
Thanks

Brayden-kwak:
[...]

Issue #305, opened 2018-01-15
Knob button hides when i scroll the view


AlbertoIHP:
Please make our job easier by filling this template out to completion. If you're requesting a feature instead of reporting a bug, please feel free to skip the Environment and Reproducible Demo sections.

Hi! , im trying to build my agenda, but when i scrolldown the knob button it dissappear, i need it to stay in the bottom position of the agenda.

![sin titulo](https://user-images.githubusercontent.com/21131515/34959163-f59a62a2-fa13-11e7-820a-360b90a7ce68.png)

Hope you can help me, greetings ...

s123121:
Usability wise, this should be the default behavior, I have my customer asked me about this way too much, they just don't know they have to click on the day to go back

PR #1430, opened 2021-03-05
Use knob to close the Agenda' s calendar view

dtsolis:
This PR resolves issues #457 #305 #937

Adds:

1. The calendar can be closed by dragging the knob upwards.
2. When the user single taps the knob, the calendar will either open or close based on the current state.
3. Exposes `setCalendarOpen` function that can be used to open/close calendar programmatically e.g. `refAgenda.current.setCalendarOpen(false)`

### DIFF
--- a/README.md
+++ b/README.md
@@ -498,6 +498,8 @@ An advanced `Agenda` component that can display interactive listings for calenda
   rowHasChanged={(r1, r2) => {return r1.text !== r2.text}}
   // Hide knob button. Default = false
   hideKnob={true}
+  // When `true` and `hideKnob` prop is `false`, the knob will always be visible and the user will be able to drag the knob up and close the calendar. Default = false
+  knobCanCloseCalendar={false}
   // By default, agenda dates are marked if they have at least one item, but you can override this if needed
   markedDates={{
     '2012-05-16': {selected: true, marked: true},
--- a/README.md
+++ b/README.md
@@ -499,7 +499,7 @@ An advanced `Agenda` component that can display interactive listings for calenda
   // Hide knob button. Default = false
   hideKnob={true}
   // When `true` and `hideKnob` prop is `false`, the knob will always be visible and the user will be able to drag the knob up and close the calendar. Default = false
-  knobCanCloseCalendar={false}
+  showClosingKnob={false}
   // By default, agenda dates are marked if they have at least one item, but you can override this if needed
   markedDates={{
     '2012-05-16': {selected: true, marked: true},
--- a/example/src/screens/agenda.js
+++ b/example/src/screens/agenda.js
@@ -23,6 +23,7 @@ export default class AgendaScreen extends Component {
         renderItem={this.renderItem.bind(this)}
         renderEmptyDate={this.renderEmptyDate.bind(this)}
         rowHasChanged={this.rowHasChanged.bind(this)}
+        knobCanCloseCalendar={true}
         // markingType={'period'}
         // markedDates={{
         //    '2017-05-08': {textColor: '#43515c'},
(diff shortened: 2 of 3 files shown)


Inbal-Tish:
@dtsolis Can you please add screenshots / gif for this feature. I'm not sure I get it as the agenda already opens the calendar on knob press and you can drag to open the calendar with it...

francesco-clementi-92:
This will be quite useful.

I'd like to be able to scroll back the knob to close the calendar list, instead of mandatory select a day from the calendar.

dtsolis:
@Inbal-Tish the PR is about showing the knob while the agenda is open and close on *press/drag upward*. 

github-pr

AlexCatch:
This would be great 👍

README.md
```diff
@@ -498,6 +498,8 @@ An advanced `Agenda` component that can display interactive listings for calenda
   rowHasChanged={(r1, r2) => {return r1.text !== r2.text}}
   // Hide knob button. Default = false
   hideKnob={true}
+  // When `true` and `hideKnob` prop is `false`, the knob will always be visible and the user will be able to drag the knob up and close the calendar. Default = false
+  knobCanCloseCalendar={false}
```
Inbal-Tish, line 502: I don't really like the prop naming... How about something like `showClosingKnob`?

src/agenda/index.js
```diff
@@ -123,12 +125,27 @@ export default class AgendaView extends Component {
     }
   };
 
-  enableCalendarScrolling() {
+  /**
+   * Opens/closes the calendar programmatically.
+   * @param {bool} open `true` when calendar should open. Otherwise `false`.
+   */
+  setCalendarOpen = (open) => {
```
Inbal-Tish, line 132: The comment is uneccessary if the method name was descriptive enough. How about `toggleCalendarPosition(open)`?

src/agenda/index.js
```diff
@@ -123,12 +125,27 @@ export default class AgendaView extends Component {
...
+    const maxY = this.initialScrollPadPosition();
+      if (open) {
+        this.setScrollPadPosition(0, true);
+        this.enableCalendarScrolling(true);
+      } else {
+        this.setScrollPadPosition(maxY, true);
+        this.enableCalendarScrolling(false);
+      }
```
Inbal-Tish, line 140: You're calling the same methods for both positions, only the param change. so why not do something like: `this.setScrollPadPosition(open  ? 0 : maxY, true);
this.enableCalendarScrolling(open);`

src/agenda/index.js
```diff
@@ -253,7 +270,9 @@ export default class AgendaView extends Component {
     this.setScrollPadPosition(snapY, true);
 
     if (snapY === 0) {
-      this.enableCalendarScrolling();
+      this.enableCalendarScrolling(true);
+    } else {
+      this.enableCalendarScrolling(false);
     }
```
Inbal-Tish, line 276: Also here you can shorten the if statement to: `this.enableCalendarScrolling(snapY === 0);`

README.md
dtsolis: Updated!

src/agenda/index.js
dtsolis: Updated!

src/agenda/index.js
dtsolis: Updated!

src/agenda/index.js
dtsolis: Updated!

example/src/screens/agenda.js
```diff
@@ -23,7 +23,7 @@ export default class AgendaScreen extends Component {
         renderItem={this.renderItem.bind(this)}
         renderEmptyDate={this.renderEmptyDate.bind(this)}
         rowHasChanged={this.rowHasChanged.bind(this)}
-        knobCanCloseCalendar={true}
+        showClosingKnob={true}
```
Inbal-Tish, line 26: No need to pass `true`. Passing the prop makes it true.

Inbal-Tish (approved):

PR #1439, opened 2021-03-15
Add gestureStyle prop

mackenziekira:
Allow configuring the outermost gesture component with whatever styles. E.g. I want to be able to pass `{flex: 1}` so that my calendar component fills a screen

Screen Shot 2021-03-15 at 12 15 56 PM

### DIFF
--- a/src/calendar/index.js
+++ b/src/calendar/index.js
@@ -70,7 +70,9 @@ class Calendar extends Component {
     /** Style passed to the header */
     headerStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array]),
     /** Allow rendering of a totally custom header */
-    customHeader: PropTypes.any
+    customHeader: PropTypes.any,
+    /** Style passed to the gesture handler */
+    gestureStyle: viewPropTypes.style
   };
 
   static defaultProps = {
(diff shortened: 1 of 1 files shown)


src/calendar/index.js
```diff
@@ -297,12 +299,12 @@ class Calendar extends Component {
...
 
   render() {
-    const {enableSwipeMonths, style} = this.props;
+    const {enableSwipeMonths, style, gestureStyle} = this.props;
     const GestureComponent = enableSwipeMonths ? GestureRecognizer : View;
     const gestureProps = enableSwipeMonths ? {onSwipe: (direction, state) => this.onSwipe(direction, state)} : {};
 
     return (
-      <GestureComponent {...gestureProps}>
+      <GestureComponent {...gestureProps} style={gestureStyle} >
```
Inbal-Tish, line 307: Why passing the style to the GestureComponent and not to the View below it using the 'style' prop?

mackenziekira: It seemed when testing that all intermediary containers had to also have flex style to behave as expected. E.g. 

```
    <View style={[g.flex, g.backgroundBlue]}>
      <View style={[g.backgroundRed]}>
        <View style={[ g.flex, g.backgroundYellow]} />
      </View>
    </View>
```
results in 
Screen Shot 2021-04-22 at 2 07 45 PM

and it is only when I add the flex style to the middle view that I see what I expect:
```
    <View style={[g.flex, g.backgroundBlue]}>
      <View style={[g.flex, g.backgroundRed]}>
        <View style={[ g.flex, g.backgroundYellow]} />
      </View>
    </View>
```
Screen Shot 2021-04-22 at 2 11 24 PM

Inbal-Tish: Ok. But I'll change the prop name to `containerStyle` as this is the top-level container, while `guesterStyle` is understood only by someone that is familiar with the source code.

PR #1441, opened 2021-03-17
AgendaList - block date updates

Inbal-Tish:
AgendaList - 'blockDateUpdate' to block updates to the CalendarContextProvider when list scrolls

### DIFF
--- a/src/expandableCalendar/agendaList.js
+++ b/src/expandableCalendar/agendaList.js
@@ -35,7 +35,7 @@ class AgendaList extends Component {
     /** style passed to the section view */
     sectionStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array]),
     /** whether to block the date change in calendar (and calendar context provider) when agenda scrolls */
-    blockDateUpdate: PropTypes.bool
+    blockDateUpdates: PropTypes.bool
   }
 
   static defaultProps = {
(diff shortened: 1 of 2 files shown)


src/expandableCalendar/agendaList.js
```diff
@@ -33,7 +33,9 @@ class AgendaList extends Component {
     /** whether to mark today's title with the "Today, ..." string. Default = true */
     markToday: PropTypes.bool,
     /** style passed to the section view */
-    sectionStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array])
+    sectionStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array]),
+    /** whether to block the date change in calendar (and calendar context provider) when agenda scrolls */
+    blockDateUpdates: PropTypes.bool
```
ethanshar, line 38: ok, so know that I understand the purpose of this prop. 
The new prop name is a little confusing, what about avoidDateUpdates? WDYT?

Inbal-Tish: Yes, I wasn't sure about the naming either. Sounds good.

ethanshar (approved):

talhauzair-28:
can we have this feature for WeekCalendar as well?

PR #1444, opened 2021-03-17
Update `currentMonth` if `current` prop changes

mackenziekira:
Let calendar be a managed component if user passes in `current` prop.

E.g. we have a filter screen that lets the user select a month to jump to:
Screen Shot 2021-03-17 at 2 43 56 PM

So it requires us to be able to control which date the calendar is showing, if we don't want to force unmounting and remounting the component.

### DIFF
--- a/src/calendar/index.js
+++ b/src/calendar/index.js
@@ -88,6 +88,12 @@ class Calendar extends Component {
 
     this.shouldComponentUpdate = shouldComponentUpdate;
   }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (this.props.current && prevProps.current !== this.props.current) {
+      this.setState({currentMonth: parseDate(this.props.current)})
+    }
+  }
   
   addMonth = count => {
     this.updateMonth(this.state.currentMonth.clone().addMonths(count, true));


src/calendar/index.js
```diff
@@ -88,6 +88,12 @@ class Calendar extends Component {
 
     this.shouldComponentUpdate = shouldComponentUpdate;
   }
+
+  componentDidUpdate(prevProps, prevState) {
```
Inbal-Tish, line 92: @ questions: 
1) Why not use `shouldComponentUpdate` as with all other props instead of adding the 'componentDidUpdate' method just for it?
2) When you what to update the calendar with a new 'current', did you try to pass the 'key' prop with this dynamic data?

mackenziekira: Hm, I'm not sure I follow the idea of a `key` prop, I don't see that prop listed in the prop type definitions.

My understanding of `shouldComponentUpdate` is that it controls whether the component should rerender or not, but re-rendering in this case would not have an effect on `currentMonth` in state, because that bit of state is only set in the constructor (before the component mounts) or when responding to a touch event

Inbal-Tish: `key` is a React component prop that should trigger rerender when changed.

PR #1446, opened 2021-03-18
moment js as an optional dependency

radekrepka:
Moment.js is a legacy project in maintenance mode now and this repo use xdate library for dates.

### DIFF
--- a/example/src/screens/calendars.js
+++ b/example/src/screens/calendars.js
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import moment from 'moment';
+import XDate from 'xdate';
 import React, {useState, Fragment} from 'react';
 import {StyleSheet, View, ScrollView, Text, TouchableOpacity, Switch} from 'react-native';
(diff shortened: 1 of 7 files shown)


Inbal-Tish:
@radekrepka I love this PR! I'm all in favor of removing 'moment'.
One issue though: removing `useMoment` from AgendaList is a breaking change. So we can do it in V2 or make 'moment' an optional dependency till then. What do you think?

NS-BOBBY-C:
Could this include an option to keep moment? I'm not sure what happens under the hood, but I can see people still wanting to use moment (namely, me! :)).

Inbal-Tish:
> Could this include an option to keep moment? I'm not sure what happens under the hood, but I can see people still wanting to use moment (namely, me! :)).

Are you using `useMoment` prop or passing a moment date object?

radekrepka:
okey, moment is an optional dependency now

Inbal-Tish (approved):

PR #1449, opened 2021-03-24
Fix issue with reversed scroll offset values on Android RTL

ethanshar:
On Android RTL there's a known issue with Scroll components returning wrong offset values. 
The value is actually reversed, which cause lots of miscalculations. 
The fix reverse the offset value back. 

Another issue I noticed (that is probably related to the whole RTL mess) is that on Android RTL there's an initial unexpected scroll event. 
I disabled this unwanted event to avoid weird issues on mounting

### DIFF
--- a/example/src/app.js
+++ b/example/src/app.js
@@ -1,9 +1,11 @@
 import './wdyr'; // <--- must be first import
 
 import {Navigation} from 'react-native-navigation';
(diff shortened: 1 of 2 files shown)


Inbal-Tish (approved):

Inbal-Tish:
@ethanshar Looks really great. Thanks!!!

PR #1460, opened 2021-04-12
Fix first item blank space

elratonmaton:


Inbal-Tish:
@elratonmaton Can you please provide a description of the issue you're fixing, your solution, and screenshots?

PR #1474, opened 2021-04-27
CalendarList - Adding `animateScroll` prop

Inbal-Tish:
CalendarList - Adding `animateScroll` prop to allow animating auto-scroll (i.e. on arrow press).


ethanshar (approved):

738:
Cool

PR #1481, opened 2021-05-03
Reduce wasteful renders by memoizing props passed to week component

ethanshar:
Few props triggered wasteful render in the Week component that being rendered in the WeekCalendar component. 
- style 
- context
- markedOptions

I fixed `style` by memoizing its value, this is almost never change so the memoization here make sense. 
For `context`, I used `extractComponentProps` util to avoid passing irrelevant props to Week component
And last, most complicated one, `markedDates` I noticed that it changes usually only for the present/current week. 
So what I did is pass the "fixed" markedDates only for the current week. 

This fix prevents rendering of all weeks when changing selected date in the current week

### DIFF
--- a/src/expandableCalendar/weekCalendar.js
+++ b/src/expandableCalendar/weekCalendar.js
@@ -182,7 +182,7 @@ class WeekCalendar extends Component {
   };
 
   renderItem = ({item}) => {
-    const {calendarWidth, style, onDayPress, markedDates, ...others} = extractComponentProps(Week, this.props);
+    const {style, onDayPress, markedDates, ...others} = extractComponentProps(Week, this.props);
 
     const weekDates = getWeekDates(item, others.firstDay, 'yyyy-MM-dd');
     const currentWeek = _.includes(weekDates, this.props.context.date);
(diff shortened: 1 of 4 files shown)


src/expandableCalendar/weekCalendar.js
```diff
@@ -177,15 +182,19 @@ class WeekCalendar extends Component {
   };
 
   renderItem = ({item}) => {
-    const {calendarWidth, style, onDayPress, ...others} = this.props;
+    const {calendarWidth, style, onDayPress, markedDates, ...others} = extractComponentProps(Week, this.props);
```
Inbal-Tish, line 185: `calendarWidth` is not a Week prop... It is a WeekCalendar prop

Inbal-Tish: But I don't think you need it here as `this.containerWidth` already reads it

ethanshar: Yea you right.. the previous code was `calendarWidth || this.containerWidth` 
I didn't notice it's a getter
Ill fix it

ethanshar: Fixed ✅

Inbal-Tish (approved):

PR #1482, opened 2021-05-03
avoid updating WeekCalendar when stayin on the same week

ethanshar:
In WeekCalendar component, the `componentDidUpdate` trigger a state change when it's not necessary. 
In the case I change between days on the same week by pressing, there's no need to update the items of the week

### DIFF
--- a/src/expandableCalendar/weekCalendar.js
+++ b/src/expandableCalendar/weekCalendar.js
@@ -57,9 +57,13 @@ class WeekCalendar extends Component {
   }
 
   componentDidUpdate(prevProps) {
-    const {updateSource, date} = this.props.context;
+    const {firstDay, context} = this.props;
+    const {updateSource, date, prevDate} = context;
 
-    if (date !== prevProps.context.date && updateSource !== UPDATE_SOURCES.WEEK_SCROLL) {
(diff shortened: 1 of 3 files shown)


src/expandableCalendar/weekCalendar.js
```diff
@@ -57,9 +57,13 @@ class WeekCalendar extends Component {
   }
 
   componentDidUpdate(prevProps) {
-    const {updateSource, date} = this.props.context;
+    const {firstDay, context} = this.props;
+    const {updateSource, date, prevDate} = context;
 
-    if (date !== prevProps.context.date && updateSource !== UPDATE_SOURCES.WEEK_SCROLL) {
+    const weekDates = getWeekDates(date, firstDay, 'yyyy-MM-dd');
+    const isSameWeek = _.includes(weekDates, prevDate);
```
Inbal-Tish, line 64: Can we move `isSameWeek` to `dateutils` file?

Inbal-Tish: Then we can move line #188 from WeekCalendar to that as well...

ethanshar: Done

Inbal-Tish (approved):

PR #1483, opened 2021-05-03
Fix the markedDates we pass to WeekCalendar

ethanshar:
In my previous PR, in WeekCalendar, I added an optimization to when I pass on markedDates to avoid unnecessary render. 

In ExpandableCalendar we call `this.getMarkedDates` and pass the result to WeekCalendar and then in WeekCalendar we call it again only if this is the current week. 
But because ExpandableCalendar go thru `this.getMarkedDates` anyway, this optimization is lost. 

Anyway, see the small fix I did, it should fix this issue

### DIFF
--- a/src/expandableCalendar/index.js
+++ b/src/expandableCalendar/index.js
@@ -212,6 +212,7 @@ class ExpandableCalendar extends Component {
     return days.length / 7;
   }
 
+  // TODO: this logic repeat itself in WeekCalendar - consider moving to a presenter
   getMarkedDates() {
     const {context, markedDates} = this.props;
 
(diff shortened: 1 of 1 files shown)


src/expandableCalendar/index.js
```diff
@@ -212,6 +212,7 @@ class ExpandableCalendar extends Component {
     return days.length / 7;
   }
 
+  // TODO: this logic repeat itself in WeekCalendar - consider moving to a presenter
```
Inbal-Tish, line 215: We should do that, this way it will be clearer what is the aim of passing only the `props.markedDates` instead of `getMarkedDates` from ExpandableCalendar.

ethanshar: I created a Card in the board for it. 
We need to address markedDates differently. 
Let's address it separately, I want to push this PR

Inbal-Tish (approved):

PR #1487, opened 2021-05-06
Fix: Arrow press callbacks on expandable calendar

cjhines:
Fix the broken prop drilling on the `onPressArrowLeft` and `onPressArrowRight` props for  the `ExpandableCalendar`.

See the issue commented below.


cjhines:
https://github.com/wix/react-native-calendars/issues/1488

Issue #1488, opened 2021-05-06
[PR included] Expandable calendar does not handle onPressArrowLeft / onPressArrowRight


cjhines:
## Description

Expandable calendar passes it's own `onPressArrowLeft` and `onPressArrowRight` props to `CalendarList` which ignores any custom props from the user:

https://github.com/wix/react-native-calendars/blob/70513c9faebf8bcee5bfbefb7bd9480e527d3b08/src/expandableCalendar/index.js#L352-L354

https://github.com/wix/react-native-calendars/blob/70513c9faebf8bcee5bfbefb7bd9480e527d3b08/src/expandableCalendar/index.js#L356-L358

## Expected Behavior

`onPressArrowLeft` and `onPressArrowRight` callbacks are triggered on press.

## Observed Behavior

Callbacks are not fired.

## Pull request

https://github.com/wix/react-native-calendars/pull/1489

cjhines:
Pull request: https://github.com/wix/react-native-calendars/pull/1487

I would appreciate hasty consideration of this PR, as this is a blocking issue for a feature release.

PR #1489, opened 2021-05-06
Fix: Arrow press callbacks on expandable calendar

cjhines:
Resolve https://github.com/wix/react-native-calendars/issues/1488.

This is a release blocking issue for us sadly, so any feedback would be hugely appreciated.

As prop drilling is not happening for `onPressArrowLeft` and `onPressArrowRight` in  the `ExpandableCalendar` component, it should be included in the custom function which are being passed.

As a result, it is not possible to use these props in the `ExpandableCalendar` component. They just aren't called.

### DIFF
--- a/src/expandableCalendar/index.js
+++ b/src/expandableCalendar/index.js
@@ -350,10 +350,18 @@ class ExpandableCalendar extends Component {
   /** Events */
 
   onPressArrowLeft = () => {
+    const {onPressArrowLeft} = this.props;
+    if (onPressArrowLeft) {
+      onPressArrowLeft();
+    }
     this.scrollPage(false);
   };
 
   onPressArrowRight = () => {
+    const {onPressArrowLeft} = this.props;
+    if (onPressArrowLeft) {
+      onPressArrowLeft();
+    }
     this.scrollPage(true);
   };
 
--- a/src/expandableCalendar/index.js
+++ b/src/expandableCalendar/index.js
@@ -350,18 +350,12 @@ class ExpandableCalendar extends Component {
   /** Events */
 
   onPressArrowLeft = () => {
-    const {onPressArrowLeft} = this.props;
-    if (onPressArrowLeft) {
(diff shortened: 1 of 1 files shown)


cjhines:
https://github.com/wix/react-native-calendars/issues/1488

cjhines:
I created a new PR without the linting issues. Please let me know if there are any further issues.

cjhines:
E2E tests seem to be failing on master, as well as all PRs.

cjhines:
Merged E2E fix into this branch. Checks are now passing.

cjhines:
Would really appreciate that this gets looked at. This bug is currently a blocker for me.

src/expandableCalendar/index.js
```diff
@@ -350,10 +350,18 @@ class ExpandableCalendar extends Component {
   /** Events */
 
   onPressArrowLeft = () => {
+    const {onPressArrowLeft} = this.props;
+    if (onPressArrowLeft) {
```
Inbal-Tish, line 354: Please use `_.invoke` instead of if statement

src/expandableCalendar/index.js
```diff
@@ -350,10 +350,18 @@ class ExpandableCalendar extends Component {
...
+      onPressArrowLeft();
+    }
     this.scrollPage(false);
   };
 
   onPressArrowRight = () => {
+    const {onPressArrowLeft} = this.props;
+    if (onPressArrowLeft) {
```
Inbal-Tish, line 362: Please use `_.invoke` instead of if statement

src/expandableCalendar/index.js
cjhines: @Inbal-Tish Done.

src/expandableCalendar/index.js
cjhines: @Inbal-Tish Done.

Inbal-Tish (approved):

PR #1497, opened 2021-05-07
arreglando bug que afectaba el calculo de las posiciones en el timeline

Dayessi:

PR #1498, opened 2021-05-10
Fix timeline event height and top calculations

KristinaHus:
I've met a problem where my events did not appear on the timeline calendar. I've dug into the code and found the `buildEvent` function which calculates event height and top in the wrong way and always returns negative values.

Screenshot 2021-05-10 at 12 43 15


alierenkara (approved):

vingrad (approved):
It looks good!

sggonzalezg (approved):

Inbal-Tish:
Duplication of #1497

Inbal-Tish (approved):

PR #1505, opened 2021-05-13
Add years change 

CHNB128:


Inbal-Tish:
@CHNB128 Hi. Thank you for the PR. I don't understand how a day press should invoke a year change. Can you add a video with the example please?

PR #1519, opened 2021-05-27
feat: Do not disable dates on different months

benomatis:
Whether or not dates should be disabled should be decided by marked dates, showing otherwise available dates in the next or previous month as disabled is not really a good user experience, because they're not disabled, they're just days of a different month.


benomatis:
Ok, makes sense @Inbal-Tish, thank you.

PR #1522, opened 2021-05-28
Add nestedScrollEnabled props

Gueoff:
this PR adds the nestedScrollEnabled prop to allows the calendar-list component to be display in another scrollView in Android.


src/calendar-list/index.js
```diff
@@ -59,7 +59,9 @@ class CalendarList extends Component {
     /** How far from the end to trigger the onEndReached callback */
     onEndReachedThreshold: PropTypes.number,
     /** Called once when the scroll position gets within onEndReachedThreshold */
-    onEndReached: PropTypes.func
+    onEndReached: PropTypes.func,
+    /** Allows FlatList to be displayed on scrollView */
```
Inbal-Tish, line 63: That's a ScrollView prop (not FlatList's) and its description reads "Enables nested scrolling for Android API level 21+". Please mantion it.

src/calendar-list/index.tsx
```diff
@@ -129,7 +131,8 @@ class CalendarList extends Component<Props, State> {
     scrollsToTop: false,
     scrollEnabled: true,
     removeClippedSubviews: Platform.OS === 'android',
-    keyExtractor: (_: any, index: number) => String(index)
+    keyExtractor: (_: any, index: number) => String(index),
+    nestedScrollEnabled: true
```
Inbal-Tish, line 135: Why the default is true? Since it's a new prop, it should be false

Inbal-Tish:
@Gueoff Sorry for the late replay. I just have a small comment. Hope you can address it and we can merge this one. Thanks!

src/calendar-list/index.tsx
Gueoff: > Why the default is true? Since it's a new prop, it should be false

The default value is true because it set to true in the ScrollView from react-native too
https://reactnative.dev/docs/scrollview#nestedscrollenabled-android

Gueoff:
> @Gueoff Sorry for the late reply. I just have a small comment. Hope you can address it and we can merge this one. Thanks!

@Inbal-Tish sorry I'm late too :p I think we can merge like this if you are ok with my reply

Gueoff:
@Inbal-Tish up to my previous comment

Inbal-Tish (approved):

PR #1539, opened 2021-06-21
Setup typescript

ethanshar:
This PR is an initial setup of typescript for this library. 
I added a basic tsconfig.json file that currently mapped on a specific file.

Also, I migrated a single component for now the basic/day component. 

I verified the published snapshot version of this changes and it works fine in private uilib - meaning, the components work and the typescript build of uilib is not failing.

**How should we move forward**
The idea is to merge it and gradually migrate files to TS like we did in uilib. 
When migrating a new file, it should be added to tsconfig "include" array. 
Also, we should make sure that renaming files (js -> ts) should keep the git history of the file


src/calendar/day/basic/index.tsx
```diff
@@ -8,7 +8,29 @@ import {shouldUpdate} from '../../../component-updater';
 import styleConstructor from './style';
 import Marking from '../marking';
 
-export default class BasicDay extends Component {
+interface BasicDayProps {
```
Inbal-Tish, line 11: Don't we need to export the props for the components that use them, like Calendar (line #31: `...Day.propTypes`)?

ethanshar: Well yea, we can also do it when we migrate the Calendar component to typescript...

Inbal-Tish (approved):

Issue #1537, opened 2021-06-18
Month change is auto selecting the first day of every month in ExpandableCalendar


joseortiz9:
## Description
As the title says, using the ExpandableCalendar every time that the gesture of changing the month or with the arrows the selectedDate is being set automatically to the first day of every month, same thing happens in the weekCalendar view when changing of week, it selects the first day of the week.

## Expected Behavior
Not change automatically the selectedDate and let the last selected of the previous month.

## Environment

* `react-native-calendars`: react-native-calendars@1.1263.0
* `react-native`: react-native@0.63.2 

Also specify:

1. Phone/emulator/simulator & version:
Phone: Galaxy A9 (2018)
Android: 10

## Code
```
<CalendarProvider
    date={selectedDate}
    onDateChanged={(date) => setSelectedDate(date)}
    onMonthChange={({month, year}) => setMonth({month, year})}
>
    <ExpandableCalendar
      firstDay={1}
      //current={selectedDate}
      markedDates={{}}
      displayLoadingIndicator
      showScrollIndicator
      horizontal
      pagingEnabled
      hideExtraDays={false}
      hideArrows={false}

      initialPosition={ExpandableCalendar.positions.OPEN}
      disableWeekScroll={true}

      pastScrollRange={20}
      futureScrollRange={20}
      theme={{}}
  />
</CalendarProvider>
```

## Reproducible Demo
I created an snack to prove it, it works okay on web, but on ios and android it happens the issue. 
https://snack.expo.io/@joseortiz09/bug-expandablecalendar-on-month-change

https://user-images.githubusercontent.com/26872450/122513025-684d2c00-cfcf-11eb-9ecd-b0295eb0897f.mp4

Please help me out!, your library is amazing but need more work :(

Inbal-Tish:
@joseortiz9 Hi. This is not a bug. It's is the intended behaviour. If you wish to toggle the feature you're welcome to submit a feature request or a PR.

joseortiz9:
how come this is an intended behavior? the selectedDay is supposed to change on every month? so what is the purpose of selecting it (?)

wikramdas:
Having the same issue. Whenever I changed the month first date of every month has been auto selected.
Any solution?

joseortiz9:
[...]

PR #1545, opened 2021-06-25
Fix build process of TS

ethanshar:
Since the current published version of calendars include ts files it causes issues for dependent modules' tests. 
In order to fix it, I updated the build process (it similar now to what we have in uilib)

- I added a `build:ts` command for building typescript 
- The build generates `js/jsx` files out `ts/tsx` files and also add appropriate `d.ts` files (for files that were migrated)
- In `.npmignore` I ignore `ts/tsx` files so they won't be included in the published version of calendars lib
- In our Jenkins configuration I added the `build:ts` step - meaning that from now on if there are TS build issues the build will fail (which is good)

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -256,3 +256,9 @@ demo/
 # eslint-rules/
 scripts/
 
+# Typescript
+# ignore the .ts files
+*.ts
+*.tsx
+# include the .d.ts files
+!*.d.ts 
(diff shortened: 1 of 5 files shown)


joseortiz9:
can you guys please check this issue. I think is a big error  [#1537](https://github.com/wix/react-native-calendars/issues/1537#issue-924571574)

Inbal-Tish (approved):

PR #1553, opened 2021-07-01
change import to require - Cannot assign to read only property 'exports' of object '#<Object>'

Moe-kanan:
Don't mix import and module.exports in the same file:
[https://github.com/webpack/webpack/issues/4039#issuecomment-419284940](https://github.com/webpack/webpack/issues/4039#issuecomment-419284940)

PR to resolve this error: `Cannot assign to read only property 'exports' of object '#<Object>'`


Moe-kanan:
@Inbal-Tish

Inbal-Tish (approved):

PR #1555, opened 2021-07-05
TS/ calendar day dot

lidord-wix:


src/calendar/day/dot/index.tsx
```diff
@@ -1,9 +1,17 @@
-import PropTypes from 'prop-types';
 import React from 'react';
 import {View} from 'react-native';
 import styleConstructor from './style';
 
-const Dot = ({theme, marked, disabled, color, today, selected}) => {
+interface DotProps {
+  theme: Object;
```
ethanshar, line 6: The props should be optional (:

lidord-wix: Fixed :)

ethanshar (approved):

PR #1564, opened 2021-07-06
Ts/calendar day

lidord-wix:
Finish calendar day migration to TS.
Return the dot propTypes (it was necessary).


src/calendar/day/dot/index.tsx
```diff
@@ -1,3 +1,4 @@
+import PropTypes from 'prop-types';
```
Inbal-Tish, line 1: Why are we using prop-types here?

src/calendar/day/index.tsx
```diff
@@ -37,7 +47,7 @@ export default class Day extends Component {
     ]);
   }
 
-  getMarkingLabel(marking) {
+  getMarkingLabel(marking: any) {
```
Inbal-Tish, line 50: Why not object?

src/calendar/day/index.tsx
```diff
@@ -68,8 +78,8 @@ export default class Day extends Component {
   }
 
   getAccessibilityLabel = memoize((day, marking, isToday) => {
-    const today = XDate.locales[XDate.defaultLocale].today;
-    const formatAccessibilityLabel = XDate.locales[XDate.defaultLocale].formatAccessibilityLabel;
+    const today = _.get(XDate, 'locales.[XDate.defaultLocale].today');
+    const formatAccessibilityLabel = _.get(XDate, 'locales[XDate.defaultLocale].formatAccessibilityLabel');
```
Inbal-Tish, line 82: Did you test these changes? I see that the first one includes `locales**.**[XDate.defaultLocale]` and the other one `locales[XDate.defaultLocale]`...

src/calendar/day/dot/index.tsx
Inbal-Tish: `marking` is using Dot props. We should include Marking component in this PR...

lidord-wix: The `extractComponentProps` function (src/component-updater.js) is based on the prop-types, when removing it the unit test are failling. So we decided to keep it until we'll make the relevant adjustment..

src/calendar/day/index.tsx
lidord-wix: Actually setting it to `object` is almost like setting it to `any` if we don't write the object keys.
And I don't really know the possible keys for `marking`

src/calendar/day/index.tsx
lidord-wix: fixed :)

src/calendar/day/dot/index.tsx
Inbal-Tish: Ok

src/calendar/day/index.tsx
Inbal-Tish: I disagree. 'object' is more restrictive than 'any' type and indicates that the value is a non-primitive.

src/calendar/day/index.tsx
```diff
@@ -4,28 +4,39 @@ import XDate from 'xdate';
...
+// @ts-expect-error
 import {xdateToData} from '../../interface';
+// @ts-expect-error
 import {SELECT_DATE_SLOT} from '../../testIDs';
-import BasicDay from './basic';
+import BasicDay, {BasicDayProps} from './basic';
 import PeriodDay from './period';
+import {MarkingProps} from './marking';
+
+const basicDayPropsTypes = _.omit(BasicDay.propTypes, 'date');
```
Inbal-Tish, line 19: Why the change of name if you're not using `BasicDayProps` to get the 'date'?

lidord-wix: I found it confusing to have both `basicDayProps` and `BasicDayProps`

Inbal-Tish: I understand but why are you still using `BasicDay.propTypes`?

lidord-wix: This is for the Day propTypes - as I mentioned, we decided to keep all the propTypes for now...

Inbal-Tish: I understand why you're keeping the propTypes, I'm not sure why you're using it here...

lidord-wix: The Day propTypes is using the BasicDay.propTypes (as before) and the DayProps is using the BasicDayProps.

Inbal-Tish (approved):

PR #1567, opened 2021-07-07
TS/calendar header

lidord-wix:
Migrate to TS


src/calendar/header/index.tsx
```diff
@@ -111,7 +150,7 @@ class CalendarHeader extends Component {
   renderWeekDays = memoize(weekDaysNames => {
     const {disabledDaysIndexes} = this.props;
 
-    return weekDaysNames.map((day, idx) => {
+    return weekDaysNames.map((day: any, idx: any) => {
```
Inbal-Tish, line 153: Why 'any'? 'day' is a string and 'idx' is a number...

src/calendar/header/index.tsx
```diff
@@ -3,21 +3,59 @@ import PropTypes from 'prop-types';
...
   HEADER_LOADING_INDICATOR,
   HEADER_MONTH_NAME
+  // @ts-expect-error
 } from '../../testIDs';
 import styleConstructor from './style';
 
-class CalendarHeader extends Component {
+type Direction = 'left' | 'right';
+interface CalendarHeaderProps {
+  theme?: any;
```
Inbal-Tish, line 24: Why 'theme' is any and not object type?

src/calendar/header/index.tsx
```diff
@@ -3,21 +3,59 @@ import PropTypes from 'prop-types';
...
+  /** Handler which gets executed when press arrow icon right. It receive a callback can go next month */
+  onPressArrowRight?: (method: () => void, month?: XDate) => void;
+  /** Disable left arrow. Default = false */
+  disableArrowLeft?: boolean;
+  /** Disable right arrow. Default = false */
+  disableArrowRight?: boolean;
+  /** Apply custom disable color to selected day indexes */
+  disabledDaysIndexes?: number[];
+  /** Replace default month and year title with custom one. the function receive a date as parameter. */
+  renderHeader?: any;
```
Inbal-Tish, line 49: 'renderHeader' is a function...

src/calendar/header/style.ts
```diff
@@ -1,9 +1,10 @@
 import {StyleSheet, Platform} from 'react-native';
+// @ts-expect-error
 import * as defaultStyle from '../../style';
 
 const STYLESHEET_ID = 'stylesheet.calendar.header';
 
-export default function (theme = {}) {
+export default function (theme: any = {}) {
```
Inbal-Tish, line 7: here too. Why not object type?

src/calendar/header/index.tsx
```diff
@@ -3,21 +3,59 @@ import PropTypes from 'prop-types';
...
   HEADER_DAY_NAMES,
   HEADER_LOADING_INDICATOR,
   HEADER_MONTH_NAME
+  // @ts-expect-error
 } from '../../testIDs';
 import styleConstructor from './style';
 
-class CalendarHeader extends Component {
+type Direction = 'left' | 'right';
+interface CalendarHeaderProps {
```
Inbal-Tish, line 23: Don't we need to export `CalendarHeaderProps`? It will be used in Calendar component props...

src/calendar/header/style.ts
lidord-wix: Changing it to object causes a ts error since it's trying to get `theme[STYLESHEET_ID]` but it doesn't recognize it from the general object type. 
I can change it to `_.get(theme, STYLESHEET_ID)` but I'm not sure it's better.. wdyt?

src/calendar/header/index.tsx
lidord-wix: Changing it to `object` causes a ts error since it's trying to get `theme.indicatorColor` but it doesn't recognize it from the general object type.
I can change it to `_.get(theme, indicatorColor)` but I'm not sure it's better.. wdyt?

Inbal-Tish: Maybe it's better to define a `theme` type in a common service?

src/calendar/header/style.ts
Inbal-Tish: Since this will repeat itself all over the library I suggest we'll define a `theme` type.

src/calendar/day/period/index.tsx
```diff
@@ -130,8 +132,8 @@ export default class PeriodDay extends Component<PeriodDayProps>{
     const {state, marking} = this.props;
     const containerStyle = [this.style.base];
     const textStyle = [this.style.text];
-    let leftFillerStyle = {backgroundColor: undefined};
-    let rightFillerStyle = {backgroundColor: undefined};
+    let leftFillerStyle: {backgroundColor?: string} = {backgroundColor: undefined};
+    let rightFillerStyle: {backgroundColor?: string} = {backgroundColor: undefined};
```
Inbal-Tish, line 136: This type is going to limit the style to have only 'backgroundColor' key... I think this is a breaking change. Did you test it?

src/commons/types.ts
```diff
@@ -0,0 +1,25 @@
+import {ColorValue} from 'react-native';
+export interface Theme {
```
Inbal-Tish, line 2: Where did you get these keys from? I think there are tones of keys missing...

src/calendar/day/period/index.tsx
lidord-wix: I did test it.. anyway - I changed it to ViewStyle

src/commons/types.ts
lidord-wix: I tried to look for usages in the code..
I've added more keys, let me know if something is still missing :)

Inbal-Tish: You can find most keys in `src/style.js`

lidord-wix: Tnx! added :)

Inbal-Tish (approved):

PR #1581, opened 2021-07-20
Rn 64 upgrade

Inbal-Tish:
upgrading react-native to 0.64.1


ethanshar:
@Inbal-Tish 
Build is red

Inbal-Tish:
> @Inbal-Tish
> Build is red

I know

PR #1582, opened 2021-07-20
make module tree shakable

EyMaddis:
Webpack would include moment.js and other unnecessary files if they bundle the library. This is especially relevant for web builds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "1.22.0",
   "main": "src/index.js",
   "description": "React Native Calendar Components",
+  "sideEffects": false,
   "scripts": {
     "build": "xcodebuild -project ios/CalendarsExample.xcodeproj build",
     "build:ts": "tsc",


EyMaddis:
@Inbal-Tish Could you merge this/take a look? :) 
Without this, react-native-web users will have very bloated js bundles

Inbal-Tish:
@EyMaddis Hi. I understand the pros of removing all side effects and their effect on the bundle size. What I'm not sure of is if we indeed don't need all side effects (that will only execute by the import, which might be removed this way). Did you test the library with this key on both web and mobile apps?

EyMaddis:
@Inbal-Tish Hi! Thanks for looking at it!

We have the calendar in production use with a branch that has this set and it works flawlessly. However we do not use the agenda, but for me the source code is pretty clear and I think only moment is getting cut away:

https://github.com/wix/react-native-calendars/blob/8aed7509b6901293d0810131dcfa97cbf92ad484/src/momentResolver.ts#L3-L10

Issue #17, opened 2017-05-18
undefined is not and object (evaluating '_props[registrationName]')


manrashid:
On android with RN0.43.4. Just trying out the example.
I was using exactly the same code as the example in AgendaScreen. When I press the date on top of the agenda list, the above error shown. 

I've found that only when I touch the Text area the error will occur. Try avoid the text when touching the date and there will be NO error.

Probably related to this http://stackoverflow.com/questions/43674781/undefined-is-not-and-object-evaluating-propsregistrationname

dongawli:
@tautvilas : I am too facing this issue on Android Emulator - https://hub.mangoapps.com/sf/ODcxMTlfMTMyNzAwMA
Can you please help us how to resolve this?

tautvilas:
I will try to look at this problem today

tautvilas:
I tried adding the stackoverflow suggested workaround. Check if that fixes the problem for you in version v1.2.14 (I did not test with RN43 and in RN44 I could not reproduce this bug)

dongawli:
Thank you. 

Yes, issue is resolved now.

PR #1583, opened 2021-07-22
Calendar to type script

Inbal-Tish:
Calendar - migrate to ts


src/calendar-list/index.tsx
```diff
@@ -138,18 +140,19 @@ class CalendarList extends Component<CalendarListProps, CalendarListState> {
     const rows = [];
     const texts = [];
     const date = parseDate(props.current) || new XDate();
+    const {pastScrollRange = 50, futureScrollRange = 50} = props;
```
ethanshar, line 143: I see we're setting default prop value for pastScrollRange and futureScrollRange in many places in this file. 
I know it's for typescript cause TS doesn't understand `static defaultProps` anyway let's at least define a const for these default values at the top of the file.

src/calendar-list/index.tsx
```diff
@@ -192,7 +195,7 @@ class CalendarList extends Component<CalendarListProps, CalendarListState> {
   }
 
   scrollToDay(d: XDate, offset: number, animated: boolean) {
-    const {horizontal, calendarHeight, calendarWidth, pastScrollRange, firstDay} = this.props;
+    const {horizontal, calendarHeight = 360, calendarWidth = width, pastScrollRange = 50, firstDay} = this.props;
```
ethanshar, line 198: same here, let's declare const values at the top file for better readability..

src/calendar-list/index.tsx
```diff
@@ -348,10 +354,10 @@ class CalendarList extends Component<CalendarListProps, CalendarListState> {
           getItemLayout={this.getItemLayout}
           onViewableItemsChanged={this.onViewableItemsChanged}
           viewabilityConfig={this.viewabilityConfig}
-          initialScrollIndex={this.state.openDate ? this.getMonthIndex(this.state.openDate) : false}
+          initialScrollIndex={this.state.openDate ? this.getMonthIndex(this.state.openDate) : undefined}
           showsVerticalScrollIndicator={showScrollIndicator}
           showsHorizontalScrollIndicator={horizontal && showScrollIndicator}
-          testID={testID}
+          testID={this.props.testID}
```
ethanshar, line 360: Why this change? 
why not destructing it from `this.props` as it was..

src/calendar-list/item.tsx
```diff
@@ -101,6 +101,7 @@ class CalendarListItem extends Component<CalendarListItemProps, CalendarListItem
       headerStyle,
       onPressArrowLeft,
       onPressArrowRight,
+      //@ts-ignore
```
ethanshar, line 104: Consider using `// @ts-expect-error` if this will be fixed eventually.
Maybe once we'll move other files this will resolve itself and it will notify us about it

src/calendar/day/basic/index.tsx
```diff
@@ -75,13 +75,7 @@ export default class BasicDay extends Component<BasicDayProps> {
   };
 
   get marking() {
-    let marking = this.props.marking || {};
```
ethanshar, line 78: It appears this code used to handle a case when marking is an array. 
Are we sure we want to remove it? 
If we're keeping it, there's definitely cleaner way to check for array

src/calendar/day/marking/index.tsx
```diff
@@ -45,14 +50,16 @@ export interface MarkingProps extends DotProps {
   startingDay?: boolean;
   endingDay?: boolean;
   accessibilityLabel?: string;
+  customStyles?: CustomStyle;
```
ethanshar, line 53: What is new type? 
I don't we're using it in this file..

src/calendar/index.tsx
```diff
@@ -70,28 +127,21 @@ class Calendar extends Component {
...
+  header: RefObject<CalendarHeader> = React.createRef();
 
-  addMonth = count => {
-    this.updateMonth(this.state.currentMonth.clone().addMonths(count, true));
+  addMonth = (count: number) => {
+    this.updateMonth(this.state.currentMonth.clone().addMonths(count, true), false);
   };
 
-  updateMonth = (day, doNotTriggerListeners) => {
+  updateMonth = (day: any, doNotTriggerListeners: boolean) => {
```
ethanshar, line 144: You can set a default value here so you won't need to pass it on there places 
```suggestion
  updateMonth = (day: any, doNotTriggerListeners?: boolean = false) => {
```

src/calendar-list/index.tsx
Inbal-Tish: Just to match other props we're not manipulating before passing

src/calendar-list/item.tsx
Inbal-Tish: Done

src/calendar-list/index.tsx
Inbal-Tish: Done

src/calendar-list/index.tsx
Inbal-Tish: Done

src/calendar/day/marking/index.tsx
Inbal-Tish: On line #17 I declare the CustomStyle type

src/calendar/index.tsx
Inbal-Tish: Done

src/calendar/day/basic/index.tsx
Inbal-Tish: We're not supporting markings as an Array...

src/calendar-list/index.tsx
ethanshar: If anything I would move all `this.props.xxx` to the destructor so it's unified.. no?

ethanshar (approved):

src/calendar-list/index.tsx
Inbal-Tish: I'm hoping to extract FlatList props and pass them as a buldle and not one by one. This is how it's written all over the lib, I was just conforming, it's not a bit issue

PR #1587, opened 2021-07-29
Solution of TypeError: Cannot assign to read only property 'exports' of object '#<Object>' 

boray:
module.exports changed to export default


Inbal-Tish:
@boray See merged PR #1615.

PR #1593, opened 2021-08-08
Ts/ Timeline

lidord-wix:
Migrate Timeline to TS


Inbal-Tish:
@lidord-wix ExpandableCalendar migration to TS is completed. Please resolve conflicts and see if there are changes to make.

lidord-wix:
> @lidord-wix ExpandableCalendar migration to TS is completed. Please resolve conflicts and see if there are changes to make.

@Inbal-Tish  Done :)

src/timeline/Timeline.tsx
```diff
@@ -2,21 +2,43 @@
...
+  eventTapped?: (event: any) => void;
+  format24h?: boolean;
+  events: {
+    start: string;
+    end: string;
+    title: string;
+    summary: string;
+    color?: string;
+  }[];
+  styles?: any;
```
Inbal-Tish, line 31: Why 'style' is any?

src/timeline/Timeline.tsx
```diff
@@ -2,21 +2,43 @@
...
+  scrollToFirst?: boolean;
+  renderEvent?: (event: any) => JSX.Element;
+}
+
+interface State {
+  _scrollY: any;
```
Inbal-Tish, line 37: _scrollY should be a number

src/timeline/Timeline.tsx
```diff
@@ -39,11 +61,14 @@ export default class Timeline extends React.PureComponent {
     events: [],
     format24h: true
   };
+  calendarHeight: number;
+  private _scrollView: any;
+  style: any;
```
Inbal-Tish, line 66: style is not any type - please use StyleProp<ViewStyle>

src/timeline/Timeline.tsx
```diff
@@ -39,11 +61,14 @@ export default class Timeline extends React.PureComponent {
     events: [],
     format24h: true
   };
+  calendarHeight: number;
+  private _scrollView: any;
```
Inbal-Tish, line 65: _scrollView is a ref type not any. Also please use React.createRef and fix its usages.

src/timeline/Timeline.tsx
Inbal-Tish: Please use StyleProp<ViewStyle>

src/timeline/Timeline.tsx
```diff
@@ -2,21 +2,43 @@
...
 
 let {width: dimensionWidth} = Dimensions.get('window');
 
-export default class Timeline extends React.PureComponent {
+export interface TimelineProps {
+  start?: number;
+  end?: number;
+  eventTapped?: (event: any) => void;
+  format24h?: boolean;
+  events: {
```
Inbal-Tish, line 24: Why no declaring type Event so we can use it elsewhere like in the packer?

src/timeline/Packer.ts
```diff
@@ -3,11 +3,11 @@ import XDate from 'xdate';
 
 const offset = 100;
 
-function buildEvent(column, left, width, dayStart) {
-  const startTime = XDate(column.start);
-  const endTime = column.end ? XDate(column.end) : XDate(startTime).addHours(1);
+function buildEvent(column: any, left: number, width: number, dayStart: any) {
```
Inbal-Tish, line 6: We need to define an Event type and use it instead of any

src/types.ts
```diff
@@ -1,4 +1,4 @@
-import {ColorValue, ViewStyle, TextStyle} from 'react-native';
+import {ViewStyle, TextStyle} from 'react-native';
```
Inbal-Tish, line 1: It doesn't look like you're using these imports

src/types.ts
```diff
@@ -28,7 +28,7 @@ export interface Theme {
   timeLabel?: object;
   todayTextColor?: string;
   calendarBackground?: string;
-  indicatorColor?: ColorValue;
+  indicatorColor?: any;
```
Inbal-Tish, line 31: Why changed to any?

Inbal-Tish (approved):

Issue #962, opened 2019-10-17
AgendaList: not scrolling to the initially selected date at component mount


debugginator:
## Description

AgendaList component does not scroll to the initially selected date (date prop in the CalendarProvider) at component mount. Also reselecting that date in the calendar doesn't scroll, because the AgendaList doesn't detect a date change. Also if the selected day is today, there is no scroll to "Today" button, although the component does not display that date.

## Expected Behavior

On component mount the AgendaList should display the date selected by the date prop in the CalendarProvider component.

## Observed Behavior

AgendaList displays the first component from the sections prop array.

## Environment

Please run these commands in the project folder and fill in their results:

* `npm ls react-native-calendars`: react-native-calendars@1.212.0 
* `npm ls react-native`: react-native@0.59.8 

Also specify:

1. Phone/emulator/simulator & version:
XCode simulator, device: IPad Pro 9.7 inch

## Reproducible Demo
1. Clone wix repo. 
2. Change line 155 in "react-native-calendars/example/src/screens/expandableCalendar.js"
from `date={ITEMS[0].title}` to `date={ITEMS[3].title}`
3. Run the expandable calendar example

Tatator01:
have same problem

Aqhamza:
Hey, I'm facing the same issue. when I update the selected date (in my case it is Today) when my data updates than AgendaList doesn't scroll to that date. It always shows the first element in the data array.
Anyone who can help me in this issue, 
Thanks. :)

jbaubree:
Same problem here!

JulioOrellana:
I am facing the same issue here. 

Using react-native-calendars `v1.282.0`. Using Android.

The code bellow is executing:

```js
componentDidMount() {
    const { date } = this.props.context;
    if (date !== this._topSection) {
      setTimeout(() => {
        const sectionIndex = this.getSectionIndex(date); // --> finding section 7 (correct)
        this.scrollToSection(sectionIndex);
      }, 500);
    }
  }
```

```js
scrollToSection(sectionIndex) {
    if (this.list.current && sectionIndex !== undefined) {
      this.sectionScroll = true; // to avoid setDate() in onViewableItemsChanged
      this._topSection = this.props.sections[sectionIndex].title;

      this.list.current.scrollToLocation({
        animated: true,
        sectionIndex: sectionIndex,
        itemIndex: 0,
        viewPosition: 0, // position at the top
[...]

PR #1594, opened 2021-08-08
agendaList - remove onScrollToIndexFailed

souvik-ghosh:
Allow `onScrollToIndexFailed` to be passed via props.
We are facing the same issue as mentioned in https://github.com/wix/react-native-calendars/issues/962. This issue can be fixed by calling `scrollToIndex` (to the failed scroll index) inside the `onScrollToIndexFailed` method. Currently, there is no way to override this prop.

### DIFF
--- a/src/expandableCalendar/agendaList.js
+++ b/src/expandableCalendar/agendaList.js
@@ -163,6 +163,12 @@ class AgendaList extends Component {
     _.invoke(this.props, 'onMomentumScrollEnd', event);
   };
 
+  onScrollToIndexFailed = info => {
+    if (this.props.onScrollToIndexFailed) {
+      this.props.onScrollToIndexFailed(info);
+    }
+  };
+
   onHeaderLayout = ({nativeEvent}) => {
     this.sectionHeight = nativeEvent.layout.height;
   };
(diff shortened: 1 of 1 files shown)


src/expandableCalendar/agendaList.js
```diff
@@ -163,10 +163,6 @@ class AgendaList extends Component {
     _.invoke(this.props, 'onMomentumScrollEnd', event);
   };
 
-  onScrollToIndexFailed = info => {
-    console.warn('onScrollToIndexFailed info: ', info);
```
Inbal-Tish, line 167: Please make this a default if `onScrollToIndexFailed` is not passed by the user.

souvik-ghosh: @Inbal-Tish  Done.

Inbal-Tish: Where is the else statement? `else { console.warn('onScrollToIndexFailed info: ', info); }`...

souvik-ghosh: @Inbal-Tish Thought the library shouldn't warn. That's why removed it. Have added it again.

Inbal-Tish (approved):

jeffthompson1971:
Does anyone have a functioning Agenda/AgendaList  on a web build?  I really want to use this component on my project because it's the best-looking solution I've seen.  But without a functioning web build there's not much I can do.   Is anyone actively working on specifically the web support for this component(s)?  I'm seeing the same issue described here (I believe) which is nothing happens when I click on a Day in the calendar.   It's not scrolling to the correct day in the Agenda.  It's doing nothing at all.   Is this supposed to have been fixed with this patch described here or do we still need a fix to support the Web?

PR #1596, opened 2021-08-09
Feat/ support inactive state for day

lidord-wix:
Add support for inactive days (in addition to disabled days)

### DIFF
--- a/example/src/screens/calendars.tsx
+++ b/example/src/screens/calendars.tsx
@@ -3,7 +3,6 @@ import {StyleSheet, View, ScrollView, Text, TouchableOpacity, Switch} from 'reac
 // @ts-expect-error
 import {Calendar} from 'react-native-calendars';
 
-
 const testIDs = require('../testIDs');
 const INITIAL_DATE = '2020-02-02';
 
@@ -24,7 +23,6 @@ const CalendarsScreen = () => {
       <Fragment>
         <Text style={styles.text}>Calendar with selectable date</Text>
         <Calendar
-          disableAllTouchEventsForInactiveDays
           testID={testIDs.calendars.FIRST}
           current={INITIAL_DATE}
           style={styles.calendar}
@@ -35,12 +33,6 @@ const CalendarsScreen = () => {
               disableTouchEvent: true,
               selectedColor: 'orange',
               selectedTextColor: 'red'
-            },
-            '2020-02-10': {
-              inactive: true
-            },
-            '2020-02-11': {
-              inactive: true
             }
           }}
         />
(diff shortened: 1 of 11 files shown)


src/calendar/day/marking/index.tsx
```diff
@@ -41,6 +41,7 @@ export interface MarkingProps extends DotProps {
   marked?: boolean;
   today?: boolean;
   disabled?: boolean;
+  inactive?: boolean;
```
Inbal-Tish, line 44: Where are you passing 'inactive' to the Marking component? I don't see it in BasicDay...

example/src/screens/calendars.tsx
```diff
@@ -24,6 +24,7 @@ const CalendarsScreen = () => {
       <Fragment>
         <Text style={styles.text}>Calendar with selectable date</Text>
         <Calendar
+          disableAllTouchEventsForInactiveDays
```
Inbal-Tish, line 27: Can you please have a separate example for the inactive state?

src/calendar/day/marking/index.tsx
lidord-wix: The type of BasicDay's `marking` is `MarkingProps` from the `Marking` component, so I was had to add it for using `this.marking.inactive` in BasicDay

example/src/screens/calendars.tsx
lidord-wix: sure

src/calendar/day/marking/index.tsx
Inbal-Tish: That's great, but you're not passing this prop in BasicDay component (here: https://github.com/wix/react-native-calendars/blob/master/src/calendar/day/basic/index.tsx#L170)

Inbal-Tish: You need to pass to the marking in order to style the marking according to the inactive state as you do in the Day component, no?

lidord-wix: I'm not sure, there's another inactive marking besides the inactive text color?

Inbal-Tish: There is no inactive marking, that's what I'm talking about. What will be the case of an inactive day with marking? How would it look?

lidord-wix: I see.. Added a support for that :)

Inbal-Tish: First, there are other marking beside Dot. Second, how will the inactive state get to the Marking and the Dot components if you're not passing them from BasicDay to the Marking?

lidord-wix: fixed

src/calendar/day/basic/index.tsx
```diff
@@ -174,6 +187,7 @@ export default class BasicDay extends Component<BasicDayProps> {
         marked={this.isMultiDot() ? true : marked}
         selected={this.isSelected()}
         disabled={this.isDisabled()}
+        inactive={this.isInactive()}
```
Inbal-Tish, line 190: Yes! That's what I meant

src/types.ts
```diff
@@ -12,6 +12,8 @@ export type DateData = {
   timestamp: number,
   dateString: string
 }
+export type DayState = 'selected' | 'disabled' | 'inactive' | 'today' | '';
```
Inbal-Tish, line 15: Love it! I was about to do it myself...

Inbal-Tish: Why the empty string?

PR #1605, opened 2021-08-22
migrate demo files to ts

Inbal-Tish:


ethanshar (approved):

ethanshar:
@Inbal-Tish 
I pushed two commits to this PR 

The [first commit](https://github.com/wix/react-native-calendars/pull/1605/commits/a7f8597d3e34e9e7333d8da02f45ff44891bcbc4) fixes issue with imports/exports of example screens' testIDs. If you'd run the demo app you would have gotten errors. 

The [second commit](https://github.com/wix/react-native-calendars/pull/1605/commits/04fd421755d6870c3c37907c8fc3a5768cdea334) fixes the issue you had with lint, but now that lint works there are lots of lint errors coming from the example screens' code - so we still need to fix those

Inbal-Tish:
> @Inbal-Tish
> I pushed two commits to this PR
> 
> The [first commit](https://github.com/wix/react-native-calendars/pull/1605/commits/a7f8597d3e34e9e7333d8da02f45ff44891bcbc4) fixes issue with imports/exports of example screens' testIDs. If you'd run the demo app you would have gotten errors.
> 
> The [second commit](https://github.com/wix/react-native-calendars/pull/1605/commits/04fd421755d6870c3c37907c8fc3a5768cdea334) fixes the issue you had with lint, but now that lint works there are lots of lint errors coming from the example screens' code - so we still need to fix those

Thanks! I'll work on the lint errors

Inbal-Tish:
@ethanshar I fixed the ts errors but there is still an issue with the example/testIDs import...

ethanshar:
> @ethanshar I fixed the ts errors but there is still an issue with the example/testIDs import...

What's the issue? 
When are you getting it?

Inbal-Tish:
> > @ethanshar I fixed the ts errors but there is still an issue with the example/testIDs import...
> 
> What's the issue?
> When are you getting it?

Error: Cannot find module '../example/src/testIDs'. See console output

PR #1612, opened 2021-08-24
now using dynamic keys for theme objects

klazbaba:
This PR fixes themes not working.
Adding for example stylesheet.calendar.main to theme is not picked up by the styles defined in the component because the object key contains a ".", which is not valid in a variable name.


src/calendar/day/marking/style.ts
```diff
@@ -26,6 +26,7 @@ export default function styleConstructor(theme: Theme = {}) {
       borderBottomRightRadius: 2,
       marginRight: 4
     },
-    ...(theme.stylesheet?.marking || {})
+    //@ts-ignore
+    ...(theme['stylesheet.marking '] || {})
```
Klemminn, line 30: Probably shouldn't be a space at the end of 'stylesheet.marking '?

klazbaba: True, thanks for pointing that out.

Klemminn (approved):

hraschan:
Can someone approve those changes? Would be really helpful! :)

hraschan (approved):
Everything looks good to me.

theneekz:
This bug applies to all of the theme ids, not just markings:
"stylesheet.calendar.header"
"stylesheet.calendar.main"
"stylesheet.calendar-list.main"
"stylesheet.agenda.main"
"stylesheet.agenda.list"
"stylesheet.day.basic"
"stylesheet.day.single"
"stylesheet.day.multiDot"
"stylesheet.day.period"
"stylesheet.dot"

Klemminn:
@Inbal-Tish could you please take a look? :)
This feature is working in older releases, would be wonderful to get it back.

Klemminn:
@ethanshar @emilisb this seems like a minor but needed change.... can either of you take a look? :)

Inbal-Tish (approved):

theneekz:
:) 👍

PR #1613, opened 2021-08-25
change lodash import way to make more tree-shakeable

xsbchen:
webpack can not remove unuse methods in lodash, because we use `require('lodash')` to import it
so change to `require('lodash/methodX')`

### DIFF
--- a/example/src/screens/expandableCalendar.tsx
+++ b/example/src/screens/expandableCalendar.tsx
@@ -1,4 +1,4 @@
-import _ from 'lodash';
+import isEmpty from 'lodash/isEmpty';
 import React, {Component, useCallback} from 'react';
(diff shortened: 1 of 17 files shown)


xsbchen:
@ethanshar @Inbal-Tish Could you help me to review this PR? Please.

ethanshar (approved):

ethanshar:
@Inbal-Tish 
Looks good to me, I approved. 
Leaving you to merge

Inbal-Tish:
@xsbchen One question: why can't we extract the methods this way `import {invoke, isFunction} from 'lodash';` instead of having two lines of imports in this case?

xsbchen:
> @xsbchen One question: why can't we extract the methods this way `import {invoke, isFunction} from 'lodash';` instead of having two lines of imports in this case?

No, lodash did not support this way for tree-shaking

PR #1615, opened 2021-08-29
Use modern export keyword

umrashrf:


umrashrf:
Related https://github.com/wix/react-native-calendars/pull/1587

Inbal-Tish (approved):

PR #1622, opened 2021-09-02
Toggle week scroll

fraserbenjamin:
This adds the scrollBetweenWeeks prop on the reservations component which gives the ability to disable the agenda moving automatically between weeks. Defaults to true to be enabled.

### DIFF
--- a/src/agenda/reservation-list/index.tsx
+++ b/src/agenda/reservation-list/index.tsx
@@ -53,6 +53,8 @@ export type ReservationListProps = ReservationProps & {
   refreshing?: boolean;
   /** If provided, a standard RefreshControl will be added for "Pull to Refresh" functionality. Make sure to also set the refreshing prop correctly */
   onRefresh?: () => void;
+  /** Disables scrolling between weeks in the main agenda view. Default = false */
+  scrollBetweenWeeks?: boolean;
 };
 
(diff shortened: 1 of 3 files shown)


fraserbenjamin:
scrollBetweenWeeks

src/agenda/reservation-list/index.tsx
```diff
@@ -104,6 +108,7 @@ class ReservationList extends Component<ReservationListProps, ReservationsListSt
   private selectedDay: XDate;
   private scrollOver: boolean;
   private list: React.RefObject<FlatList> = React.createRef();
+  private scrollBetweenWeeks: boolean;
```
Inbal-Tish, line 111: Why do you need this const? Can't you use the prop value directly?

src/agenda/reservation-list/index.tsx
```diff
@@ -118,6 +123,7 @@ class ReservationList extends Component<ReservationListProps, ReservationsListSt
     this.heights = [];
     this.selectedDay = props.selectedDay;
     this.scrollOver = true;
+    this.scrollBetweenWeeks = props.scrollBetweenWeeks === false ? false : true;
```
Inbal-Tish, line 126: This seems redundent

PR #1627, opened 2021-09-13
ExpandableCalendar's today button bugs

lidord-wix:
Fix two bugs with ExpandableCalendar's today button:
1. Pressing on today button takes to the wrong date - production bug - WOAUILIB-2106
2. Pressing on today button while the visible month is the same month in a different year (09/2022 for example) takes to the right date but to the wrong month (the visible one).

### DIFF
--- a/src/expandableCalendar/index.tsx
+++ b/src/expandableCalendar/index.tsx
@@ -135,7 +135,8 @@ class ExpandableCalendar extends Component<Props, State> {
   _weekCalendarStyles: {
     style: ViewStyle;
   };
-  visibleMonth: any;
+  visibleMonth: number;
+  visibleYear: number;
   initialDate: XDate;
   headerStyleOverride: Theme;
   header: React.RefObject<any> = React.createRef();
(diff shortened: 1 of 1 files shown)


src/expandableCalendar/index.tsx
```diff
@@ -427,8 +429,10 @@ class ExpandableCalendar extends Component<Props, State> {
   };
 
   onVisibleMonthsChange = (value: DateData[]) => {
-    if (this.visibleMonth !== first(value)?.month) {
-      this.visibleMonth = first(value)?.month; // equivalent to this.getMonth(value[0].dateString)
+    const month = first(value)?.month;
+    if (month && this.visibleMonth !== month) {
+      this.visibleMonth = month; // equivalent to this.getMonth(value[0].dateString)
```
Inbal-Tish, line 434: the comment should go to line 432

src/expandableCalendar/index.tsx
```diff
@@ -427,8 +429,10 @@ class ExpandableCalendar extends Component<Props, State> {
...
+      this.visibleYear = first(value)?.year || this.getYear(new Date());
```
Inbal-Tish, line 435: this.visibleYear is already set on line 161. So why do we need to default to this.getYear(new Date())? Can't we just not set the value if it's undefined?

lidord-wix: Done :)

Inbal-Tish (approved):

PR #1636, opened 2021-09-26
Fix/ AgendaList dayPress

lidord-wix:
Update the agenda position after pressing on any date (and not only while pressing date with a content in the agenda)
WOAUILIB-2119

### DIFF
--- a/example/src/screens/expandableCalendar.tsx
+++ b/example/src/screens/expandableCalendar.tsx
@@ -190,7 +190,7 @@ export default class ExpandableCalendarScreen extends Component<Props> {
         <AgendaList
           sections={ITEMS}
           renderItem={this.renderItem}
-          // scrollToFutureEvents
+          // scrollToNextEvent
           // sectionStyle={styles.section}
         />
       </CalendarProvider>
--- a/example/src/screens/expandableCalendar.tsx
+++ b/example/src/screens/expandableCalendar.tsx
@@ -190,6 +190,7 @@ export default class ExpandableCalendarScreen extends Component<Props> {
         <AgendaList
           sections={ITEMS}
           renderItem={this.renderItem}
+          // scrollToFutureEvents
           // sectionStyle={styles.section}
         />
       </CalendarProvider>
--- a/src/expandableCalendar/agendaList.tsx
+++ b/src/expandableCalendar/agendaList.tsx
@@ -6,7 +6,18 @@ import PropTypes from 'prop-types';
 import XDate from 'xdate';
 
 import React, {Component} from 'react';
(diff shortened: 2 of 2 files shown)


example/src/screens/expandableCalendar.tsx
```diff
@@ -190,6 +190,7 @@ export default class ExpandableCalendarScreen extends Component<Props> {
         <AgendaList
           sections={ITEMS}
           renderItem={this.renderItem}
+          // scrollToFutureEvents
```
Inbal-Tish, line 193: Naming is a bit confusing. How about something like 'scrollToNextMarkedDate' or 'scrollToNextEvent'?
Also, You need to adjust the example to show the feature. I don't see it working in the example...

src/expandableCalendar/agendaList.tsx
```diff
@@ -37,6 +50,8 @@ interface Props extends SectionListProps<any, DefaultSectionT> {
   avoidDateUpdates?: boolean;
   /** offset scroll to section */
   viewOffset?: number;
+  /** enable scrolling the agenda list to the next date with events while pressing a date without events */
```
Inbal-Tish, line 53: Will skip empty dates? Or scroll to them if they are the next date with content?

src/expandableCalendar/agendaList.tsx
```diff
@@ -86,10 +101,11 @@ class AgendaList extends Component<Props> {
 
   componentDidMount() {
     const {date} = this.props.context;
+    const {scrollToFutureEvents} = this.props;
     if (date !== this._topSection) {
       setTimeout(() => {
-        const sectionIndex = this.getSectionIndex(date);
-        if (sectionIndex) {
+        const sectionIndex = scrollToFutureEvents ? this.getNextSectionIndex(date) : this.getSectionIndex(date);
```
Inbal-Tish, line 107: We should move these few lines to a method as it repeats on didUpdate

example/src/screens/expandableCalendar.tsx
lidord-wix: I updated the prop name.
The example works for me, try to uncomment this line and press on the next date in the weekCalendar

src/expandableCalendar/agendaList.tsx
lidord-wix: will scroll to the next date with content.. I changed the description :)

src/expandableCalendar/agendaList.tsx
lidord-wix: I moved it into `scrollToSection`

example/src/screens/expandableCalendar.tsx
Inbal-Tish: I did... But when I press the 3rd of the month it doesn't scroll to the 4th, where there is content...

Inbal-Tish: Sorry. It works now. Probably cache...

Inbal-Tish (approved):

PR #1643, opened 2021-10-02
RenderItem props not delivered in weekCalendar

seunghoking:
Initially, When switching from a monthly calendar to a weekly calendar, correct issues that cannot be re-rended because marked dates are not properly delivered.

### ScreenShot
- monthly calendar

- The weekly calendar of the existing code

- The weekly calendar of the New code


Inbal-Tish:
@seunghoking Hi. Thank you for the PR. Just to clarify, how do you switch between monthly and weekly calendars? Can you add a code sample please?

PR #1644, opened 2021-10-04
Finalize ts

ethanshar:
This PR finalize the TS migration and expose its typings to external users


src/index.ts
```diff
@@ -0,0 +1,11 @@
+export { default as Calendar } from './calendar';
```
Inbal-Tish, line 1: We need to export all the props as well

ethanshar: Fixed. 
And I trigger a new snapshot build for this branch

Inbal-Tish (approved):

PR #1648, opened 2021-10-05
Mention that newer versions are in Github Releases

swrobel:


swrobel:
Nevermind, CHANGELOG is just in ascending order 🤦🏼

PR #1651, opened 2021-10-06
Fix import path of calendar provider

bluedaniel:
This small change fixes a broken import.

Bug was found whilst mimicking this example

https://github.com/wix/react-native-calendars/blob/master/example/src/screens/expandableCalendar.tsx

Simulator Screen Shot - iPhone 13 mini - 2021-10-06 at 08 47 49


Inbal-Tish:
@bluedaniel Hi. Thank you for the PR. I can't reproduce the issue. Did you manage to reproduce it in our demo app?

CelfocusJawwy (approved):
This was needed! Require cycle is happening. 

Thanks!

Inbal-Tish (approved):

Issue #1653, opened 2021-10-08
Extending the knob to fill the entire row for agenda


OwenTsoi:
## Description

I'd like to know if there's a way to extend the knob to fill the entire row in the agenda since right now it is only the middle part that is functional.

## Expected Behavior

I rendered my own knob to fill the entire row, and I would expect the calendar would be extended when I drag down from anywhere in the knob row.

## Observed Behavior

Only the middle part is functional regardless of the length of my knob.

## Environment
<img width="433" alt="Screen Shot 2021-10-08 at 4 37 53 pm" src="https://user-images.githubusercontent.com/54052624/136503923-b8855e85-f003-479b-a356-80f088497135.png">

<img width="313" alt="Screen Shot 2021-10-08 at 4 38 05 pm" src="https://user-images.githubusercontent.com/54052624/136503943-dae2db62-7964-4e89-8e25-1cd840e636cc.png">

1. Device/emulator/simulator & OS version:

  Model Name:	MacBook Pro
  Model Identifier:	MacBookPro16,4
  Processor Name:	6-Core Intel Core i7
  Processor Speed:	2.6 GHz
  Number of Processors:	1
  Total Number of Cores:	6
  L2 Cache (per Core):	256 KB
  L3 Cache:	12 MB
  Hyper-Threading Technology:	Enabled
  Memory:	32 GB
  System Firmware Version:	1554.140.20.0.0 (iBridge: 18.16.14759.0.1,0)

Simulator: iPhone 12 mini

## Screenshots
![image](https://user-images.githubusercontent.com/54052624/136504227-cab8fd66-8102-4101-b936-ca2f2deb2685.png)

BobbyCarelse:
I need to see your style properties too

kei95:
Hi @OwenTsoi I think I can work on this issue. What's happening here is that the `Agenda` component doesn't have props that allow you to modify the knob's style which is coming from `Animated.ScrollView` as you can see in the screenshot. So you can currently change the appearance of knob but the knob's style itself as of now.

I'll open a PR soon to add a props that allows you to modify its style so that you can expand its touchable zone.

![Screen Shot 2021-10-12 at 10 43 48 PM](https://user-images.githubusercontent.com/44686790/137074366-5a7c6cfc-f99b-443a-ac0e-76fef60f290a.png)
e

OwenTsoi:
[...]

PR #1657, opened 2021-10-14
Add style props to Agenda to change knob's style

kei95:
### Description
This PR addresses: https://github.com/wix/react-native-calendars/issues/1653
I added a new props `knobStyle` to let make the knob's style changeable, but if the maintainer thinks just making the knob's width wider is enough, I can also open the PR for it and close this one.

### Changes
 - Add `knobStyle` props to let users to change knob's style
 - Update Agenda's example

### DIFF
--- a/example/src/screens/agenda.tsx
+++ b/example/src/screens/agenda.tsx
@@ -20,6 +20,14 @@ export default class AgendaScreen extends Component {
         renderEmptyDate={this.renderEmptyDate.bind(this)}
         rowHasChanged={this.rowHasChanged.bind(this)}
         showClosingKnob={true}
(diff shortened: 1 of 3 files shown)


Inbal-Tish:
@kei95 Hi. If all you want to achieve is a full-width knob, why not add a prop that will adjust the 'scrollPadStyle' (line 467) accordingly? I don't see why we need a full style object for that...

kei95:
@Inbal-Tish Hi! I just opened another PR for the style change to just accomplish the full-width behavior. Thus I'll close this PR. Please take a look when you get a chance. Thanks in advance!

https://github.com/wix/react-native-calendars/pull/1730

PR #1673, opened 2021-11-04
Add closeOnDayPress prop to ExpandableCalendar

cmorbitzer:
Currently, if `ExpandableCalendar` is open, it will always close when a day is pressed, and there is no way to override this behavior. This PR adds a new prop `closeOnDayPress` to `ExpandableCalendar`, which when `false` will not close an open calendar when a day is pressed. This property defaults to `true` to maintain backward compatibility.


cmorbitzer:
@Inbal-Tish Are there plans to review this?

Inbal-Tish (approved):

PR #1674, opened 2021-11-05
Fix contentRef in CalendarConsumer

### LINKED ISSUE #1634: How to open/close ExpandableCalendar of react-native-calendars programmatically?
I want to open/close ExpandableCalendar programmatically. I found there's only an ```initialPosition``` prop. It's not working when changing ```initialPosition```. So I tried to use ref and call ```setPosition()``` function of ExpandableCalendar

```typescript
export const MyScreen = () => {

    const ref = useRef<ExpandableCalendar>()

    return (
        <CalendarProvider date={new Date()}>
            <ExpandableCalendar ref={ref} />
        </CalendarProvider>
    )
}
```

when I use this code, the ```ref.current.contentRef``` is always ```undefined```

## My Question is
* How to get the ```ref``` of ExpandableCalendar? 
   It seems the ```asCalendarConsumer``` saves the ref as contentRef:
  ```typescript
  function asCalendarConsumer(WrappedComponent: React.ComponentType<any>): React.ComponentClass {
  
    class CalendarConsumer extends Component {
      contentRef: any;
  
      saveRef = (r: Ref<React.Component<any>>) => {
        this.contentRef = r;
      };
  
      render() {
        return (
          <CalendarContext.Consumer>
            {context => <WrappedComponent ref={this.contentRef} context={context} {...this.props} />}
          </CalendarContext.Consumer>
        );
      }
    }
  
    hoistNonReactStatic(CalendarConsumer, WrappedComponent);
  
    return CalendarConsumer;
  }
  
  export default asCalendarConsumer;
  ```
  ```typescript
  export default asCalendarConsumer(ExpandableCalendar);
  ```

* Is there a way to open ExpandableCalendar programmatically? Does the setPosition() function work or there's another way to do this. I found an issue #1362 is in progress.

## My Environment is

* `react-native-calendars@1.1266.0`:
* `react-native@0.64.2`:

cmorbitzer:
The conversion of `ExpandableCalendar` to TypeScript in #1598 introduced a breaking change to the behavior of `contentRef` in `CalendarConsumer`. Before, `contentRef` held a reference to the wrapped component. Now, `contentRef` is always `undefined`. This PR returns to the previous behavior.

Fixes #1634

### DIFF
--- a/src/expandableCalendar/asCalendarConsumer.tsx
+++ b/src/expandableCalendar/asCalendarConsumer.tsx
@@ -16,7 +16,7 @@ function asCalendarConsumer(WrappedComponent: React.ComponentType<any>): React.C
     render() {
       return (
         <CalendarContext.Consumer>
-          {context => <WrappedComponent ref={this.contentRef} context={context} {...this.props} />}
+          {context => <WrappedComponent ref={this.saveRef} context={context} {...this.props} />}
         </CalendarContext.Consumer>
       );
     }


cmorbitzer:
@Inbal-Tish Are there plans to review this?

Inbal-Tish (approved):

PR #1681, opened 2021-11-15
Using React Fragment to avoid style issue

jungleyu:
Here's one case for me:
I add styles for container view, but with a View for GestureRecognizer, my style will mess up, if we use Fragment, there will be no side effect on container.

### DIFF
--- a/src/calendar/index.tsx
+++ b/src/calendar/index.tsx
@@ -309,7 +309,7 @@ class Calendar extends Component<CalendarProps, CalendarState> {
 
   render() {
     const {enableSwipeMonths, style} = this.props;
-    const GestureComponent = enableSwipeMonths ? GestureRecognizer : View;
+    const GestureComponent = enableSwipeMonths ? GestureRecognizer : React.Fragment;
     const gestureProps = enableSwipeMonths ? this.swipeProps : undefined;
 
     return (


Inbal-Tish:
@jungleyu Hi. Please add a description to explain the issue with using a View vs. Fregment here?

jungleyu:
> @jungleyu Hi. Please add a description to explain the issue with using a View vs. Fregment here?

sure

jungleyu:
@Inbal-Tish While I'm trying to set the width of the calendar to "100%", it's not working.

jungleyu:
image!


Expected:
[image](https://user-images.githubusercontent.com/6914315/157392901-700c4a06-2c12-4c0d-baaf-ea8619fe90bf.jpeg)

Inbal-Tish:
@jungleyu How do you pass the width to the calendar? Are you not using the 'style' prop? Can you please add a code snippet?

jungleyu:
@Inbal-Tish I'm using style={width: "100%"}, but worked with a fixed number, I thought that was caused by the View wrapper as I mentioned.
<GestureComponent> <-- This View layer do have side effect 
   <View style={props.style}></View>
</GestureComponent>

PR #1687, opened 2021-11-18
feature: calendarList, numberOfColumn prop

RajRohitYadav:
Have added a prop(numberOfColumn) in calendarList to control number of column, which will be 1 by default.

`  <CalendarList
        numberOfColumn={3}
      />
`

Below are the images in Landscape and Portrait

### DIFF
--- a/src/calendar-list/index.tsx
+++ b/src/calendar-list/index.tsx
@@ -57,6 +57,8 @@ interface Props extends CalendarProps, Omit<FlatListProps<any>, 'data' | 'render
   /** onLayout event */
   onLayout?: (event: LayoutChangeEvent) => void;
   removeClippedSubviews?: boolean;
+  /** Control number of columns */
+  numberOfColumn: number;
 }
 export type CalendarListProps = Props;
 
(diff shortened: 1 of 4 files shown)


src/calendar-list/index.tsx
```diff
@@ -343,12 +348,14 @@ class CalendarList extends Component<Props, State> {
   }
 
   render() {
-    const {style, pastScrollRange, futureScrollRange, horizontal, showScrollIndicator} = this.props;
+    const {style, pastScrollRange, futureScrollRange, horizontal, showScrollIndicator, numberOfColumn} = this.props;
 
     return (
       <View style={this.style.flatListContainer}>
         <FlatList
+          key={numberOfColumn}
```
Inbal-Tish, line 356: Why do you need to pass 'key' here? Do you need to reconstruct the FlatList when the 'numberOfColums' change?

src/calendar-list/item.tsx
```diff
@@ -110,17 +110,19 @@ class CalendarListItem extends Component<CalendarListItemProps, CalendarListItem
...
-          testID={testID}
-          current={item}
-          style={calStyle}
-          headerStyle={horizontal ? headerStyle : undefined}
-          disableMonthChange
-          onPressArrowLeft={horizontal ? this.onPressArrowLeft : onPressArrowLeft}
-          onPressArrowRight={horizontal ? this.onPressArrowRight : onPressArrowRight}
-          context={context}
-        />
+        <View style={this.style.calendarContainer}>
```
Inbal-Tish, line 113: Let's just wrap with an extra View when numberOfColumns > 1

PR #1688, opened 2021-11-19
Fix overriding hideExtraDays on ExpandableCalendar

cmorbitzer:
Currently, `hideExtraDays` on `ExpandableCalendar` does not change the display of extra days on the calendar, even though `hideExtraDays` is a property of `ExpandableCalendar` inherited from `Calendar`. `hideExtraDays` is always set to `false` for horizontal scroll and `true` for vertical scroll. This fixes `ExpandableCalendar` to respect `hideExtraDays` if it is explicitly set, and defaults to the currently behavior if it is not.

### DIFF
--- a/src/expandableCalendar/index.tsx
+++ b/src/expandableCalendar/index.tsx
@@ -570,6 +570,7 @@ class ExpandableCalendar extends Component<Props, State> {
             <CalendarList
               testID="calendar"
               horizontal={horizontal}
+              hideExtraDays={!horizontal}
               {...others}
               theme={themeObject}
               ref={this.calendar}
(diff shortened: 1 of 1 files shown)


Inbal-Tish:
@cmorbitzer Hi. There's a ux reason for this logic. Can you share a design that makes sense otherwise?

cmorbitzer:
The design I am implementing is to have extra days hidden in month view and shown in week view on a horizontal scroll calendar. So in my case, I'm actually passing a dynamic value to `hideExtraDays` and not a primitive boolean. This change allows me to implement that and does work for this purpose. I've attached screenshots below. I am open to considering a better way to implement this.

Screen Shot 2021-12-07 at 10 13 01 AM
Screen Shot 2021-12-07 at 10 13 24 AM

Inbal-Tish:
@cmorbitzer I see what you mean. When ExpandableCalendar is with 'horizontal={false}' you want to have the extra days shown on the week view. Right? If so I think it's a good idea, and it should be fixed in the component and not something you need to control as a user of the component.

cmorbitzer:
Agreed. Closing in favor of #1705

PR #1694, opened 2021-12-01
fix: day accessibility label localization

pbr1111:
Lodash get can't resolve XDate.locales[XDate.defaultLocale] because it can't retrieve the value of XDate.defaultLocale since it's used as a string and not as a value in the path parameter of the _.get method. .

This PR implements a fix to get the values natively with JS instead of using lodash. 

It also fixes the case that when the prop `formatAccessibilityLabel` is not defined, the hardcoded string 'today' is used instead of using the XDate translated value.

### DIFF
--- a/src/calendar/day/index.tsx
+++ b/src/calendar/day/index.tsx
@@ -77,15 +77,13 @@ export default class Day extends Component<DayProps> {
   }
 
   getAccessibilityLabel = memoize((day, marking, isToday) => {
+    // @ts-expect-error
     const today = XDate.locales[XDate.defaultLocale].today || 'today';
-    const formatAccessibilityLabel = XDate.locales[XDate.defaultLocale].formatAccessibilityLabel;
+    // @ts-expect-error
+    const formatAccessibilityLabel = XDate.locales[XDate.defaultLocale].formatAccessibilityLabel || 'dddd d MMMM yyyy';
     const markingLabel = this.getMarkingLabel(marking);
 
-    if (formatAccessibilityLabel) {
-      return `${isToday ? today : ''} ${day.toString(formatAccessibilityLabel)} ${markingLabel}`;
-    }
-
-    return `${isToday ? today : ''} ${day.toString('dddd d MMMM yyyy')} ${markingLabel}`;
+    return `${isToday ? today : ''} ${day.toString(formatAccessibilityLabel)} ${markingLabel}`;
   });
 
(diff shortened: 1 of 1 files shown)


src/calendar/day/index.tsx
```diff
@@ -78,15 +77,15 @@ export default class Day extends Component<DayProps> {
...
+    const today = XDate.locales[XDate.defaultLocale].today || 'today';
+    const formatAccessibilityLabel = XDate.locales[XDate.defaultLocale].formatAccessibilityLabel;
     const markingLabel = this.getMarkingLabel(marking);
 
     if (formatAccessibilityLabel) {
       return `${isToday ? today : ''} ${day.toString(formatAccessibilityLabel)} ${markingLabel}`;
     }
 
-    return `${isToday ? 'today' : ''} ${day.toString('dddd d MMMM yyyy')} ${markingLabel}`;
+    return `${isToday ? today : ''} ${day.toString('dddd d MMMM yyyy')} ${markingLabel}`;
```
Inbal-Tish, line 88: You can loos the condition (lines 84-86) if you pass `const format = formatAccessibilityLabel || 'dddd d MMMM yyyy';` to the day.toString(format) on line 88;

src/calendar/day/index.tsx
```diff
@@ -78,15 +77,15 @@ export default class Day extends Component<DayProps> {
   }
 
   getAccessibilityLabel = memoize((day, marking, isToday) => {
-    const today = get(XDate, 'locales[XDate.defaultLocale].today');
-    const formatAccessibilityLabel = get(XDate, 'locales[XDate.defaultLocale].formatAccessibilityLabel');
+    const today = XDate.locales[XDate.defaultLocale].today || 'today';
+    const formatAccessibilityLabel = XDate.locales[XDate.defaultLocale].formatAccessibilityLabel;
```
Inbal-Tish, line 81: You'll need to resolve ts errors as both 'today' and 'formatAccessibilityLabel'  does not exist on type 'locale_detail'

src/calendar/day/index.tsx
pbr1111: Right, I hadn't realised, thanks 😄

src/calendar/day/index.tsx
pbr1111: Fixed!

Inbal-Tish (approved):

PR #1695, opened 2021-12-01
Calendar - add 'initialDate' prop

Inbal-Tish:
Calendar - add 'initialDate' prop that initializes the component when changed (will replace 'current' after migration)


src/calendar/index.tsx
```diff
@@ -91,6 +93,8 @@ class Calendar extends Component<CalendarProps, CalendarState> {
     style: PropTypes.oneOfType([PropTypes.object, PropTypes.array, PropTypes.number]),
     /** Initially visible month. Default = Date() */
     current: PropTypes.any,
+    /** Initially visible month. If changed will initialize the calendar to this value */
+    initialDate: PropTypes.any,
```
ethanshar, line 97: I think we can kill 2 birds here. 
What if `initialDate` will be a timestamp (number) 
As we discussed before, as part of the big refactor, we'll want to use timestamp for our value source.

src/calendar/index.tsx
```diff
@@ -132,12 +136,24 @@ class Calendar extends Component<CalendarProps, CalendarState> {
     enableSwipeMonths: false
   };
 
+  currentDate = this.props.current || this.props.initialDate;
+
   state = {
-    currentMonth: this.props.current ? parseDate(this.props.current) : new XDate()
+    currentMonth:
```
ethanshar, line 142: I know the code was already like that, but why are we setting the current date into something called `currentMonth`

src/calendar/index.tsx
```diff
@@ -280,7 +296,7 @@ class Calendar extends Component<CalendarProps, CalendarState> {
 
   renderHeader() {
     const {customHeader, headerStyle, displayLoadingIndicator, markedDates, testID} = this.props;
-    const current = parseDate(this.props.current);
+    const current = parseDate(this.currentDate);
```
ethanshar, line 299: Why are you talking the current date from the `this` and not from the state? 
In general, why are we keeping the current date also on `this`?

src/calendar/index.tsx
Inbal-Tish: The calendar only handles the month/year so that's is focus

src/calendar/index.tsx
Inbal-Tish: Changed

src/calendar/index.tsx
Inbal-Tish: Do you mean to change 'initialDate' type to a timestamp in preparation for the timestamp support? I will simply convert to XDate at this point...

Inbal-Tish: Did we decide if to support timestamp as number or string?

ethanshar: Keep the string we have today
and add a ticket to create a util for formatting the value string format

Inbal-Tish:
@ethanshar I merged master and moved 'initialDate' to be string.

ethanshar (approved):

PR #1697, opened 2021-12-05
Add numberOfWeeks prop to calendar

kkacquah:
Allows for truncated calendar that only shows a specific number of weeks.

Simulator Screen Shot - iPhone 12 - 2021-12-05 at 13 48 18

### DIFF
--- a/src/calendar/index.tsx
+++ b/src/calendar/index.tsx
@@ -69,8 +69,8 @@ export interface CalendarProps extends CalendarHeaderProps, DayProps {
   customHeader?: any;
   /** Allow selection of dates before minDate or after maxDate */
   allowSelectionOutOfRange?: boolean;
-  /** Max number of weeks to show on each month of month view */
-  numberOfWeeks?: number;
+  /** Max number of weeks to show from the current date */
+  numberOfWeeksFromCurrent?: number;
 }
 
 interface CalendarState {
@@ -130,7 +130,7 @@ class Calendar extends Component<CalendarProps, CalendarState> {
     /** Allow selection of dates before minDate or after maxDate */
     allowSelectionOutOfRange: PropTypes.bool,
     /** Allow selection of dates before minDate or after maxDate */
-    numberOfWeeks: PropTypes.number
+    numberOfWeeksFromCurrent: PropTypes.number
   };
   static defaultProps = {
     enableSwipeMonths: false
--- a/src/calendar/index.tsx
+++ b/src/calendar/index.tsx
@@ -268,17 +268,14 @@ class Calendar extends Component<CalendarProps, CalendarState> {
     );
   }
 
-  belowMaxNumberOfWeeks(weekLength: number) {
-    return (this.props.numberOfWeeks ? weekLength <= (this.props.numberOfWeeks - 1): true) 
-  }
   renderMonth() {
     const {currentMonth} = this.state;
-    const {firstDay, showSixWeeks, hideExtraDays} = this.props;
+    const {firstDay, showSixWeeks, hideExtraDays, numberOfWeeks} = this.props;
     const shouldShowSixWeeks = showSixWeeks && !hideExtraDays;
-    const days = page(currentMonth, firstDay, shouldShowSixWeeks);
+    const days = page(currentMonth, firstDay, shouldShowSixWeeks, numberOfWeeks);
     const weeks = [];
 
-    while (days.length && this.belowMaxNumberOfWeeks(weeks.length)) {
+    while (days.length) {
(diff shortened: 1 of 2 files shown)


src/dateutils.ts
```diff
@@ -67,8 +76,8 @@ export function weekDayNames(firstDayOfWeek = 0) {
   return weekDaysNames;
 }
 
-export function page(date: XDate, firstDayOfWeek = 0, showSixWeeks = false) {
-  const days = month(date);
+export function page(date: XDate, firstDayOfWeek = 0, showSixWeeks = false, numberOfWeeks: number | undefined = undefined) {
```
Inbal-Tish, line 79: Why not use `numberOfWeeks?: number` instead of giving it `number | undefined` type and then set it to undefined?

src/dateutils.ts
```diff
@@ -58,6 +58,15 @@ export function month(date: XDate) {
   return fromTo(firstDay, lastDay);
 }
 
+export function weeks(firstDate: XDate, numberOfWeeks: number) {
```
Inbal-Tish, line 61: Why are you exporting this function?

src/dateutils.ts
kkacquah: Will fix

src/dateutils.ts
```diff
@@ -58,6 +58,15 @@ export function month(date: XDate) {
   return fromTo(firstDay, lastDay);
 }
 
+function weeks(firstDate: XDate, numberOfWeeks: number) {
```
Inbal-Tish, line 61: Little comment about naming, why the 'first' prefix on the param and the consts (can't it just be 'date' or 'startDate', 'year' etc)? Especially when you eventually assign it to 'lastDate' const... (which should rename 'endDate').

src/calendar/index.tsx
```diff
@@ -69,6 +69,8 @@ export interface CalendarProps extends CalendarHeaderProps, DayProps {
   customHeader?: any;
   /** Allow selection of dates before minDate or after maxDate */
   allowSelectionOutOfRange?: boolean;
+  /** Max number of weeks to show on each month of month view */
```
Inbal-Tish, line 72: From the 'current' date. You need to mention that the weeks will start on the date passed to the 'current' prop and not from the first of the month.

kkacquah: Yeah, let me change the name of the prop to make this more clear as well.

src/dateutils.ts
kkacquah: I can change it from start and end, instead of first and last. The thought here is that this returns a list of dates, given the `numberOfWeeks` argument. I can add a comment here making this more clear as well.

src/dateutils.ts
```diff
@@ -57,6 +57,22 @@ export function month(date: XDate) {
...
 }
+/**
+ * Return a range of days from provided startDate that extends in the future for the 
+ * number of weeks provided by numberOfWeeksFromCurrent.
+ * 
+ * @param startDate Date to start add numberOfWeeksFromCurrent to
+ * @param numberOfWeeksFromCurrent Number of weeks which should be added to provided startDate
+ * @returns 
+ */
+function getDaysFromWeeksFromCurrent(startDate: XDate, numberOfWeeksFromCurrent: number) {
```
Inbal-Tish, line 68: The name now is very confusing. 
How about 'weeksDays(date: XDate, numberOfWeeks: number)`; to keep the naming similar to the other methods in this util? (Notice that your param name doesn't have to be equal to the prop name you pass to it)

src/dateutils.ts
```diff
@@ -57,6 +57,22 @@ export function month(date: XDate) {
...
   return fromTo(firstDay, lastDay);
 }
+/**
+ * Return a range of days from provided startDate that extends in the future for the 
+ * number of weeks provided by numberOfWeeksFromCurrent.
+ * 
+ * @param startDate Date to start add numberOfWeeksFromCurrent to
+ * @param numberOfWeeksFromCurrent Number of weeks which should be added to provided startDate
+ * @returns 
+ */
```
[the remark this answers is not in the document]
kkacquah: I can remove this comment

src/dateutils.ts
kkacquah: Will fix

src/dateutils.ts
```diff
@@ -58,6 +58,15 @@ export function month(date: XDate) {
   return fromTo(firstDay, lastDay);
 }
 
+function weeksDays(startDate: XDate, numberOfWeeks: number) {
```
Inbal-Tish, line 61: One last thing, sorry for not mentioning this before, please add tests for the function in the dateutils.spec file

PR #1699, opened 2021-12-06
Agenda renders only for the first item of the day

bmatasar:
The doc states :
`// Specify how each date should be rendered. day can be undefined if the item is not first in that day`
However, the tag 1.1269.0  has broken the behaviour and `Agenda.renderDate` is called only for the first item of the day now.

### DIFF
--- a/src/agenda/reservation-list/reservation.tsx
+++ b/src/agenda/reservation-list/reservation.tsx
@@ -75,8 +75,8 @@ class Reservation extends Component<ReservationProps> {
   }
 
   renderDate(date?: XDate, item?: DayReservations) {
-    if (isFunction(this.props.renderDay) && date) {
-      return this.props.renderDay(date, item);
+    if (isFunction(this.props.renderDay)) {
+      return this.props.renderDay(date ? xdateToData(date) : undefined, item);
     }
 
     const today = date && isToday(date) ? this.style.today : undefined;
--- a/src/agenda/reservation-list/reservation.tsx
+++ b/src/agenda/reservation-list/reservation.tsx
@@ -19,8 +19,8 @@ export interface ReservationProps {
   theme: Theme;
   /** specify your item comparison function for increased performance */
   rowHasChanged?: (a: any, b: any) => boolean;
-  /** specify how each date should be rendered. day can be undefined if the item is not first in that day */
-  renderDay?: (date: XDate, item?: DayReservations) => React.Component;
+  /** specify how each date should be rendered. date can be undefined if the item is not first in that day */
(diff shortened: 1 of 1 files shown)


src/agenda/reservation-list/reservation.tsx
```diff
@@ -75,8 +76,8 @@ class Reservation extends Component<ReservationProps> {
   }
 
   renderDate(date?: XDate, item?: DayReservations) {
-    if (isFunction(this.props.renderDay) && date) {
-      return this.props.renderDay(date, item);
+    if (isFunction(this.props.renderDay)) {
+      return this.props.renderDay(date ? xdateToData(date) : undefined, item);
```
Inbal-Tish, line 80: You're changing the return value, it should be XDate, not DateData object

Inbal-Tish: You'll also need to change type on line 23 to `renderDay?: (**date?**: XDate, item?: DayReservations) => React.Component;`

bmatasar: Actually I should only pass date which is `XDate`, same as the type required in `renderDay`.

Inbal-Tish (approved):

src/agenda/reservation-list/reservation.tsx
```diff
@@ -75,7 +75,7 @@ class Reservation extends Component<ReservationProps> {
   }
 
   renderDate(date?: XDate, item?: DayReservations) {
-    if (isFunction(this.props.renderDay) && date) {
+    if (isFunction(this.props.renderDay)) {
       return this.props.renderDay(date, item);
```
Inbal-Tish, line 79: You need to change the type on 'renderDay' prop. date param must be optional

Inbal-Tish (changes requested):

src/agenda/reservation-list/reservation.tsx
Inbal-Tish: That's the ts error you get: `src/agenda/reservation-list/reservation.tsx(79,35): error TS2345: Argument of type 'XDate | undefined' is not assignable to parameter of type 'XDate'.
14:56:22   Type 'undefined' is not assignable to type 'XDate'.`

Inbal-Tish (approved):

bmatasar:
Tish, sorry to waste your time with my sloppiness. Thanks!

Inbal-Tish:
> Tish, sorry to waste your time with my sloppiness. Thanks!

No worries. Thank you for the PR!!!

PR #1701, opened 2021-12-07
Calendars date format service

Inbal-Tish:
Adding a service to format the required date for Calendars' components


src/services/index.ts
```diff
@@ -0,0 +1,16 @@
+import {isString, isUndefined} from 'lodash';
+import XDate from 'xdate';
+const {parseDate, toMarkingFormat} = require('../interface');
+
+export function getCalendarDateString(date?: any) {
```
ethanshar, line 5: Why are you using `any` for the date type?

Inbal-Tish: You can pass different date formats and I'll parse it and return the date string in the required format. For example, you can pass Date(), number timestamp and DateData object, like {timestamp: xxxx} or {year: xxxx, month: xx, day: xx}

ethanshar: I personally think you should support a single format/type.
No reason to make it difficult for us, the user should know what to pass and should handle on their end. 

Anyway, if you do decide to support multiple formats, I suggest having a union type and not `any`. 
for example: `date?: number | string | Date` 

In any case I would avoid the dateData object cause it's not a trivial date type and maybe we'll end up deprecating it during our refactor, so better not have it here.

Inbal-Tish: If I support only one format then I don't see a point in this tool - The user can pass the required string format and that's it. I can limit the support to the types you mention as they make the most sense. The support for DateData was for values returned from Calendars, as there are few callbacks that return this type and I thought it will benefit the user that it doesn't need to parse or extract from it in order to resent a value to Calendars, but it's not a must...

Inbal-Tish: Done

src/services/index.ts
```diff
@@ -0,0 +1,17 @@
...
+  if (!isUndefined(date)) {
+    if (isDate(date) && !isNaN(date.getFullYear())) {
+      return date.getFullYear() + '-' + padNumber(date.getMonth() + 1) + '-' + padNumber(date.getDate());
+    } else if (isString(date)) {
+      // issue with strings and XDate's utc-mode - returns one day before
+      return toMarkingFormat(new XDate(date, false));
+    } else if (isNumber(date)) {
+      return toMarkingFormat(new XDate(date, true));
+    }
+    return 'Invalid Date';
```
ethanshar, line 15: One more thing, I missed that before. 
instead of returning 'Invalid Date` string, you can throw an error (with the `Invalid Date` message).

ethanshar (approved):

ethanshar (approved):

PR #1704, opened 2021-12-09
Fix/ts configuration

ethanshar:
- I fixed the TS configuration in tsconfig file so our example screens can recognize our typings
- I currently only added a single example screen (`example/src/screens/calendars.tsx`)
  - Once we'll migrate all examples screen we can change in tsconfig to include "example/**/*"
  - If you're planning to start fixing other example screens, please do one at a time to avoid big PRS
- Other than that I fixed some wrong typing I found, the main one is the `current` typing which should be `string`, we'll have to fix it internally with the refactor we talked about.

### DIFF
--- a/src/expandableCalendar/week.tsx
+++ b/src/expandableCalendar/week.tsx
@@ -12,7 +12,8 @@ import Day from '../calendar/day/index';
 // import BasicDay from '../calendar/day/basic';
 
 
-interface Props extends CalendarProps {
+// TODO: current type should be a string
+interface Props extends Omit<CalendarProps, 'current'> {
   current: XDate;
 }
 export type WeekProps = Props;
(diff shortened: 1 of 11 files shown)


src/expandableCalendar/week.tsx
```diff
@@ -12,7 +12,8 @@ import Day from '../calendar/day/index';
 // import BasicDay from '../calendar/day/basic';
 
 
-interface Props extends CalendarProps {
+// TODO: current type should be a string
+interface Props extends Omit<CalendarProps, 'current'> {
```
Inbal-Tish, line 16: Why no do the same for ExpandableCalendar until we start the migration from XDate to string?

Inbal-Tish (approved):

PR #1712, opened 2021-12-13
Types mismatch

bmatasar:
`CalendarList.current` should be `string`, not `XDate`.
If I create an `<Agenda selected="2021-12-13" />` or `<Calendar current="2021-12-13" />`, I get the error:

> Warning: Failed prop type: Invalid prop `current` of type `object` supplied to `CalendarList`, expected `string`.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -21,6 +21,6 @@ steps:
       npm install
       npm run test
       npm run build:ts
-      if [[ $(echo $BUILDKITE_MESSAGE | tr '[:upper:]' '[:lower:]') =~ release ]];then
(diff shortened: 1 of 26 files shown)


Inbal-Tish:
@bmatasar Please see latest version

bmatasar:
Tish, you need one more change to make it work. I rolled back the one already fixed.

bmatasar:
Without this fix, it shows warnings in the case  `<Agenda selected="2021-12-13" />`.

bmatasar:
It is already included in #1721

Inbal-Tish:
@bmatasar Hi. Yes, I'm working on a type fix for the component. Thanks for the PR anyway!!!

PR #1714, opened 2021-12-14
Change release process to publish snapshot versions for master commits

ethanshar:
@Inbal-Tish 
I changed the script and the pipeline so we'll have `snapshot` version for each master build, exactly like we have in public uilib. 
Please take a look.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -21,6 +21,6 @@ steps:
       npm install
       npm run test
       npm run build:ts
-      if [[ $(echo $BUILDKITE_MESSAGE | tr '[:upper:]' '[:lower:]') =~ release ]];then
+      if [[ $BUILDKITE_PULL_REQUEST == 'false' ]];then
         npm run release
       fi
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -6,11 +6,14 @@ const exec = require('shell-utils').exec;
 const cp = require('child_process');
 
 let IS_SNAPSHOT;
-if (process.env.BUILDKITE_MESSAGE.match(/^release$/i)){
+const isReleaseBuild = process.env.BUILDKITE_MESSAGE.match(/^release$/i);
+const isPRBuild = process.env.BUILDKITE_PULL_REQUEST === 'true';
+
+if (isReleaseBuild) {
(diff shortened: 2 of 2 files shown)


scripts/release.js
```diff
@@ -6,11 +6,14 @@ const exec = require('shell-utils').exec;
...
-if (process.env.BUILDKITE_MESSAGE.match(/^release$/i)){
+const isReleaseBuild = process.env.BUILDKITE_MESSAGE.match(/^release$/i);
+const isPRBuild = process.env.BUILDKITE_PULL_REQUEST === 'true';
+
+if (isReleaseBuild) {
   IS_SNAPSHOT = cp.execSync(`buildkite-agent meta-data get is-snapshot`).toString();
 }
 const ONLY_ON_BRANCH = 'release';
-const isSnapshotBuild = IS_SNAPSHOT === 'true';
+const isSnapshotBuild = !isPRBuild || IS_SNAPSHOT === 'true';
```
Inbal-Tish, line 16: It will build a snapshot version for 'isReleaseBuild' since it will pass the '!isPRBuild' condition, so the 'IS_SNAPSHOT' is obsolete... The condition should be `!isPRBuild && !isReleaseBuild || IS_SNAPSHOT === 'true'`

ethanshar: right.. fixing

.buildkite/pipeline.yml
```diff
@@ -21,6 +21,6 @@ steps:
       npm install
       npm run test
       npm run build:ts
-      if [[ $(echo $BUILDKITE_MESSAGE | tr '[:upper:]' '[:lower:]') =~ release ]];then
```
Inbal-Tish, line 24: I think this line checks if the message includes the word "release" to know if to have the release option. By changing it to check if it's not a PR means you'll have the release button on any build (even though you won't be able to actually release since the script checks if you're on 'release' branch...)

scripts/release.js
ethanshar: Fixed

.buildkite/pipeline.yml
ethanshar: I don't think so.. 
What determines if to show the release button is at the top of the file. 
This is how the pipeline of uilib looks also

Inbal-Tish: Ok

Inbal-Tish (approved):

PR #1716, opened 2021-12-14
Infra/removing x date type

Inbal-Tish:
Removing XDate as a type from props in AgendaList, Week and Provider components


ethanshar:
@Inbal-Tish 
What about fixing the relevant example screens as well. 
So you'll have a validation the types are working as expected

Inbal-Tish:
> @Inbal-Tish What about fixing the relevant example screens as well. So you'll have a validation the types are working as expected

I have really strange ts errors in this screen...

Inbal-Tish:
> @Inbal-Tish What about fixing the relevant example screens as well. So you'll have a validation the types are working as expected

Done

ethanshar (approved):

PR #1720, opened 2021-12-14
Adding JSX.Element on Agenda reservation types

mateusvidaljs:
I had a need to use a JSX.Element on renderItem method, because my application runs with Function Components instead Class Components, and i had to add this JSX.Element type inside the reservation.ts file, and it runs ok, without any bugs or problems!

Captura de Tela 2021-12-14 às 10 12 07


Inbal-Tish:
@mateusvidaljs Hi. I'm working on a PR to fix all Agenda types and I'm including this fix as well. Thank you for the PR!!!

PR #1739, opened 2021-12-29
Support Timeline now indicator and fix UI for timeline grid

ethanshar:
- Support showing Now Indicator for the Timeline component. 
- Improve UI for timeline grid

### DIFF
--- a/src/timeline/NowIndicator.tsx
+++ b/src/timeline/NowIndicator.tsx
@@ -1,0 +1,27 @@
+import React from 'react';
+import {View, TextStyle, ViewStyle} from 'react-native';
+import {HOUR_BLOCK_HEIGHT} from './Packer';
+
+export {HOUR_BLOCK_HEIGHT} from './Packer';
+
+export interface NowIndicatorProps {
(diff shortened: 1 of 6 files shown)


src/timeline/NowIndicator.tsx
```diff
@@ -0,0 +1,27 @@
...
+import {View, TextStyle, ViewStyle} from 'react-native';
+import {HOUR_BLOCK_HEIGHT} from './Packer';
+
+export {HOUR_BLOCK_HEIGHT} from './Packer';
+
+export interface NowIndicatorProps {
+  styles: {[key: string]: ViewStyle | TextStyle};
+}
+
+const TodayIndicator = (props: NowIndicatorProps) => {
```
lidord-wix, line 11: Maybe better to call it `NowIndicator` here as well

lidord-wix (approved):
Small comment, feel free to change and merge

src/timeline/NowIndicator.tsx
ethanshar: Right, missed that one. Fixed!

PR #1740, opened 2021-12-30
Support passing date to Timeline

ethanshar:
This for identifying the Timeline instance, especially later on when we'll support 3 days view

### DIFF
--- a/example/src/screens/timelineCalendar.tsx
+++ b/example/src/screens/timelineCalendar.tsx
@@ -103,16 +103,17 @@ export default class TimelineCalendarScreen extends Component {
   };
 
   createNewEvent: TimelineProps['onBackgroundLongPress'] = (timeString, timeObject) => {
-    const {currentDate} = this.state;
     const hourString = `${(timeObject.hour + 1).toString().padStart(2, '0')}`;
     const minutesString = `${timeObject.minutes.toString().padStart(2, '0')}`;
 
     const newEvent = {
-      start: `${currentDate} ${timeString}`,
-      end: `${currentDate} ${hourString}:${minutesString}:00`,
+      start: `${timeString}`,
+      end: `${timeObject.date} ${hourString}:${minutesString}:00`,
       title: 'New Event',
       color: '#ffffff'
     };
+
+    console.log('ethan - what', newEvent)
 
     this.setState({newEvent});
   };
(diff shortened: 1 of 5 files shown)


Inbal-Tish:
@ethanshar 
/Users/builder/work/example/src/screens/timelineCalendar.tsx -- error  Missing semicolon  semi

example/src/screens/timelineCalendar.tsx
```diff
@@ -103,17 +103,18 @@ export default class TimelineCalendarScreen extends Component {
...
     const newEvent = {
-      start: `${currentDate} ${timeString}`,
-      end: `${currentDate} ${hourString}:${minutesString}:00`,
+      start: `${timeString}`,
+      end: `${timeObject.date} ${hourString}:${minutesString}:00`,
       title: 'New Event',
       color: '#ffffff'
     };
 
+    console.log('ethan - what', newEvent)
```
Inbal-Tish, line 116: log

example/src/screens/timelineCalendar.tsx
```diff
@@ -158,6 +159,7 @@ export default class TimelineCalendarScreen extends Component {
           markedDates={this.marked}
         />
         <Timeline
+          date={currentDate}
```
Inbal-Tish, line 162: Since this is a new prop, the existing usages do not pass it to the timeline component, thus a new event is not created for them. Maybe we should mention it in the prop/feature description (i.e. 'onBackgroundLongPress' / 'date' props)?

src/timeline/Timeline.tsx
```diff
@@ -13,6 +13,13 @@ const LEFT_MARGIN = 60 - 1;
 const {width: dimensionWidth} = Dimensions.get('window');
 
 export interface TimelineProps {
+  /**
+   * The date of this timeline instance
+   */
+  date?: string;
```
Inbal-Tish, line 19: Is there a required format to the date string? If so you should add it to the description

example/src/screens/timelineCalendar.tsx
Inbal-Tish: Is there a way to determine internally on which date the user long-pressed without passing the date from the screen?

example/src/screens/timelineCalendar.tsx
ethanshar: oops fixed

example/src/screens/timelineCalendar.tsx
ethanshar: I can default it to a `new Date()` but that won't be necessarily true when we'll have 3 days view. 
Anyway, I will update the prop description to make it more clear

src/timeline/Timeline.tsx
ethanshar: Fixed

example/src/screens/timelineCalendar.tsx
ethanshar: Done

Inbal-Tish (approved):

PR #1750, opened 2022-01-13
Convert HorizontalList to InfiniteList

ethanshar:
- I renamed HorizontalList to InfiniteList and moved it out of timeline-list folder so it'll be a common component for all other components to use. 
- I made the component more dynamic so it can now accept pageWidth and pageHeight for different types of pages
- I added APIs for onReachEdge and onReachNearEdge since it's a logic that is needed in all cases (instead of each component to calculate that)

### DIFF
--- a/src/infinite-list/index.tsx
+++ b/src/infinite-list/index.tsx
@@ -3,31 +3,24 @@ import React, {forwardRef, useCallback, useMemo, useRef} from 'react';
 import {DataProvider, LayoutProvider, RecyclerListView, RecyclerListViewProps} from 'recyclerlistview';
 import constants from '../commons/constants';
 
-const layoutProvider = new LayoutProvider(
-  () => 'page',
-  (_type, dim) => {
-    dim.width = constants.screenWidth;
-    dim.height = constants.screenHeight;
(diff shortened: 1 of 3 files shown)


src/infinite-list/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+export interface InfiniteListProps
+  extends Omit<RecyclerListViewProps, 'dataProvider' | 'layoutProvider' | 'rowRenderer'> {
+  data: any[];
+  renderItem: RecyclerListViewProps['rowRenderer'];
+  pageWidth?: number;
+  pageHeight?: number;
+  onPageChange?: (pageIndex: number, prevPageIndex: number) => void;
+  onReachEdge?: (pageIndex: number) => void;
+  onReachNearEdge?: (pageIndex: number) => void;
+  nearEdgeThreshold?: number;
```
Inbal-Tish, line 20: `onReachNearEdgeThreshold` is a better name

ethanshar: Done

Inbal-Tish (approved):

PR #1751, opened 2022-01-15
Infra/use infinite list for week cal

ethanshar:
I created a new implementation for WeekCalendar. 
Instead of refactoring the existing one I created a new file in the same folder (new.tsx) 

- I refactored the component into a Function Component
- I'm getting the ExpandableCalendar context using `useContext` hook instead of the `asCalendarConsumer`
- I replaced FlatList with our new InfiniteList component
- I moved the week days names into a separate sub component (WeekDaysNames.tsx) 
- I didn't use the existing presenter cause it contains too much logic I didn't need anymore, the only thing I needed I copied to the `new.tsx` file - this is temporary, eventually I'll create a new presenter instead of the one we have today. 

At the moment only the WeekCalendar example screen is using the new implementation, Other components still use the old implementation. 
Gradually, we'll move to the new one when we feel safe with it.


Inbal-Tish:
@ethanshar There's a bug when you scroll the week calendar and then press the 'Today' button, the calendar doesn't scroll to the today's date.

ethanshar:
> @ethanshar There's a bug when you scroll the week calendar and then press the 'Today' button, the calendar doesn't scroll to the today's date.

Fixed

Inbal-Tish (approved):

Inbal-Tish:
@ethanshar So good!! Great job!
I do think `WeekDaysNames` component should be common so we can use it in other components as well. Decide if you want to change it now and merge.

ethanshar:
> @ethanshar So good!! Great job!
> I do think `WeekDaysNames` component should be common so we can use it in other components as well. Decide if you want to change it now and merge.

I completely agree, I prefer we'll do it in another PR or refactor when another component will have to use it.

Issue #1686, opened 2021-11-17
 🔥🐞 Attempted import error: 'AgendaListProps' is not exported from './expandableCalendar/agendaList'.


FrancisLeigh:
## Description

I'm trying to install `react-native-calendars` onto our React-Native for Web x NextJS platform. 
I have setup the `next-transpile-modules` 'module in our `next.config.js` which helped with the initial `static Proptypes` transpilation issue but now i am getting an export error.
## Expected Behavior
Expecting the module to work 😉

## Observed Behavior
Page fails to load with error

### Error
```ruby
./node_modules/react-native-calendars/src/index.ts
Attempted import error: 'AgendaListProps' is not exported from './expandableCalendar/agendaList'.
```

## Environment

Please run these commands in the project folder and fill in their results:

* `npm ls react-native-calendars`: 
└── react-native-calendars@1.1269.0
* `npm ls react-native`:
└─┬ @react-native-community/cli@4.14.0
  └── react-native@0.63.2 deduped

* `npm ls next`
  ├─┬ next-i18next@8.8.0
│ └── next@10.1.2 deduped
├─┬ next-router-mock@0.1.4
│ └── next@10.1.2 deduped
└── next@10.1.2

* `npm ls react`
  ├─┬ @apollo/client@3.4.12
│ └── react@17.0.2 deduped
├─┬ @emotion/react@11.5.0
│ └── react@17.0.2 deduped
├─┬ @emotion/styled@11.3.0
│ └── react@17.0.2 deduped
├─┬ @fortawesome/react-fontawesome@0.1.15
│ └── react@17.0.2 deduped
├─┬ @maisonsport/common-ui@1.3.8
│ ├─┬ react-native@0.63.2
│ │ └── react@17.0.2 deduped invalid: "16.13.1" from node_modules/react-native
│ ├─┬ react-test-renderer@16.14.0
│ │ └── react@17.0.2 deduped invalid: "16.13.1" from node_modules/react-native, "^16.14.0" from node_modules/react-test-renderer
│ ├── react@17.0.2 deduped
│ └─┬ styled-components@5.3.3
│   └── react@17.0.2 deduped invalid: "16.13.1" from node_modules/react-native, "^16.14.0" from node_modules/react-test-renderer
├─┬ @material-ui/unstyled@5.0.0-alpha.44
│ ├─┬ @material-ui/utils@5.0.0-beta.5
[...]

PR #1752, opened 2022-01-16
fix type import error

connectdotz:
follow up #1745 and fix the remaining type import error reported in [comment](https://github.com/wix/react-native-calendars/issues/1686#issuecomment-1013966402)


leikoilja (approved):
Nice fix, thanks

Inbal-Tish (approved):

PR #1756, opened 2022-01-18
Support initial scroll to current time

ethanshar:
Add `scrollToNow` prop in Timeline components


example/src/screens/timelineCalendar.tsx
```diff
@@ -9,95 +9,98 @@ import {
   CalendarUtils
 } from 'react-native-calendars';
 import _ from 'lodash';
+import XDate from 'xdate';
```
Inbal-Tish, line 12: The screens should not use XDate

src/timeline/Timeline.tsx
```diff
@@ -100,24 +108,25 @@ const Timeline = (props: TimelineProps) => {
...
+    const width = constants.screenWidth - HOURS_SIDEBAR_WIDTH;
     return populateEvents(events, width, start);
   }, [events, start]);
 
   useEffect(() => {
+    let initialPosition = 0;
     if (scrollToFirst) {
-      const firstTop = min(map(packedEvents, 'top')) ?? 0;
-      const initPosition = firstTop - calendarHeight.current / (end - start);
-      const verifiedInitPosition = initPosition < 0 ? 0 : initPosition;
```
Inbal-Tish, line 111: What if initPosition is less than 0?

src/timeline-list/index.tsx
```diff
@@ -82,6 +83,7 @@ const TimelineList = (props: TimelineListProps) => {
             {...timelineProps}
             key={item}
             date={item}
+            scrollToNow={scrollToNow && index === INITIAL_PAGE}
```
Inbal-Tish, line 86: The 'date' passed to 'timelineProps' determines the date of the middle page (the INITIAL_PAGE), this date is not necessarily today's date, so you won't be able to scroll to now (the current time on today's date). So I'm not sure your condition is correct... Maybe you need to check that this page is today's page? Did I understand correctly am I missing something?

example/src/screens/timelineCalendar.tsx
ethanshar: Technically I just used it for formatting the strings, I didn't pass it to the component. 
Anyway, fair enough, I replaced it with `Date`

src/timeline/Timeline.tsx
ethanshar: You can see in line 126, that I maxing it to `0` in case we get a negative value
```
y: Math.max(0, initialPosition - HOUR_BLOCK_HEIGHT),
```

Inbal-Tish: Yes, but you're maxing it after you reduce the HOUR_BLOCK_HEIGHT... but ok, you're not allowing a negative offset at the top to it doesn't matter

src/timeline-list/index.tsx
ethanshar: Yes, that's a good point. 
I checked again with the product, the behavior might be a little different. 
Ill change it and update

ethanshar: Ok, I fixed the logic.
First of all, I added a new prop - `initialTime` that defines the initial time to scroll to
In TimelineList I do the following
- If initial page is "today" - I pass the scrollToNow given by the user
- If the initial page is not "today" - I pass the `initialTime` prop 

The reset of the timeline pages (any page that is not the initial) should be sync accordingly, based on the `scrollOffset` prop

lidord-wix: I find it a little bit confusing.. I'd guess that the initialTime will work for all cases, even for the current day

lidord-wix: Also, I think it's weird that the `scrollToNow` prop works only when the `initialPage` is the current day.
I think we can have only one prop - `initialTime` and the user will decide whether to pass the current time or any other time..

lidord-wix (commented):
Looks good! I wrote one comment (in addition to those I commented before :))

src/timeline-list/index.tsx
ethanshar: TBH, these requirements came from Bookings product, but they do make sense.. 
Also, I changed the logic a little more since the last time.
There are few "initial scroll" logics, I'll explain them from the most prioritized to the least

- **scrollToNow** - will scroll to current time. Specifically, In the context of TimelineList it's relevant only when the initial page is "today"
- **scrollToFirst** - will scroll to the first event of the day (if there are events that day). Again, in the context of TimelineList it's relevant only if the initial page is NOT "today"
- **initialTIme** - Will scroll to an initial time (usually represent the business opening hours) - and like before, In the context of TimelineList, it's relevant only if the initial page is NOT "today" 

Logically it make sense to scroll to the current time only when you're in "today", otherwise the current time has no meaning. 
And again, all this logic refers to the initial page and the initial scroll. This logic won't take affect when moving between the pages/timelines in the list

src/timeline/helpers/presenter.ts
```diff
@@ -10,3 +10,10 @@ export function calcTimeByPosition(yPosition: number, hourBlockHeight: number) {
 export function buildTimeString(hour = 0, minutes = 0, date = '') {
   return `${date} ${hour.toString().padStart(2, '0')}:${minutes.toString().padStart(2, '0')}:00`.trimStart();
 }
+
+export function calcNowOffset(hourBlockHeight: number, hour?: number, minutes?: number) {
```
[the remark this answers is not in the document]
ethanshar: done 👌

lidord-wix (approved):

src/timeline-list/index.tsx
Inbal-Tish: I think the 2 last types should work for **any** initial page (today or not) since they have a very specific intent that does not exclude today's page (which will be the initial page in most cases), while the 'scrollToNow' should work only if the current page is today since it doesn't make sense otherwise (as it is currently).

PR #1758, opened 2022-01-19
Feature: Allow the user to select a week via a checkbox + clicking the knob in expandableCalendar

riamon-v:
Hi !

I needed in my app a way to select a week through the calendar so it can give me back the week number.
I made a small feature to do so, here is an example

ezgif-5-f484b1e1c3

This PR also include the "TODO", clickable knob in expandable calendar. I also added a callback in case the developer wants to do something when the knob is clicked.

Here it is:
ezgif-5-e3a2a5a0b7


riamon-v:
@ethanshar or @Inbal-Tish are you guys available to review this PR ?

Inbal-Tish:
@riamon-v ExpandableCalendar's knob press to close was added (the callback fires is `onCalendarToggled`). 
I don't really understand the "selecting week" feature. What is happening to the Calendar when you press the checkbox? Is it only for the callback's info?

riamon-v:
Hey @Inbal-Tish , I needed in my app a way to select a week to know which week the user will have been available or not. When pressing the checkbox it  calls the callback passed from props to know which week you have clicked on.

Inbal-Tish:
@riamon-v Then why not make the week numbers (with `showWeekNumbers` true) pressable when passing a callback (like `onWeekNumberPressed`) that returns the week number?

riamon-v:
> @riamon-v Then why not make the week numbers (with `showWeekNumbers` true) pressable when passing a callback (like `onWeekNumberPressed`) that returns the week number?

IMO it was not very clear to the user that the week by clicking the week number you would actually select it for another use case. I wanted to bring a clearer way to let the user know you could select weeks

Inbal-Tish:
> > @riamon-v Then why not make the week numbers (with `showWeekNumbers` true) pressable when passing a callback (like `onWeekNumberPressed`) that returns the week number?
> 
> IMO it was not very clear to the user that the week by clicking the week number you would actually select it for another use case. I wanted to bring a clearer way to let the user know you could select weeks

Yes, I understand. I would look for a UX solution for that though and add a support for styling the week numbers so you can pass the proper style to indicate the user it is pressable. What do you think?

riamon-v:
> > > @riamon-v Then why not make the week numbers (with `showWeekNumbers` true) pressable when passing a callback (like `onWeekNumberPressed`) that returns the week number?
> > 
> > 
> > IMO it was not very clear to the user that the week by clicking the week number you would actually select it for another use case. I wanted to bring a clearer way to let the user know you could select weeks
> 
> Yes, I understand. I would look for a UX solution for that though and add a support for styling the week numbers so you can pass the proper style to indicate the user it is pressable. What do you think?

Yep could be promising !

PR #1760, opened 2022-01-19
Update agenda.tsx

eladgel:
The documentations is wrong 
Since XDATE is trying to parse ISO dates as strings we should pass 

a date object like so:
new Date().toISOString(). and not like in the example

### DIFF
--- a/example/src/screens/agenda.tsx
+++ b/example/src/screens/agenda.tsx
@@ -18,7 +18,8 @@ export default class AgendaScreen extends Component<State> {
         testID={testIDs.agenda.CONTAINER}
         items={this.state.items}
         loadItemsForMonth={this.loadItems}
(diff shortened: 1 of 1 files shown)


example/src/screens/agenda.tsx
```diff
@@ -18,7 +18,8 @@ export default class AgendaScreen extends Component<State> {
         testID={testIDs.agenda.CONTAINER}
         items={this.state.items}
         loadItemsForMonth={this.loadItems}
-        selected={'2017-05-16'}
+        // make sure you pass ISOString - new Date().toISOString()
+        selected={'2022-01-19T22:02:59.246Z'}
```
Inbal-Tish, line 22: There is no need for that. XDate can handle strings in a 'yyyy-MM-dd' format.

PR #1765, opened 2022-01-26
Support passing right spacing for overlapping Timeline events

ethanshar:
First of all, I refactored the Packer util for Timeline component. 
I mainly changed some function names, fixed typings and add some comments to explain the code (the part of code is very not intuitive) 

Also, I added support for injecting options to the Packer util and specifically in this PR added support to pass a custom right spacing between overlapping events.

### DIFF
--- a/example/src/screens/timelineCalendar.tsx
+++ b/example/src/screens/timelineCalendar.tsx
@@ -224,7 +224,7 @@ export default class TimelineCalendarScreen extends Component {
     // scrollToFirst: true,
     // start: 0,
     // end: 24,
-    eventRightSpacing: 4
+    overlapEventsSpacing: 8
   };
 
   render() {
--- a/example/src/screens/timelineCalendar.tsx
+++ b/example/src/screens/timelineCalendar.tsx
@@ -29,6 +29,20 @@ const EVENTS: TimelineEventProps[] = [
     color: '#e6add8'
   },
   {
+    start: `${getDate()} 01:45:00`,
+    end: `${getDate()} 02:45:00`,
+    title: 'Dr. Mariana Joseph',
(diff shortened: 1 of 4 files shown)


lidord-wix (commented):
Looks good overall, but I think we should have a prop called `spaceBetweenOverlappedEvents` for example, that will be applied between overlapped events and not for all the events (because `eventRightSpacing` applied to all the events).

ethanshar:
> Looks good overall, but I think we should have a prop called `spaceBetweenOverlappedEvents` for example, that will be applied between overlapped events and not for all the events (because `eventRightSpacing` applied to all the events).

You mean that the name (`eventRightSpacing`) hints that it applies to all events?

lidord-wix:
> > Looks good overall, but I think we should have a prop called `spaceBetweenOverlappedEvents` for example, that will be applied between overlapped events and not for all the events (because `eventRightSpacing` applied to all the events).
> 
> You mean that the name (`eventRightSpacing`) hints that it applies to all events?

Yes, and it does apply to events. but we should also provide a way to control only the space between overlapped events

ethanshar:
> > > Looks good overall, but I think we should have a prop called `spaceBetweenOverlappedEvents` for example, that will be applied between overlapped events and not for all the events (because `eventRightSpacing` applied to all the events).
> > 
> > 
> > You mean that the name (`eventRightSpacing`) hints that it applies to all events?
> 
> Yes, and it does apply to events. but we should also provide a way to control only the space between overlapped events

Ok, I fixed it. 
Not overlapping event don't get this spacing and I renamed to `overlapEventsSpacings`

lidord-wix (approved):

PR #1767, opened 2022-01-27
fix expandable calendar custom styling

timhubert:
For custom styling as in the other components

### DIFF
--- a/src/expandableCalendar/style.ts
+++ b/src/expandableCalendar/style.ts
@@ -173,6 +173,7 @@ export default function styleConstructor(theme: Theme = {}) {
       marginLeft: appStyle.todayButtonPosition === 'right' ? 7 : undefined,
       marginRight: appStyle.todayButtonPosition === 'right' ? undefined : 7
     },
-    ...(theme?.stylesheet?.expandable?.main || {})
+    // @ts-expect-error
+    ...(theme['stylesheet.expandable.main'] || {})
   });
 }


Inbal-Tish:
@timhubert Hi. Thank you for the PR. Can you please explain why the other usage is not working and add a code snippet? I want to understand what is the reason for the code change.

timhubert:
Hi @Inbal-Tish, the override works, but the form of how to override is different for this one component as it is for the other components.

E.g. In the other components and and according to [documentation](https://github.com/wix/react-native-calendars#advanced-styling):
```...(theme['stylesheet.day.basic'] || {})```
Stylesheet:
```'stylesheet.day.basic': { text: { ... } }```


Only this one component
`...(theme?.stylesheet?.expandable?.main || {})`
Stylesheet:
```stylesheet: { expandable: { main: { text: { ... } } } }```

It's not a big deal, but it's annoying.

Inbal-Tish:
@timhubert Hi. When you use this style in the demo screen and run `npm run build:dev`, do you get a ts error?

timhubert:
@Inbal-Tish No, I don't. Because of `// @ts-expect-error` (is exactly the same as in the other components)

Inbal-Tish:
@timhubert Hi. I understand the usage is different here and it shouldn't, but this is a breaking change so I can't remove the old usage. We can keep both until the next major version...

Seamus1989:
I want to override the day container - how do I do this? It flexes but II want to control its height and width which that the days in the calendar have no spacing between them?

Inbal-Tish:
@Seamus1989 Hi. Add a screenshot with your design and we'll try to assist.

PR #1772, opened 2022-01-31
Remove comments from old PropTypes

Inbal-Tish:
1) Remove comments from old propTypes for cleaner code.
2) Use 'sameMonth' instead of string compare.
3) Use 'getDefaultLocal'.
4) Move 'isPastDate' to dateutils.
5) Optional params in 'sameDate', 'sameMonth' and 'parseDate'.

### DIFF
--- a/src/agenda/index.tsx
+++ b/src/agenda/index.tsx
@@ -32,10 +32,6 @@ const HEADER_HEIGHT = 104;
 const KNOB_HEIGHT = 24;
 
 export type AgendaProps = CalendarListProps & ReservationListProps & {
-  /** the list of items that have to be displayed in agenda. If you want to render item as empty date
-   the value of date key has to be an empty array []. If there exists no value for date key it is
-   considered that the date in question is not yet loaded */
-  items?: AgendaSchedule;
   /** callback that gets called when items for a certain month should be loaded (month became visible) */
   loadItemsForMonth?: (data: DateData) => void;
   /** callback that fires when the calendar is opened or closed */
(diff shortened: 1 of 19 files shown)


src/calendar/header/index.tsx
```diff
@@ -30,12 +30,17 @@ import styleConstructor from './style';
 import {Theme, Direction} from '../../types';
 
 export interface CalendarHeaderProps {
+  month?: XDate;
```
lidord-wix, line 33: Should be replaced by string as well?

src/calendar/index.tsx
```diff
@@ -26,8 +26,6 @@ type MarkedDatesType = {
 };
 
 export interface CalendarProps extends CalendarHeaderProps, DayProps {
-  /** Specify theme properties to override specific styles for calendar parts */
-  theme?: Theme;
```
lidord-wix, line 30: Is it safe to remove all these props?

lidord-wix (commented):
Looks good!
I wrote few comments to make sure we don't break anything :)

src/agenda/index.tsx
```diff
@@ -32,10 +31,6 @@ const HEADER_HEIGHT = 104;
 const KNOB_HEIGHT = 24;
 
 export type AgendaProps = CalendarListProps & ReservationListProps & {
-  /** the list of items that have to be displayed in agenda. If you want to render item as empty date
-   the value of date key has to be an empty array []. If there exists no value for date key it is
-   considered that the date in question is not yet loaded */
-  items?: AgendaSchedule;
```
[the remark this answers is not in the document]
Inbal-Tish: Agenda get this prop from `ReservationListProps`. I don't see the need for the duplication

src/calendar/header/index.tsx
Inbal-Tish: It should be a private prop... We'll need to sort all these props issues in V2, it's currently a big mess

src/calendar/index.tsx
Inbal-Tish: They are header props, but I'll put them back and we'll sort these props issues in another occasion

Inbal-Tish: Done

src/agenda/index.tsx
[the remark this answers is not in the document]
Inbal-Tish: Done

lidord-wix (approved):

PR #1782, opened 2022-02-07
Fix tsc --isolatedModules error

ktalartukano:


Inbal-Tish:
@ktalartukano Thank you for the PR. I just merged a similar PR. The fix will be included in the next release. Thanks!!!

PR #1790, opened 2022-02-14
CalendarListItem inside TouchableOpacity

BernardoQuina:
CalendarListItem inside TouchableOpacity to prevent unreliable/janky scrolling when CalendarList is inside a scrollView or a swipeable modal.

This makes it way smoother for my use case.

Alternatively, would be great to have a prop that allowed to opt in for this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "memoize-one": "^5.2.1",
     "prop-types": "^15.5.10",
     "react-native-swipe-gestures": "^1.0.5",
+    "react-native-gesture-handler": "~2.1.0",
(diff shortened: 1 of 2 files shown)


BernardoQuina:
Also changed the FlatList import to use 'react-native-gesture-handler'. Now works super smooth both on ios and android when nested in other scrollable content.

Inbal-Tish:
@BernardoQuina Hi. Can you please elaborate on the issue you're trying to fix in this PR. A video with the existing vs the new feature will be very helpful. Thanks!

PR #1792, opened 2022-02-15
added a new key extractor to overcome an issue with flatlist

eladgel:
Added a new property

### DIFF
--- a/src/agenda/reservation-list/index.tsx
+++ b/src/agenda/reservation-list/index.tsx
@@ -44,6 +44,8 @@ export type ReservationListProps = ReservationProps & {
   refreshing?: boolean;
   /** If provided, a standard RefreshControl will be added for "Pull to Refresh" functionality. Make sure to also set the refreshing prop correctly */
   onRefresh?: () => void;
+
+  reservationKeyExtractor: PropTypes.Requireable<(...args: any[]) => any>,
 };
 
 interface DayAgenda {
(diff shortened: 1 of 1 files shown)


src/agenda/reservation-list/index.tsx
```diff
@@ -258,7 +260,9 @@ class ReservationList extends Component<ReservationListProps, State> {
     );
   };
 
-  keyExtractor = (_item: DayAgenda, index: number) => String(index);
+  keyExtractor = (_item: DayAgenda, index: number) => {
+      return this.props.reservationKeyExtractor && this.props.reservationKeyExtractor(_item, index) || String(index);
```
Inbal-Tish, line 264: You're not fixing the default key...

Inbal-Tish:
@eladgel Thank you for the pr. There is a lint error that fails the build and some comments. I opened a fix to move things faster - #1795. Hope that will do the trick. Thank you and looking forward to your collaboration again!

src/agenda/reservation-list/index.tsx
```diff
@@ -44,6 +44,8 @@ export type ReservationListProps = ReservationProps & {
   refreshing?: boolean;
   /** If provided, a standard RefreshControl will be added for "Pull to Refresh" functionality. Make sure to also set the refreshing prop correctly */
   onRefresh?: () => void;
+
+  reservationKeyExtractor: PropTypes.Requireable<(...args: any[]) => any>,
```
Inbal-Tish, line 48: No need for such an elaborate name, `keyExtractor` is good enough.

eladgel:
I thought that keyExtractor was already declared and used by Calendar List.
That's why I didn't use it again

PR #1794, opened 2022-02-15
Agenda Fixes: Generic Items, Key Extractor

### LINKED ISSUE #1791: `Agenda` list Items incorrectly rendering on update
Hello -- I think I found a bug in the `AgendaList` due to the current key extraction.

When I navigate back in the agenda, e.g. from a Monday to Sunday, and if both of those days have arrays of events, the FlatList isn't rendering the correct item. Instead, an item previously rendered gets used. I'm pretty sure this is because of the string coercion  of the indices used in the current key extraction: 

https://github.com/wix/react-native-calendars/blob/74bc7fe05ed3883d3d8cbd56e3cd5a565543c385/src/agenda/reservation-list/index.tsx#L261:
```tsx
  keyExtractor = (_item: DayAgenda, index: number) => String(index);
 ```
 
 this seems to me like a pretty good example of the index-key anti pattern described here:
 https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318
 
 This seems straightforward to fix, by at least exposing a `keyExtractor` prop to the `AgendaList`.
 
 I'll put together a minimal reproduction of this when I get the time.

### PR DESCRIPTION
**Pull request by JacobJaffe**
Hello! First, thanks for this library. This is an awesome set of tools.

This PR attempts to solve two issues with the current `Agenda`:
1. The component is not generic. Specifically, `AgendaEntry` is not generic, nor the functions that use it, such as `renderItem`
2. The Agenda has a bug with using indexes as the keys. This is causing rendering issues because data is falsely considered to be stable across renders. The result is that, while updating the `data` list of the underlying `FlatList` (either through calendar navigation, or `items` updating. Anything that touches `state.reservations`), items are duplicated and / or not shown. (this is documented at https://github.com/wix/react-native-calendars/issues/1791#issue-1138143626)

This is a backwards compatible solution to both of these. The types of the functions will now be inferred from the data type, and the key extractor can now be specified. (FWIW, I'm pretty sure the default prop of the index -> string extractor isn't needed. This is the default the flatlist would use anyways, and is bug prone. However, I've left it to keep this PR backwards compatible).

This should also:
close #1792 -- this attempts to add a fix to issue (2) above.
close #1683 -- this attempts to add a fix to issue (1) above.
close #1791

#1777 doesn't have much info, but it looks like this will solve that too.

Please let me know if this needs any updates, or any other suggestions.

### DIFF
--- a/src/agenda/reservation-list/index.tsx
+++ b/src/agenda/reservation-list/index.tsx
@@ -3,21 +3,30 @@ import PropTypes from 'prop-types';
 import XDate from 'xdate';
 
 import React, {Component} from 'react';
-import {ActivityIndicator, View, FlatList, StyleProp, ViewStyle, TextStyle, NativeSyntheticEvent, NativeScrollEvent, LayoutChangeEvent} from 'react-native';
+import {
+  ActivityIndicator,
+  View,
+  FlatList,
+  StyleProp,
+  ViewStyle,
+  TextStyle,
+  NativeSyntheticEvent,
+  NativeScrollEvent,
+  LayoutChangeEvent
+} from 'react-native';
 
 import {extractComponentProps} from '../../componentUpdater';
 import {sameDate} from '../../dateutils';
 import {toMarkingFormat} from '../../interface';
 import styleConstructor from './style';
 import Reservation, {ReservationProps} from './reservation';
-import {AgendaEntry, AgendaSchedule} from '../../types';
+import {AgendaSchedule} from '../../types';
 
-
-export type ReservationListProps = ReservationProps & {
+export type ReservationListProps<T> = ReservationProps<T> & {
   /** the list of items that have to be displayed in agenda. If you want to render item as empty date
   the value of date key kas to be an empty array []. If there exists no value for date key it is
   considered that the date in question is not yet loaded */
-  items?: AgendaSchedule;
+  items?: AgendaSchedule<T>;
   selectedDay?: XDate;
   topDay?: XDate;
   /** Show items only for the selected date. Default = false */
@@ -44,18 +53,20 @@ export type ReservationListProps = ReservationProps & {
   refreshing?: boolean;
   /** If provided, a standard RefreshControl will be added for "Pull to Refresh" functionality. Make sure to also set the refreshing prop correctly */
   onRefresh?: () => void;
+  /** Extractor for underlying FlatList. Ensure that this is unique per item, or else scrolling may have duplicated and / or missing items.  */
+  keyExtractor?: () => string;
 };
 
-interface DayAgenda {
-  reservation?: AgendaEntry;
+interface DayAgenda<T> {
+  reservation?: T;
   date?: XDate;
 }
 
-interface State {
-  reservations: DayAgenda[];
+interface State<T> {
+  reservations: DayAgenda<T>[];
 }
 
-class ReservationList extends Component<ReservationListProps, State> {
+class ReservationList<T = any> extends Component<ReservationListProps<T>, State<T>> {
   static displayName = 'ReservationList';
 
   static propTypes = {
@@ -64,7 +75,7 @@ class ReservationList extends Component<ReservationListProps, State> {
     selectedDay: PropTypes.instanceOf(XDate),
     topDay: PropTypes.instanceOf(XDate),
     onDayChange: PropTypes.func,
-    
+
     showOnlySelectedDayItems: PropTypes.bool,
     renderEmptyData: PropTypes.func,
 
@@ -77,10 +88,11 @@ class ReservationList extends Component<ReservationListProps, State> {
     refreshing: PropTypes.bool,
     onRefresh: PropTypes.func
   };
-  
+
   static defaultProps = {
     refreshing: false,
-    selectedDay: new XDate(true)
+    selectedDay: new XDate(true),
+    keyExtractor: (_item: unknown, index: number) => String(index)
   };
 
   private style: {[key: string]: ViewStyle | TextStyle};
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Inbal-Tish** [comment]
@JacobJaffe Thank you for the PR. Please remove formatting so I can review only the relevant code changes. Also, I'll add the fix for the Agenda keyExtractor and allow passing one, so please focus your PR on the generic type and I'll review it with priority. Thank you again!

**Comment by vlkpa** [comment]
It would be great to merge this. We want to use custom data in entries and this is easy way how to do it.

**Comment by 1ErikaMarques** [comment]
That would be very helpful, I'm planning to add custom data to the items, but it is impossible at the moment.

**Comment by JacobJaffe** [comment]
Hi @Inbal-Tish , I've updated this PR. I rebased it today, and updated it to the changes that have occurred since it originally was posted.

Here's an example of what this will look like, also. In this screenshot, you can see how the inline `renderItem` is correctly inferring the generic `AgendaEntry` defined by `items`

`example/src/screens/agenda.tsx`:
image


As for the formatting changes... this is auto formatted to the `prettier` configuration. I would suggest ripping through the whole repo with a prettier write, and then enforcing prettier at a CI level.

## Want type safety now, before this lands?
Here's how I'm currently doing this. I wrap the `Agenda` in an opinionated abstraction that types for the data type I use in my app: 

Something like this:
```tsx
import { Agenda, AgendaSchedule } from "react-native-calendars";

type CustomAgendaItem = { ... }; // Your data type!

export const CustomAgenda: React.FC<
  {
    items: CustomAgendaItem[];
    onPressItem?: (item: CustomAgendaItem) => void;
  } & 
  // Expose a subset of the Agenda props, whichever ones you need.
  // Or do all of them, but I wouldn't recommend that, since the motivation
  // here is not trusting underlying types, so vetting props is good.
  Pick<
    React.ComponentProps<typeof Agenda>,
    "onCalendarToggled" | "refreshing" | "onRefresh"
  >
> = ({ items, onPressItem, ...agendaProps }) => {
  return (
    <Agenda
      {...agendaProps}
      items={items as unknown as AgendaSchedule}
      renderItem={(item, isFirst) => (
        <AgendaItem
          isFirst={isFirst}
          item={item as unknown as CustomAgendaItem}
          onPress={
            onPressItem
              ? () => onPressItem(item as unknown as CustomAgendaItem)
              : undefined
          }
        />
      )}
    />
  );
};

...

const AgendaItem: React.FC<{
  item: CustomAgendaItem;
  onPress?: () => void;
  isFirst?: boolean;
}> = ({ item, onPress, isFirst }) => {
   ...
};
```

Tobertet:
@Inbal-Tish Can you, or anyone in wix, please review this PR?

rokoala:
@Inbal-Tish, can anyone help on this issue?

rokoala:
> Hi @Inbal-Tish , I've updated this PR. I rebased it today, and updated it to the changes that have occurred since it originally was posted.
> 
> Here's an example of what this will look like, also. In this screenshot, you can see how the inline `renderItem` is correctly inferring the generic `AgendaEntry` defined by `items`
> 
> `example/src/screens/agenda.tsx`: image
> 
> ## Want type safety now, before this lands?
> Here's how I'm currently doing this. I wrap the `Agenda` in an opinionated abstraction that types for the data type I use in my app:
> 
> Something like this:
> 
> ```tsx
> import { Agenda, AgendaSchedule } from "react-native-calendars";
> 
> type CustomAgendaItem = { ... }; // Your data type!
> 
> export const CustomAgenda: React.FC<
>   {
>     items: CustomAgendaItem[];
>     onPressItem?: (item: CustomAgendaItem) => void;
>   } & 
>   // Expose a subset of the Agenda props, whichever ones you need.
>   Pick<
>     React.ComponentProps<typeof Agenda>,
>     "onCalendarToggled" | "refreshing" | "onRefresh"
>   >
> > = ({ items, onPressItem, ...agendaProps }) => {
>   return (
>     <Agenda
>       {...agendaProps}
>       items={items as unknown as AgendaSchedule}
>       renderItem={(item, isFirst) => (
>         <AgendaItem
>           isFirst={isFirst}
>           item={item as unknown as CustomAgendaItem}
>           onPress={
>             onPressItem
>               ? () => onPressItem(item as unknown as CustomAgendaItem)
>               : undefined
>           }
>         />
>       )}
>     />
>   );
> };
> 
> ...
> 
> const AgendaItem: React.FC<{
>   item: CustomAgendaItem;
>   onPress?: () => void;
>   isFirst?: boolean;
> }> = ({ item, onPress, isFirst }) => {
>    ...
> };
> ```

@JacobJaffe, I've just tried this workaround and it seems that the key property "day" is required on "items" (which is transformed into "date") so that the [reservation component](https://github.com/wix/react-native-calendars/blob/18001b00b712de14871a68ecb85b49626b473fd1/src/agenda/reservation-list/reservation.tsx#L52) could update:

```tsx
shouldComponentUpdate(nextProps: ReservationProps) {
    const d1 = this.props.date;
    const d2 = nextProps.date;
    const r1 = this.props.item;
    const r2 = nextProps.item;
    
    let changed = true;
    if (!d1 && !d2) {
      changed = false;
    } else if (d1 && d2) {
      if (d1.getTime() !== d2.getTime()) {
        changed = true;
      } else if (!r1 && !r2) {
        changed = false;
      } else if (r1 && r2) {
        if ((!d1 && !d2) || (d1 && d2)) {
          if (isFunction(this.props.rowHasChanged)) {
            changed = this.props.rowHasChanged(r1, r2);
          }
        }
      }
    }
    return changed;
  } 
  ```

LazyAfternoons:
Any news on this?

thaynarbo:
Is anyone taking a look at it?

ddedic:
+1

PR #1795, opened 2022-02-16
Agenda - fix keyExtractor for reservation list

Inbal-Tish:
Agenda - fix `keyExtractor` for reservation list and allow passing a custom one

### DIFF
--- a/example/src/screens/agenda.tsx
+++ b/example/src/screens/agenda.tsx
@@ -37,6 +37,7 @@ export default class AgendaScreen extends Component<State> {
         // theme={{calendarBackground: 'red', agendaKnobColor: 'green'}}
         //renderDay={(day, item) => (<Text>{day ? day.day: 'item'}</Text>)}
         // hideExtraDays={false}
+        // showOnlySelectedDayItems
       />
     );
   }
--- a/src/agenda/reservation-list/index.tsx
+++ b/src/agenda/reservation-list/index.tsx
@@ -44,6 +44,8 @@ export type ReservationListProps = ReservationProps & {
   refreshing?: boolean;
   /** If provided, a standard RefreshControl will be added for "Pull to Refresh" functionality. Make sure to also set the refreshing prop correctly */
   onRefresh?: () => void;
+  /** Extractor for underlying FlatList. Ensure that this is unique per item, or else scrolling may have duplicated and / or missing items.  */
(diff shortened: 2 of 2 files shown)


lidord-wix (approved):

eladgel:
@lidord-wix @Inbal-Tish this does not fix the issue as the declared keyExtractor in your solution propagates to the CalendarList instead of the agenda. We use the same property for both lists!!

Inbal-Tish:
@eladgel Thank you! I pushed a fix for the prop name

eladgel:
@Inbal-Tish Any ETA on when is **1.12789.0** coming out?

Inbal-Tish:
> @Inbal-Tish Any ETA on when is **1.12789.0** coming out?

We're usually releasing a version every two weeks, so begging of March should be the next version.

prabir-vora:
Can someone post an example of how to use the keyExtractor with Agenda? I am facing the same issue with updating the items in Agenda and incorrect items displaying. I am hoping using keyExtractor fixes it so if someone can guide me on how to use it that would be very helpful.

juspeeh:
I don't understand how keyExtractor should be used either. It's just returning XDate for each month I have set for scrollRange. How is that supposed to help me separate items from each other?

prabir-vora:
@juspeeh If it helps, in my case I went inside the node_modules and in the Reservations List component, I added my own implementation of keyExtractor. I believe the **reservationKeyExtractor** is the prop they expect for your own implementation but it wasn't getting called for me.

juspeeh:
Yep it's not getting called. reservationsKeyExtractor is always undefined even tho you pass it to the Agenda. Noticed that keyExtractor has fallback  ``${item?.reservation?.day}${index}`;` which I'm able to use for the purpose. I just generate random string for every single object to day property. Now it seems to be working. 

@prabir-vora Thanks for reaching back.

Why are these things not documented?

Completely breaks the whole calendar to a state where it cannot be used unless you start debugging the library and find these things yourself.

Inbal-Tish:
Hi all. Bug fixed (will be released on the next version). You can now use `<Agenda ... reservationsKeyExtractor={yourKeyExtractor}/>`

gkasireddy202:
I still getting this issue in the latest version.
Any updates?

PR #482, opened 2018-05-10
Feat on press year

rknell:
Kept command in line with text order for the arrows ie onPressX instead of onXPress

This is useful so I can implement a new component to select the month / year by clicking on it, similar to the workflow of other calendars.

PR #1798, opened 2022-02-21
Add onPressHeader prop on Calendar

Altimar21:
Hello,

This is an update for https://github.com/wix/react-native-calendars/pull/482. 
Add prop onPressHeader on Header component.

This is my first contribution, if I need to correct something say me,

Thanks,


goxr3plus:
...

kiranojhanp:
.

src/calendar/header/index.tsx
```diff
@@ -192,7 +195,18 @@ class CalendarHeader extends Component<CalendarHeaderProps> {
       return customHeaderTitle;
     }
 
-    return (
+    return onPressHeader ? (
+      <TouchableOpacity onPress={() => onPressHeader(month)}>
```
Inbal-Tish, line 199: First, no need to wrap, you can use Text's `onPress` callback. Second, please don't use inline function calls, this is bad for performance

PR #1800, opened 2022-02-21
Migrate to functional components - 1

Inbal-Tish:
private components migration to functional component


src/calendar/day/basic/index.tsx
```diff
@@ -28,214 +26,188 @@ export interface BasicDayProps {
...
+    markingType,
+    marking,
+    state,
+    disableAllTouchEventsForDisabledDays,
+    disableAllTouchEventsForInactiveDays,
+    accessibilityLabel,
+    children,
+    testID
+  } = props;
+  const style = styleConstructor(theme);
```
ethanshar, line 51: Keep style as a ref, otherwise you are constructing it on every render (and creating a new style object ref)

src/calendar/header/index.tsx
```diff
@@ -72,116 +70,101 @@ export interface CalendarHeaderProps {
...
+    disabledDaysIndexes,
+    displayLoadingIndicator,
+    customHeaderTitle,
+    renderHeader,
+    webAriaLevel,
+    testID,
+    accessibilityElementsHidden,
+    importantForAccessibility
+  } = props;
+  const style = styleConstructor(theme);
```
ethanshar, line 98: use ref/state for this, otherwise you creating a new object each render

src/calendar/day/index.tsx
```diff
@@ -74,42 +52,46 @@ export default class Day extends Component<DayProps> {
...
-
-    return `${isToday ? today : ''} ${day.toString(formatAccessibilityLabel)} ${markingLabel}`;
-  });
 
-  getDayComponent() {
-    const {dayComponent, markingType} = this.props;
+    return `${_isToday ? today : ''} ${day?.toString(formatAccessibilityLabel)} ${markingLabel}`;
+  }, [day, marking, _isToday]);
 
+  const Component = useMemo(() => {
```
ethanshar, line 64: No reason memoizing it, cause it just returns the component and not an instance of a component. 
The component is always the same because it is being imported at the top of the file.

src/calendar/header/index.tsx
```diff
@@ -247,64 +229,69 @@ class CalendarHeader extends Component<CalendarHeaderProps> {
...
+  return (
+    <View
+      testID={testID}
+      style={propsStyle}
+      accessible
+      accessibilityRole={'adjustable'}
+      accessibilityActions={[
+        {name: 'increment', label: 'increment'},
+        {name: 'decrement', label: 'decrement'}
+      ]}
```
ethanshar, line 254: Avoid passing static inline object, you can define as a file const

ethanshar (changes requested):
@Inbal-Tish 
Overall the refactor looks good, But I notice a performance degradation in this PR. 

I see that the Basic Day component for example now render each time (all days are being rendered when switching between two days!)

I suggest using [this hook](https://github.com/craig1123/react-recipes/blob/master/docs/useWhyDidYouUpdate.md) to investigate the source of wasteful renders. 

In general from a quick look, I see that we pass BasicDay the `date` prop which seem to be a new object reference each time. 
Any reason why shouldn't we pass a day string? (and generate an object internally with date utils)

Also I notice that the Day component accepts a `day` prop of XDate type. Also here, I think we can use date string instead and internally create an XDate. 

I believe that if we'll avoid passing date objects between components we'll see a huge improvement.

src/calendar/day/basic/index.tsx
Inbal-Tish: Thanks

src/calendar/header/index.tsx
Inbal-Tish: Thanks. Done

Inbal-Tish:
> this hook

Done and done. Good catch, great hook!

ethanshar:
> > this hook
> 
> Done and done. Good catch, great hook!

@Inbal-Tish 
I still see performance issues in the first example screen (`Calendars`) 
It might be related to the example, but when I console the renders. The days are still re-render on every day change

src/calendar/day/index.tsx
```diff
@@ -74,42 +53,37 @@ export default class Day extends Component<DayProps> {
...
-      >
-        {formatNumbers(day?.getDate())}
-      </Component>
-    );
-  }
-}
+    return `${_isToday ? today : ''} ${_day?.toString(formatAccessibilityLabel)} ${markingLabel}`;
+  }, [_day, marking, _isToday]);
+  
+  const dayProps = useMemo(() => {
```
ethanshar, line 65: Why the dayProps has a dependency on the _day object? 
Also, consider destructing props you don't want to pass on at the top of the component.
Something like this
```
const {day, ...others} = props;
```
everything in `others` are stuff you want to pass forward. 
And if you want to use props from `others` internally, just use `others.marking` 

Does that make sense? I think it'll make the code cleaner and we can avoid redundant memoization. 


Actually the `useMemo` here is unnecessary anyway cause you don't create a new ref of anything you just extract the relevant props

src/calendar/header/index.tsx
```diff
@@ -95,7 +95,11 @@ const CalendarHeader = forwardRef((props: CalendarHeaderProps, ref) => {
     accessibilityElementsHidden,
     importantForAccessibility
   } = props;
-  const style = styleConstructor(theme);
+  const style = useRef(styleConstructor(theme));
+  const accessibilityActions = useRef([
```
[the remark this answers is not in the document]
Inbal-Tish: Done

Inbal-Tish:
> > > this hook
> > 
> > 
> > Done and done. Good catch, great hook!
> 
> @Inbal-Tish I still see performance issues in the first example screen (`Calendars`) It might be related to the example, but when I console the renders. The days are still re-render on every day change

Which render do you console?

src/calendar/day/index.tsx
Inbal-Tish: The `omit` creates a new object so I'm returning a new `dayProps` object every time. In this case, I need to pass all the props except `day`, so I don't see how using `...others` will assist here. Anyways, I'm removing the 'day' prop in favor of 'date' so I can now pass the props directly.

Inbal-Tish:
> > Which render do you console? If you're checking Day component then passing `onPress` is causing a render. I'm working on moving Calendar to functional as well so I can improve this as well. Do you think it's a regression?
> 
> Im adding console to the day component. I thought that we fixed that yesterday, didn't we?

Yes. Fix was pushed

src/calendar/day/index.tsx
```diff
@@ -12,21 +11,17 @@ import BasicDay, {BasicDayProps} from './basic';
...
   dayComponent?: React.ComponentType<DayProps & {date?: DateData}>;
 }
 
 const Day = (props: DayProps) => {
-  const {day, marking, dayComponent, markingType} = props;
-  const _day = day ? new XDate(day) : undefined;
-  const _isToday = _day ? isToday(_day) : undefined;
+  const {date, marking, dayComponent, markingType} = props;
+  const _date = date ? new XDate(date) : undefined;
+  const _isToday = _date ? isToday(_date) : undefined;
```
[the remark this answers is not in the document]
Inbal-Tish: Yes. I already do. fixing here

ethanshar (approved):

PR #1811, opened 2022-03-02
Migrate to functional components - 2

Inbal-Tish:
PeriodDay, Calendar


src/calendar/day/period/index.tsx
```diff
@@ -1,247 +1,197 @@
...
+  startingDay?: ViewStyle;
+  endingDay?: ViewStyle;
+  day?: ViewStyle;
+}
 
-    return shouldUpdate(this.props, nextProps, ['children', 'state', 'marking', 'onPress', 'onLongPress', 'date']);
-  }
+const PeriodDay = (props: PeriodDayProps) => {
+  const {theme, marking, date, onPress, onLongPress, state, accessibilityLabel, testID, children} = props;
+  const dateData = date ? xdateToData(new XDate(date)) : undefined;
```
ethanshar, line 34: I noticed this a few times, where you check if a value is undefined before passing it to the a util function. 
Consider moving this logic to the util and return undefined if the argument passed is undefined. 
It's very similar to how lodash works (they know how to handle undefined parameters) 
Also, it makes the code a little cleaner

src/calendar/index.tsx
```diff
@@ -72,143 +72,102 @@ export interface CalendarProps extends CalendarHeaderProps, DayProps {
...
 
-  onSwipeRight = () => {
-    this.header?.current?.onPressLeft();
+  const onSwipeRight = () => {
+    // @ts-expect-error
+    header.current?.onPressLeft();
   };
 
-  renderWeekNumber = memoize(weekNumber => {
+  const renderWeekNumber = memoize(weekNumber => {
```
ethanshar, line 163: what's the purpose of memoization here? 
And why not use `useMemo`?

src/calendar/index.tsx
```diff
@@ -72,143 +72,102 @@ export interface CalendarProps extends CalendarHeaderProps, DayProps {
...
+  }, [onDayPress]);
 
-  swipeProps = {onSwipe: (direction: string) => this.onSwipe(direction)};
+  const longPressDay = useCallback((date?: DateData) => {
+    if (date)
+    handleDayInteraction(date, onDayLongPress);
+  }, [onDayLongPress]);
 
-  onSwipe = (gestureName: string) => {
+  const onSwipe = (gestureName: string) => {
```
ethanshar, line 137: Wrap with useCallback

ethanshar:
@Inbal-Tish 
One more thing I just noticed, In the HorizontalCalendarList example there are performance issues now. 
All days are being rendered again, it doesn't reproduce on master. 
I suspect it might related to the other comments I wrote.

src/calendar/day/period/index.tsx
Inbal-Tish: Changing the util to return undefined means to change the API for callback props that return this object. We can do it in V2

example/src/screens/expandableCalendar.tsx
```diff
@@ -181,7 +181,7 @@ export default class ExpandableCalendarScreen extends Component<Props> {
   render() {
     return (
       <CalendarProvider
-        date={ITEMS[0].title}
+        date={ITEMS[1].title}
```
CezarCobuz, line 184: Isn't this subject to undefined if there is nothing inside ITEMS?

Inbal-Tish: ITEMS is a const defined in the example screen. It's not dynamic data.

CezarCobuz (approved):

Inbal-Tish:
@ethanshar I would like to release this one on this cycle so we can have a full QA before we continue with other components since these are the basics that all the others rely on. Are you waiting for something on my end?

src/calendar/index.tsx
```diff
@@ -72,143 +72,102 @@ export interface CalendarProps extends CalendarHeaderProps, DayProps {
...
   };
 
-  handleDayInteraction(date: DateData, interaction?: (date: DateData) => void) {
-    const {disableMonthChange, allowSelectionOutOfRange} = this.props;
+  const addMonth = useCallback((count: number) => {
+    const newMonth = currentMonth.clone().addMonths(count, true);
+    updateMonth(newMonth);
+  }, [currentMonth, updateMonth]);
+
+  const handleDayInteraction = (date: DateData, interaction?: (date: DateData) => void) => {
```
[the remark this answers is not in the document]
ethanshar: @Inbal-Tish What about this one?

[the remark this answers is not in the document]
Inbal-Tish: Missed it. Done

ethanshar (approved):

PR #1820, opened 2022-03-08
NEW CalendarList

Inbal-Tish:
CalendarList - new implementation using InfiniteList component.
InfiniteList - adding support for vertical, removing horizontal as the default.
WeekCalendar (new) and TimelineList - add 'isHorizontal' prop.


src/calendar-list/new.tsx
```diff
@@ -0,0 +1,196 @@
...
+        onPressArrowRight={scrollToNextMonth}
+        onPressArrowLeft={scrollToPreviousMonth} 
+        hideExtraDays={calendarProps?.hideExtraDays || true}
+        style={[style.current.calendar, calendarProps?.style]}
+        headerStyle={horizontal ? calendarProps?.headerStyle : undefined}
+        testID={`${testID}_${item}`}
+        // context={context}
+      />
+    );
+  }, [calendarProps, items]);
```
ethanshar, line 147: Technically shouldn't you depend on `scrollToNextMonth` and `scrollToPreviousMonth` that transitively depends on items. 
I think it's easier to understand the code when you see a function depends on something that it uses

src/calendar-list/item.tsx
```diff
@@ -76,7 +76,7 @@ class CalendarListItem extends Component<CalendarListItemProps, CalendarListItem
     }
   };
 
-  onPressArrowRight = (_: any, month: any) => {
+  onPressArrowRight = (_: () => void, month: XDate) => {
```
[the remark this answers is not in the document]
Inbal-Tish: `_` is a way to mute lint error for param that is declared but not used. It's a convention.

example/src/screens/horizontalCalendarListScreen.tsx
```diff
@@ -0,0 +1,42 @@
+import React, {useMemo, useState, useCallback} from 'react';
+// @ts-expect-error
```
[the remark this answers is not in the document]
Inbal-Tish: `Cannot find module 'react-native-calendars' or its corresponding type declarations.ts(2307)`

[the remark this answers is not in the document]
Inbal-Tish: Strange, I get this error just for CalndarList and HorizontalCalendarList screens but their export is the same as all the others...

src/calendar-list/new.tsx
Inbal-Tish: Correct. Fixed

example/src/screens/horizontalCalendarListScreen.tsx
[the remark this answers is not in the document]
Inbal-Tish: I removed it anyways since it doesn't throw ts error

Inbal-Tish:
@ethanshar I fixed vertical updates issue, please take a look.

Inbal-Tish:
@ethanshar Is there an issue with this PR?

src/infinite-list/index.tsx
```diff
@@ -101,15 +117,18 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
...
   const onMomentumScrollEnd = useCallback(
     event => {
-      if (isOnEdge.current) {
-        onReachEdge?.(pageIndex.current!);
-        reloadPagesDebounce?.(pageIndex.current);
-      } else if (isNearEdge.current) {
-        reloadPagesDebounce?.(pageIndex.current);
-        onReachNearEdge?.(pageIndex.current!);
+      console.warn('onMomentumScrollEnd: ', pageIndex.current);
+      if (pageIndex.current) {
```
ethanshar, line 121: What's the purpose of checking `if (pageIndex.current)` here?

src/infinite-list/index.tsx
```diff
@@ -39,8 +41,10 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
     onReachNearEdgeThreshold,
     initialPageIndex = 0,
     extendedState,
-    scrollViewProps
+    scrollViewProps,
+    positionIndex = 0
```
ethanshar, line 45: What's the purpose of `positionIndex`? 
Is that like `initialPageIndex`?

src/calendar-list/index.tsx
```diff
@@ -230,10 +230,8 @@ class CalendarList extends Component<CalendarListProps, State> {
   }
 
   addMonth = (count: number) => {
-    this.updateMonth(this.state.currentMonth.clone().addMonths(count, true));
-  };
-
-  updateMonth(day: XDate) {
+    const day = this.state.currentMonth.clone().addMonths(count, true);
```
ethanshar, line 233: Something that occur to me that we might have missed. 
There are many ways to clone an object and it's possible the way we clone is expensive especially if we're doing it in many places (which I know we do in Calendars) 

There are many questions about it, for [example](https://stackoverflow.com/questions/122102/what-is-the-most-efficient-way-to-deep-clone-an-object-in-javascript)
And someone mention this benchmark testing -> http://jsben.ch/bWfk9
And it looks like cloning using `object.assign` is best option. 
Maybe we should consider handling that as well (as a different task of course)

src/calendar-list/item.tsx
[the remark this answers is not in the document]
Inbal-Tish: Done

src/infinite-list/index.tsx
```diff
@@ -101,15 +117,18 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
 
   const onMomentumScrollEnd = useCallback(
     event => {
-      if (isOnEdge.current) {
-        onReachEdge?.(pageIndex.current!);
-        reloadPagesDebounce?.(pageIndex.current);
-      } else if (isNearEdge.current) {
-        reloadPagesDebounce?.(pageIndex.current);
-        onReachNearEdge?.(pageIndex.current!);
+      console.warn('onMomentumScrollEnd: ', pageIndex.current);
```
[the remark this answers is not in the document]
Inbal-Tish: Done

src/infinite-list/index.tsx
Inbal-Tish: Not to trigger callbacks with undefined param

src/infinite-list/index.tsx
Inbal-Tish: It's like the index of the item where we need to append the data when updating with new items.

src/calendar-list/index.tsx
Inbal-Tish: I agree. We have it alot in Calendars and we should replace it with the best option. I can open a ticket for it

src/infinite-list/index.tsx
ethanshar: But should pageIndex be undefined at all? 
I see we update its value in `onScroll`, how do we get to `onMomentumScrollEnd` without a scroll happening? 
Moreover, should we maybe initialize the `pageIndex` ref with a value to avoid an undefined value?

src/infinite-list/index.tsx
ethanshar: The name is a little vague IMO, can you think of another name?

ethanshar (approved):

src/infinite-list/index.tsx
Inbal-Tish: `dataAppendIndex` is more descriptive I guess. Done

src/infinite-list/index.tsx
Inbal-Tish: There is an issue with `onMomentumScrollEnd` on Android (not being called) and keeping the `pageIndex` undefined in the init was to avoid some code in the `onScroll` but still execute the rest. I don't want to change it unless we recheck the whole scenario, but we can do it later.

src/infinite-list/index.tsx
Inbal-Tish: Sorry. The change is here: https://github.com/wix/react-native-calendars/pull/1851

PR #1822, opened 2022-03-09
 Fix - Add reservationsKeyExtractor function to proptypes

jsp3536:
reservationsKeyExtractor  function is missing from propTypes event though it is available in the ReservationListProps  type.


Inbal-Tish:
@jsp3536 Thanks for the PR. This issue was already fixed.

PR #1825, opened 2022-03-09
WeekDaysNames component

Inbal-Tish:
Move WeekDaysNames component to commons dir and use it in other components


src/agenda/style.ts
```diff
@@ -34,13 +34,13 @@ export default function styleConstructor(theme: Theme = {}) {
       alignItems: 'center',
       backgroundColor: appStyle.calendarBackground
     },
-    weekday: {
+    dayHeader: {
```
ethanshar, line 37: Shouldn't we call it `daysHeader`? 
btw here and in other places.. 

```suggestion
    daysHeader: {
```

src/calendar/header/index.tsx
```diff
@@ -235,11 +235,15 @@ const CalendarHeader = forwardRef((props: CalendarHeaderProps, ref) => {
...
+    return showWeekNumbers && <View style={style.current.dayHeader}/>;
+  };
+
   const renderDayNames = () => {
     if (!hideDayNames) {
       return (
         <View style={style.current.week} testID={testID ? `${HEADER_DAY_NAMES}-${testID}` : HEADER_DAY_NAMES}>
-          {showWeekNumbers && <Text allowFontScaling={false} style={style.current.dayHeader}></Text>}
+          {renderWeekNumbersSpace()}
           {renderWeekDays}
```
ethanshar, line 247: Why didn't you migrate here to WeekDaysNames component as well?

src/calendar/header/index.tsx
```diff
@@ -235,11 +235,15 @@ const CalendarHeader = forwardRef((props: CalendarHeaderProps, ref) => {
...
+  const renderWeekNumbersSpace = () => {
+    return showWeekNumbers && <View style={style.current.dayHeader}/>;
+  };
+
   const renderDayNames = () => {
     if (!hideDayNames) {
       return (
         <View style={style.current.week} testID={testID ? `${HEADER_DAY_NAMES}-${testID}` : HEADER_DAY_NAMES}>
-          {showWeekNumbers && <Text allowFontScaling={false} style={style.current.dayHeader}></Text>}
+          {renderWeekNumbersSpace()}
```
ethanshar, line 246: What's the purpose of the Space render here?

src/agenda/style.ts
Inbal-Tish: It's a style that repeats on other components for the same part. I would like to keep it this way and not change on all files. The aim is to make it the default style and delete it eventually.

src/calendar/header/index.tsx
Inbal-Tish: It is an overlay for the vertical calendar. When you have the week numbers, the calendar offset a day and this means adding the same offset to the weekdays names

src/calendar/header/index.tsx
Inbal-Tish: It is an overlay for the vertical calendar. When you have the week numbers, the calendar offset a day and this means adding the same offset to the weekdays names

src/calendar/header/index.tsx
Inbal-Tish: This header has more functionality, mainly the 'dayTextAtIndex', passing this functionality to WeekDaysNames component will complicate the code as it is adding styles to the header/style StyleSheet

ethanshar: IMO it makes more sense to make the WeekDaysNames more generic and maybe a little more complex but support all use cases accordingly. 
Anyway, it's up to you, we can also migrate this use case later on..

ethanshar (approved):

src/calendar/header/index.tsx
Inbal-Tish: That was my first thought, and I tried it, but then it became very complex and hard to follow that I decided to leave it for now. We can revisit it in the future.

PR #1839, opened 2022-03-27
Infra/expandable calendar to function

Inbal-Tish:


src/expandableCalendar/Context/Provider.tsx
```diff
@@ -34,158 +43,132 @@ interface Props {
...
-    return (
-      <CalendarContext.Provider value={this.getProviderContextValue()}>
-        <View style={[styles.container, this.props.style]}>{this.props.children}</View>
-        {this.props.showTodayButton && this.renderTodayButton()}
-      </CalendarContext.Provider>
-    );
-  }
-}
+  return (
+    <CalendarContext.Provider value={getProviderContextValue()}>
```
ethanshar, line 156: By invoking `getProviderContextValue` each render you create a new object and a new ref

src/expandableCalendar/Context/Provider.tsx
```diff
@@ -34,158 +43,132 @@ interface Props {
...
+      <View style={[style.current.contextWrapper, propsStyle]}>{children}</View>
```
ethanshar, line 157: consider memoizing this `[style.current.contextWrapper, propsStyle]`

ethanshar, line 157: consider memoizing this `[style.current.contextWrapper, propsStyle]`

src/expandableCalendar/index.tsx
```diff
@@ -91,515 +96,491 @@ interface State {
...
+        <Calendar
+          testID="calendar"
+          {...others}
+          theme={themeObject}
+          onDayPress={_onDayPress}
+          hideExtraDays
+          renderArrow={_renderArrow}
+        />
+      ) : (
+        <Animated.View ref={wrapper} style={{height: deltaY.current}} {...panResponder.panHandlers}>
```
ethanshar, line 527: consider memoizing even that it depends on a ref object. 
The object here always recreated

ethanshar, line 527: consider memoizing even that it depends on a ref object. 
The object here always recreated

ethanshar (commented):
Overall look good, wrote a few comments. 
In general, I see there not a few places where you pass style inline and not memoizing, I think the thumb rule of this should be to use memoize for components that are wrappers for other components cause then a re-render of parent will cause a render of the whole tree under it. 

Another thing, I'd consider breaking big components like expandableCalendar (~550 lines) to smaller parts, using hooks for example. It's really difficult to go over the whole component - you can do it on another PR, not necessarily this one.

src/expandableCalendar/Context/Provider.tsx
Inbal-Tish: Moved to useMemo

src/expandableCalendar/Context/Provider.tsx
Inbal-Tish: moved to useMemo

Inbal-Tish: duplication I'm guessing...

src/expandableCalendar/index.tsx
Inbal-Tish: Moved all array styles to useMemo

src/expandableCalendar/Context/Provider.tsx
```diff
@@ -34,158 +43,136 @@ interface Props {
...
+      animateOpacity(disabled);
     };
 
-    setDisabled(showTodayButton, disabled, this.state.disabled, updateState);
+    setDisabled(showTodayButton, disabled, isDisabled, updateState);
   };
 
-  animateTodayButton(date: string) {
-    const {shouldAnimateTodayButton, getPositionAnimation} = this.presenter;
+  const getProviderContextValue = useMemo(() => {
```
ethanshar, line 106: since the const is not a method anymore, I suggest renaming it

ethanshar (approved):
Approved, wrote a small comment on the provider value const name 
Consider fixing it before merging

Inbal-Tish:
> getProviderContextValue

Done!

PR #1843, opened 2022-03-29
Export timeline-list types using type export

jdpnielsen:
This is another followup to https://github.com/wix/react-native-calendars/pull/1745 and https://github.com/wix/react-native-calendars/pull/1752 solving type exports.


Inbal-Tish:
@jdpnielsen Hi. Thank you for the PR, I just merged this same one... Will release the fix on the next version. Thanks again!!!

PR #1845, opened 2022-03-30
types export fix

yairopro:
Just adding the `type` after the `export` makes it compliant for  webpack's `ts-loader`.


Inbal-Tish (approved):

yairopro:
Thanks 😁

PR #1848, opened 2022-03-31
componentUpdater.ts uses propTypesMap instead of Component.propTypes

yairopro:
The purpose of this PR is to be able to use this library on the web.
If you try the library locally or on a  CodeSandbox, it works as a charm.
The problem is when CRA (`create-react-app`) builds a release. In its build process, CRA remove the components' `propTypes` for production.
Since the library uses `propTypes` to know what props to pass to lower components, all props can't pass down and it breaks in production (on web).

This PR comes to allow `componentUpdater.ts` to get components' `propTypes` by another way using a shared `Map` called `propTypesMap`. Each component set its own `propTypes` in the `propTypesMap`.

This pattern prevent CRA to remove `propTypes`.


yairopro:
Made the PR on the `release` branch
https://github.com/wix/react-native-calendars/pull/1849

yairopro:
Updated to master

white-room:
What needs to happen to get this merged in? Would love to use this as a universal component!

CheepyD:
When it will be merged in?

dohomi:
any chance this getting merged?

TNAJanssen:
Why was this never merged?

Issue #1733, opened 2021-12-27
Does this library work with the Web version ？


pxim:
## Description

Does this library work with the Web version ？

yairopro:
Almost.
Since react-scripts removes the `propTypes` fields from third party components in the production build, it cannot be shipped though it works locally  like a charm.

sirshad01:
Any chance web support is coming soon?

yairopro:
No stale

yairopro:
No stale

TNAJanssen:
This is not completely finished and should not be stale. There are still issues on the web version.

PR #1849, opened 2022-03-31
componentUpdater.ts uses propTypesMap instead of Component.propTypes

yairopro:
The purpose of this PR is to be able to use this library on the web.
If you try the library locally or on a  CodeSandbox, it works as a charm.
The problem is when CRA (`create-react-app`) builds a release. In its build process, CRA remove the components' `propTypes` for production.
Since the library uses `propTypes` to know what props to pass to lower components, all props can't pass down and it breaks in production (on web).

This PR comes to allow `componentUpdater.ts` to get components' `propTypes` by another way using a shared `Map` called `propTypesMap`. Each component set its own `propTypes` in the `propTypesMap`.

This pattern prevent CRA to remove `propTypes`.

### DIFF
--- a/src/calendar-list/index.tsx
+++ b/src/calendar-list/index.tsx
@@ -14,6 +14,7 @@ import Calendar, {CalendarProps} from '../calendar';
 import CalendarListItem from './item';
 import CalendarHeader from '../calendar/header/index';
 import constants from '../commons/constants';
+import propTypesMap from '../propTypesMap';
 
 
 const CALENDAR_WIDTH = constants.screenWidth;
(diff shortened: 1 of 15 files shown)


yairopro:
Will solve #1733

yairopro:
This PR depends on the #1845

yairopro:
Do you prefer the PR on the master or the release branch.

Inbal-Tish:
> Do you prefer the PR on the master or the release branch?

Only on master branch please

yairopro:
So I close this one and I reopen the previous #1848

PR #1851, opened 2022-03-31
Feat/new calendar list header props

Inbal-Tish:
Depends on https://github.com/wix/react-native-calendars/pull/1820 - don't merge before it is merged!!!
Adding 'headerProps' prop.


ethanshar (approved):
@Inbal-Tish 
I didn't find anything problematic but maybe it worth checking the changes in Owners app after you merge, just in case.

PR #1867, opened 2022-04-13
deps: remove redundant fs

BethGriggs:
fs is a core Node.js module - at the moment this is installing https://www.npmjs.com/package/fs which is a security holding package with no content

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
   "author": "Wix.com",
   "license": "MIT",
   "dependencies": {
-    "fs": "^0.0.1-security",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -31,7 +31,6 @@
   "author": "Wix.com",
   "license": "MIT",
   "dependencies": {
-    "fs": "^0.0.1-security",
```
Inbal-Tish, line 34: This dep is being used. Why do you wish to remove it?

Inbal-Tish (approved):

PR #1885, opened 2022-05-16
Feat/ timeline x days view

lidord-wix:
Timeline x days view
To see the changes, please uncomment line # 242 in the timelineCalendarScreen (and you can change it to any other number from 1 to 7) 
And sorry about this big PR 🙈 
WOAUILIB-2613

### DIFF
--- a/src/expandableCalendar/Context/Provider.tsx
+++ b/src/expandableCalendar/Context/Provider.tsx
@@ -45,7 +45,7 @@ export interface CalendarContextProviderProps extends ViewProps {
   /** The number of days to present in the timeline calendar */
   numberOfDays?: number;
   /** The left inset of the timeline calendar (sidebar width), default is 72 */
-  leftInset?: number;
+  timelineLeftInset?: number;
 }
 
 /**
@@ -61,7 +61,7 @@ const CalendarProvider = (props: CalendarContextProviderProps) => {
     todayButtonStyle,
     style: propsStyle,
     numberOfDays,
-    leftInset = 72,
+    timelineLeftInset = 72,
     children
   } = props;
   const style = useRef(styleConstructor(theme));
@@ -117,7 +117,7 @@ const CalendarProvider = (props: CalendarContextProviderProps) => {
       setDate: _setDate,
       setDisabled: _setDisabled,
       numberOfDays,
-      leftInset
+      timelineLeftInset
     };
   }, [currentDate, prevDate, updateSource]);
 
@@ -184,5 +184,5 @@ CalendarProvider.propTypes = {
   todayButtonStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array]),
   disabledOpacity: PropTypes.number,
   numberOfDays: PropTypes.number,
-  leftInset: PropTypes.number
+  timelineLeftInset: PropTypes.number
 };
--- a/src/expandableCalendar/Context/Provider.tsx
+++ b/src/expandableCalendar/Context/Provider.tsx
@@ -42,6 +42,8 @@ export interface CalendarContextProviderProps extends ViewProps {
   todayButtonStyle?: ViewStyle;
   /** The opacity for the disabled today button (0-1) */
   disabledOpacity?: number;
+  /** The number of days to present in the timeline calendar */
+  numberOfDays?: number;
 }
 
 /**
@@ -56,6 +58,7 @@ const CalendarProvider = (props: CalendarContextProviderProps) => {
     todayBottomMargin,
     todayButtonStyle,
     style: propsStyle,
+    numberOfDays,
     children
   } = props;
   const style = useRef(styleConstructor(theme));
@@ -109,7 +112,8 @@ const CalendarProvider = (props: CalendarContextProviderProps) => {
       prevDate: prevDate,
       updateSource: updateSource,
       setDate: _setDate,
-      setDisabled: _setDisabled
+      setDisabled: _setDisabled,
+      numberOfDays
     };
   }, [currentDate, prevDate, updateSource]);
 
@@ -174,5 +178,6 @@ CalendarProvider.propTypes = {
   showTodayButton: PropTypes.bool,
   todayBottomMargin: PropTypes.number,
   todayButtonStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array]),
-  disabledOpacity: PropTypes.number
+  disabledOpacity: PropTypes.number,
+  numberOfDays: PropTypes.number
 };
--- a/src/expandableCalendar/Context/Provider.tsx
+++ b/src/expandableCalendar/Context/Provider.tsx
@@ -119,7 +119,7 @@ const CalendarProvider = (props: CalendarContextProviderProps) => {
       numberOfDays,
       timelineLeftInset
     };
-  }, [currentDate, prevDate, updateSource]);
+  }, [currentDate, prevDate, updateSource, numberOfDays]);
 
   const animateTodayButton = (date: string) => {
     if (shouldAnimateTodayButton(props)) {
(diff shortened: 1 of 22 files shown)


src/timeline/Timeline.tsx
```diff
@@ -132,24 +148,30 @@ const Timeline = (props: TimelineProps) => {
...
+    } else if (scrollToFirst && getEvents().length > 0) {
+      initialPosition = min(map(getEvents(), 'top')) ?? 0;
     } else if (initialTime) {
       initialPosition = calcTimeOffset(HOUR_BLOCK_HEIGHT, initialTime.hour, initialTime.minutes);
     }
 
     if (initialPosition) {
       setTimeout(() => {
-        scrollView?.current?.scrollTo({
+        (Number(numberOfDays) <= 1) && scrollView?.current?.scrollTo({
```
ethanshar, line 174: What's the purpose of converting `numberOfDays` into a number, it's already a number.. 

Also why not add this condition to the `if (initialPosition) {` check?

src/timeline/Timeline.tsx
```diff
@@ -159,19 +181,19 @@ const Timeline = (props: TimelineProps) => {
 
   const _onEventPress = useCallback(
     (eventIndex: number) => {
-      const event = packedEvents[eventIndex];
+      const event = getEvents()[eventIndex];
```
ethanshar, line 184: See my previous comment on re-calculating this data by calling `getEvents`

src/timeline/Timeline.tsx
```diff
@@ -186,12 +208,22 @@ const Timeline = (props: TimelineProps) => {
     });
 
     return (
-      <View>
-        <View style={{marginLeft: HOURS_SIDEBAR_WIDTH}}>{events}</View>
+      <View style={{marginLeft: index === 0 ? HOURS_SIDEBAR_WIDTH : undefined, width: width / numberOfDays}}>
```
ethanshar, line 211: Have you considered passing `flex: 1` instead of the fixed width. 
It's way more dynamic and probably will work better with orientation change

src/timeline/Timeline.tsx
```diff
@@ -107,7 +121,8 @@ const Timeline = (props: TimelineProps) => {
     start = 0,
     end = 24,
     date,
-    events = [],
+    pageEvents = [[], [], [], [], [], [], []],
+    pageDates = [],
```
ethanshar, line 125: Notice that we already have the `date` prop, maybe we should unify it with `pageDates` 
Meaning, we can change `date` type to be either `string` or `string[]` and it can serve two purposed. 
I think it's better than having two different props with a similar purpose

Another option is use `date` and `numberOfDays` to calculate the dates array yourself.

Reducing props can make the API more clear and prune to misuse

src/timeline-list/useTimelinePages.ts
```diff
@@ -7,18 +7,22 @@ import debounce from 'lodash/debounce';
...
 interface UseTimelinePagesProps {
   date: string;
   listRef: RefObject<any>;
+  numberOfDays: number;
 }
 
-const UseTimelinePages = ({date, listRef}: UseTimelinePagesProps) => {
-  const pagesRef = useRef(times(PAGES_COUNT, i => generateDay(date, i - Math.floor(PAGES_COUNT / 2))));
+const UseTimelinePages = ({date, listRef, numberOfDays}: UseTimelinePagesProps) => {
+  const initialPage = Math.ceil(PAGES_COUNT / 2);
```
ethanshar, line 20: What's the purpose of moving `initialPage` here? 
I see you now calculate it in other files as well..

src/expandableCalendar/style.ts
```diff
@@ -2,6 +2,7 @@ import {StyleSheet, Platform} from 'react-native';
 import * as defaultStyle from '../style';
 import {Theme} from '../types';
 import constants from '../commons/constants';
+import {HOURS_SIDEBAR_WIDTH} from '../timeline/style';
```
ethanshar, line 5: I find it a little weird that ExpandableCalendar depends on the Timeline style. 
Is there any other way around it? Why does it need to know the hours sidebar width?

lidord-wix: Hmm, the header (dates and day names) should be flexed above the timeline with a left inset (we don't want it to be full width).. 
We can also pass this inset as a prop if you think it's better

src/timeline/Timeline.tsx
```diff
@@ -158,40 +178,51 @@ const Timeline = (props: TimelineProps) => {
...
+          onPress={onEventPress}
           renderEvent={renderEvent}
         />
       );
     });
 
     return (
-      <View>
-        <View style={{marginLeft: HOURS_SIDEBAR_WIDTH}}>{events}</View>
+      <View pointerEvents={'none'} style={{marginLeft: dayIndex === 0 ? HOURS_SIDEBAR_WIDTH : undefined, flex: 1}}>
```
ethanshar, line 210: Move static styles to stylesheet

src/expandableCalendar/style.ts
ethanshar: I think it's more correct to pass it as a prop

ethanshar (commented):
See my new comments 
In general notice where you create data that can affect the renders 
See which data you need to memoize and which can be kept as ref

src/timeline/Timeline.tsx
```diff
@@ -119,35 +127,47 @@ const Timeline = (props: TimelineProps) => {
...
-    } else if (scrollToFirst && packedEvents.length > 0) {
-      initialPosition = min(map(packedEvents, 'top')) ?? 0;
+    } else if (scrollToFirst && packedEvents[0].length > 0) {
+      initialPosition = min(map(packedEvents[0], 'top')) ?? 0;
     } else if (initialTime) {
       initialPosition = calcTimeOffset(HOUR_BLOCK_HEIGHT, initialTime.hour, initialTime.minutes);
     }
 
-    if (initialPosition) {
+    if (initialPosition && numberOfDays <= 1) {
```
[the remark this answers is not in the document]
lidord-wix: I reverted the condition change

src/timeline/Timeline.tsx
```diff
@@ -158,40 +178,51 @@ const Timeline = (props: TimelineProps) => {
   }, []);
 
   const _onEventPress = useCallback(
```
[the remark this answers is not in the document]
lidord-wix: I think it's a leftover, I removed it

src/expandableCalendar/Context/Provider.tsx
```diff
@@ -44,6 +44,8 @@ export interface CalendarContextProviderProps extends ViewProps {
   disabledOpacity?: number;
   /** The number of days to present in the timeline calendar */
   numberOfDays?: number;
+  /** The left inset of the timeline calendar (sidebar width), default is 72 */
+  leftInset?: number;
```
ethanshar, line 48: Does it make sense to call this prop `timelineLeftInset` or `timelineSidebarWidth` - this way the prop name also hints it's only relevant to Timeline. `leftInset` feels to general

src/expandableCalendar/index.tsx
```diff
@@ -98,7 +98,7 @@ const headerStyleOverride = {
  */
 
 const ExpandableCalendar = (props: ExpandableCalendarProps) => {
-  const {date, setDate, numberOfDays = 1} = useContext(Context);  
+  const {date, setDate, numberOfDays = 1, leftInset = 72} = useContext(Context);
```
ethanshar, line 101: We're setting the `72` default values in several places. 
if we'll need to change it in the future we might miss one instance and have a bug. 
Can we maybe set its default once in the context value?

src/expandableCalendar/Context/Provider.tsx
lidord-wix: Changed to `timelineLeftInset`

src/expandableCalendar/index.tsx
lidord-wix: Done :)

ethanshar (approved):

Inbal-Tish:
@ethanshar @lidord-wix Don't you want to merge this one?

lidord-wix:
> @ethanshar @lidord-wix Don't you want to merge this one?

@Inbal-Tish no, it requires QA session

PR #1887, opened 2022-05-20
Feat/ Support Ref methods in Calendar Component 

simone989:
In order to call some methods by ref, I implemented the `forwardRef` and `useImperativeHandle` hook inside Calendar Component.

### DIFF
--- a/src/calendar/index.tsx
+++ b/src/calendar/index.tsx
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import XDate from 'xdate';
 
-import React, {useRef, useState, useEffect, useCallback} from 'react';
+import React, {useRef, useState, useEffect, useCallback, useImperativeHandle, forwardRef} from 'react';
 import {View, ViewStyle, StyleProp} from 'react-native';
 // @ts-expect-error
 import GestureRecognizer, {swipeDirections} from 'react-native-swipe-gestures';
@@ -19,7 +19,6 @@ import CalendarHeader, {CalendarHeaderProps} from './header';
 import Day, {DayProps} from './day/index';
 import BasicDay from './day/basic';
 import {MarkingProps} from './day/marking';
-
 
 type MarkedDatesType = {
   [key: string]: MarkingProps;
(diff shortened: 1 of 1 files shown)


dimitrisafendras (approved):

Inbal-Tish:
@simone989 Hi. Thank you for the PR. Can you please remove formatting so I can see only the relevant code changes?

simone989:
> @simone989 Hi. Thank you for the PR. Can you please remove formatting so I can see only the relevant code changes?

Done

src/calendar/index.tsx
```diff
@@ -72,18 +71,26 @@ export interface CalendarProps extends CalendarHeaderProps, DayProps {
   allowSelectionOutOfRange?: boolean;
 }
 
+export type CalendarRef = {
+  addMonth: (count: number) => void;
+  updateMonth: (newMonth: XDate) => void;
+  onSwipeLeft: () => void;
+  onSwipeRight: () => void;
+  onPressDay: (date?: DateData) => void;
```
Inbal-Tish, line 79: Why do you need to invoke these methods? What are the use cases that are missing API?

Inbal-Tish:
@simone989 Hi. I need to understand why you're using the component's ref. What are the use cases that are missing API?

src/calendar/index.tsx
Inbal-Tish: How the user is switching a month? Not with the header's arrows or a calendar swipe?
The methods were available by ref, which is a risky and not a good practice. If you need to expose callbacks for the onSwipe action we can do that, but I still don't understand why you need the `onPressDay` methods when you have the `onDayPress` callbacks...

simone989: I only use the `addMonth` method, I exported the other methods in order to make backward compatible with the previous version.

I pushed a new commit in order to export only the `addMonth` method.

Inbal-Tish: I'm not kine on adding the forward ref, but I want to find the proper solution to your issue. For that I need to know your scenario, how does the user changes the month? Why aren't you using the `initialDate` prop to set the new month to the calendar?

simone989: My scenario is the following: 
I have a custom header with two icons: one icon to switch in the previous month and another icon to switch to next month. When the user press one of the two icons the day selected didn't change but changes only the visible month.

> Why aren't you using the initialDate prop to set the new month to the calendar?

Simple, the name of the prop and the comment in the README.md didn't tell me I can use the variable for this purpose.
Additionally, I don't think is the correct approach because, if I use `initialDate` i should create another state variable in my class/function just for the purpose to update the `currentMonth` calendar internal State. 
For this reason, I prefer to export the internal methods `addMonth` because manipulate directly the internal state.

Inbal-Tish: Ok. For your scenario, are your two icons to replace the arrow icons or they are not placed in the Calendar's header at all? If so, you should implement `renderArrow` prop to pass your Image component and implement `onPressArrowRight` and `onPressArrowLeft` to move the months. These callbacks return two params: the `addMonth` function to be invoked in your callback and the `month` which is the current month. See the code for ExpandableCalendar.index it does all that.

Inbal-Tish: Updating `initialDate` does exactly the same as `updateMonth(date)`. The `addMonth(count)` method calculates the next/previous month from the current month using the count param passed and invokes `updateMonth()`. So basically the only difference is that you need to calculate the month you need to move to (instead of passing a count) and pass it to the `initialDate` prop.

Saisumanth99: Hello, In my case I am using the calendar inside a modal. swipe gestures of calendar are not working inside modal. So I wanted to have a explicit gesture handler sorrounding the modal using whose gestures I can modify the month.In my usecase I would like to find a way to modify month with my own gesture handler, I believe using ref is the only way for it. Please suggest me if there is any other way to implement my usecase.

PR #1891, opened 2022-05-31
Day - performance fix

Inbal-Tish:
Day - fix performance - all days are being rendered when selecting a new day

### DIFF
--- a/src/calendar/index.tsx
+++ b/src/calendar/index.tsx
@@ -107,7 +107,7 @@ const Calendar = (props: CalendarProps) => {
       return;
     }
     setCurrentMonth(newMonth);
-  }, []);
+  }, [currentMonth]);
 
   const addMonth = useCallback((count: number) => {
     const newMonth = currentMonth.clone().addMonths(count, true);
--- a/src/calendar/index.tsx
+++ b/src/calendar/index.tsx
@@ -102,12 +102,12 @@ const Calendar = (props: CalendarProps) => {
     }
   }, [currentMonth]);
 
-  const updateMonth = (newMonth: XDate) => {
+  const updateMonth = useCallback((newMonth: XDate) => {
     if (sameMonth(newMonth, currentMonth)) {
       return;
     }
     setCurrentMonth(newMonth);
-  };
+  }, []);
 
   const addMonth = useCallback((count: number) => {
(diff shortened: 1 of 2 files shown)


src/calendar/index.tsx
```diff
@@ -102,12 +102,12 @@ const Calendar = (props: CalendarProps) => {
...
   }, [currentMonth]);
 
-  const updateMonth = (newMonth: XDate) => {
+  const updateMonth = useCallback((newMonth: XDate) => {
     if (sameMonth(newMonth, currentMonth)) {
       return;
     }
     setCurrentMonth(newMonth);
-  };
+  }, []);
```
ethanshar, line 110: Shouldn't this callback be dependent on `currentMonth`, otherwise the check will be wrong. 
And If you add it as a dependency, just make sure the render issue for days is still fixed

Inbal-Tish: Right. Thanks. Fixed

ethanshar: How do you test it? 
I see still lots of renders when I play with it..

Inbal-Tish: I console.log the Day component

ethanshar: I console.log as well and Im getting lots of renders when I change the date. 
Which example screen are you checking?

Inbal-Tish: calendars screen with one example - the initial render I get 35 renders and then 2 when selecting a day

ethanshar: I checked ExpandableCalendar example, it looks like each press there trigger render of all days..

Inbal-Tish: oh. Fixed

ethanshar: Ok, I see it's fixed kinda - there are still more renders than it should, but it's much better now. 

Anyway, regarding the fix, I see you removed the `setDate` from the dependencies array, it looks problematic cause then you might use an "old" version of `setDate` callback... but then I see that the `setDate` callback is not even wrapped with useCallback, so maybe that's why it's working. 

TBH, the code and components are so tangled in each other I never know what we'll break next, especially when we don't have any test coverage. 

I'm approving for now, but we'll need to figure how to move forward

ethanshar (approved):

PR #1894, opened 2022-06-07
feat:Dynamic hourBlockHeight prop to change row height

MahmoudHemaid:
Hi,

I'm working on zoom in/out feature to control the timeline height. To be able to do that the hourBlockHeight should be controlled not a static number.


lidord-wix:
@MahmoudHemaid thank you very much for the contribution!
I see you have some conflicts with the master branch, can you please resolve them?

MahmoudHemaid:
@lidord-wix Can you review this change? I fixed the conflicts

src/timeline/Timeline.tsx
```diff
@@ -152,8 +158,8 @@ const Timeline = (props: TimelineProps) => {
     return map(pageDates, d => groupedEvents[d] || []);
   }, [pageDates, groupedEvents]);
   const scrollView = useRef<ScrollView>();
-  const calendarHeight = useRef((end - start) * HOUR_BLOCK_HEIGHT);
-  const styles = useRef(styleConstructor(theme || props.styles, calendarHeight.current));
+  const calendarHeight = useMemo(() => (end - start) * hourBlockHeight, [hourBlockHeight]);
+  const styles = useMemo(() => styleConstructor(theme || props.styles, calendarHeight), [calendarHeight]);
```
lidord-wix, line 162: There are some places in this file where you still use `styles.current...` even tough it's not a ref anymore

lidord-wix (commented):
Hi, thank you for the PR!
I wrote two small comments.
also, when I run the demo app and get into the `Timeline Calendar` screen I get a red screen, do you get it too? 
(I don't get it on master, but I still get it on your branch with the fixes I mentioned in the comments).

HaseebUrRehman98 (approved):
having same problem when needs to render multiple events in single block, so by removing fixed height for block hour it's should solve the problem, and also it's correct to recieve block hour height as TimelineProps

src/timeline/Timeline.tsx
```diff
@@ -167,25 +173,26 @@ const Timeline = (props: TimelineProps) => {
...
+        rightEdgeSpacing: rightEdgeSpacing / numberOfDays,
+        hourBlockHeight
       });
     });
   }, [pageEvents, start, numberOfDays]);
 
   useEffect(() => {
     let initialPosition = 0;
     if (scrollToNow) {
-      initialPosition = calcTimeOffset(HOUR_BLOCK_HEIGHT);
```
HaseebUrRehman98, line 178: there is problem with method calcTimeOffset, by changing hour block height won't return correct result for initialPosition.

PR #1899, opened 2022-06-15
Agenda - added ability to disable calendar

polcats:
This change adds the option to disable/hide the calendar on top of the Agenda component


src/agenda/index.tsx
```diff
@@ -318,9 +322,11 @@ export default class Agenda extends Component<AgendaProps, State> {
   };
 
   onDayChange = (day: XDate) => {
+    if (this.props.disableCalendar) return;
```
Inbal-Tish, line 325: Seems like it should be called `disableDayChange`... Also, it should disable the touch events as well (meaning the press indication), otherwise it will look buggy

PR #1923, opened 2022-07-06
Fix/calendar list performance

Inbal-Tish:


src/calendar-list/index.tsx
```diff
@@ -290,11 +288,12 @@ class CalendarList extends Component<CalendarListProps, State> {
...
 
   renderItem = ({item}: any) => {
-    const {calendarStyle, horizontal, calendarWidth, testID, ...others} = this.props;
-
+    const {horizontal, calendarStyle, calendarWidth, testID, markedDates, ...others} = this.props;
+    // NOTE: now only 'item' and 'markedDates' change for the 3 calendar (item.getTime) items
     return (
       <CalendarListItem
         {...others}
+        markedDates={item.getTime ? markedDates : undefined}
```
ethanshar, line 296: We can improve more by not passing to months that don't have any marking dates.. 
it basically can reduce render of 2 more items, but it still something

src/calendar-list/item.tsx
```diff
@@ -21,120 +20,100 @@ export type CalendarListItemProps = CalendarProps & {
...
-    if (onPressArrowLeft) {
-      onPressArrowLeft(method, monthClone);
-    } else if (scrollToMonth) {
-      const currentMonth = monthClone.getMonth();
-      monthClone.addMonths(-1);
-
-      // Make sure we actually get the previous month, not just 30 days before currentMonth.
-      while (monthClone.getMonth() === currentMonth) {
-        monthClone.setDate(monthClone.getDate() - 1);
+const CalendarListItem = (props: CalendarListItemProps) => {
```
ethanshar, line 23: It's very possible you still need to wrap this with React.memo so the component behave as a pure component

src/calendar-list/index.tsx
Inbal-Tish: Ok. Check my last commit please

PR #1925, opened 2022-07-07
Calendar & Day - Performance improvements

Inbal-Tish:
Calendar & Day - replacing 'extractComponentProps' with less generic method, and reimplementing toMarkingString function

### DIFF
--- a/src/calendar/day/index.tsx
+++ b/src/calendar/day/index.tsx
@@ -17,7 +17,8 @@ import PeriodDay from './period';
 function areEqual(prevProps: DayProps, nextProps: DayProps) {
   const prevPropsWithoutMarkDates = omit(prevProps, 'marking');
   const nextPropsWithoutMarkDates = omit(nextProps, 'marking');
-  const areOthersEqual = some(prevPropsWithoutMarkDates, function(value, key) { 
+  const areOthersEqual = some(prevPropsWithoutMarkDates, function(value, key) {
+    //@ts-expect-error
     return value !== nextPropsWithoutMarkDates[key]; 
   });
   const isMarkingEqual = isEqual(prevProps.marking, nextProps.marking);
--- a/src/calendar/day/index.tsx
+++ b/src/calendar/day/index.tsx
@@ -17,12 +17,12 @@ import PeriodDay from './period';
 function areEqual(prevProps: DayProps, nextProps: DayProps) {
   const prevPropsWithoutMarkDates = omit(prevProps, 'marking');
   const nextPropsWithoutMarkDates = omit(nextProps, 'marking');
-  const areOthersEqual = some(prevPropsWithoutMarkDates, function(value, key) {
+  const didPropsChange = some(prevPropsWithoutMarkDates, function(value, key) {
     //@ts-expect-error
     return value !== nextPropsWithoutMarkDates[key]; 
   });
   const isMarkingEqual = isEqual(prevProps.marking, nextProps.marking);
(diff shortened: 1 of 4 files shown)


src/componentUpdater.ts
```diff
@@ -27,3 +28,33 @@ export function extractComponentProps(component: any, props: any, ignoreProps?:
   }
   return {};
 }
+
+export function extractDayProps(props: CalendarProps) {
```
ethanshar, line 32: @Inbal-Tish 
Weren't there other extract methods for other components? 
Only the Day component has an "extractProps" method?

ethanshar (approved):
@Inbal-Tish 
Left one comment, anyway approved

src/componentUpdater.ts
Inbal-Tish: We have many usages for extractComponentProps, should I include the change in this PR? I thought of opening a different one for the rest...

ethanshar: Ok, let's merge this and have a separate one..

Inbal-Tish: I added another commit to fix `areEqual` function. Can you please take a look?

src/calendar/day/index.tsx
```diff
@@ -16,8 +16,9 @@ import PeriodDay from './period';
 function areEqual(prevProps: DayProps, nextProps: DayProps) {
   const prevPropsWithoutMarkDates = omit(prevProps, 'marking');
   const nextPropsWithoutMarkDates = omit(nextProps, 'marking');
+  const arePropsWithoutMarkedDatesEqual = isEqual(prevPropsWithoutMarkDates, nextPropsWithoutMarkDates);
```
ethanshar, line 19: Is `isEqual` runs a deep check? 
If so, this makes the whole separation redundant. 
If you check for `isEqual` for both `markedDates` and the rest of the props you can just check the whole props object. 

The whole idea was to avoid deep equal check on the "other" props. cause deep check is considered heavy
But now I see that lodash's `omit` util creates a new object so it will return false when shallow comparing. 
Do we have a util for shallow comparison (one that runs on 1st level)?

src/componentUpdater.ts
ethanshar: See my comment.

src/calendar/day/index.tsx
Inbal-Tish: Yes, `isEqual` runs deep. The problem is when you check `prevPropsWithoutMarkDates === nextPropsWithoutMarkDates` it's always false (as the references are not equal...).
The purpose of the separation is to know if some other prop was updated, and if not deep check the `marking` prop if it was really changed before updating the component. This way it re-renders only the dates where the `marking` value changed

Inbal-Tish: Ok. I see what you mean... I can just us `isEqual` on all props.

ethanshar: Ok sababa
`20.72 ms` with the fix? 
How much was it before?

Inbal-Tish: 7.44 we need a shallow compare for the props

ethanshar: ohh ok. 
(It's really nice that we can see the immediate impact.. )


Anyway, we can just iterate the otherProps and check them using `===` comparison, basically it'll check their reference has not changed and that's enough (that's what React does). 
We can use `_.find` or `_.some` to iterate all the props keys and compare

PR #1927, opened 2022-07-09
Fix onLongPress for expandable calendar's week component

maks-io:
The expandable calendar's week component was re-using the onDayPress function also for the long press action. This PR aims to fix this.

### DIFF
--- a/src/expandableCalendar/week.tsx
+++ b/src/expandableCalendar/week.tsx
@@ -15,7 +15,7 @@ import Day from '../calendar/day/index';
 export type WeekProps = CalendarProps;
 
 const Week = (props: WeekProps) => {
-  const {theme, current, firstDay, hideExtraDays, markedDates, onDayPress, style: propsStyle, numberOfDays = 1, timelineLeftInset} = props;
+  const {theme, current, firstDay, hideExtraDays, markedDates, onDayPress, onLongPress, style: propsStyle, numberOfDays = 1, timelineLeftInset} = props;
   const style = useRef(styleConstructor(theme));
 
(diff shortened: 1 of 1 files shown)


src/expandableCalendar/week.tsx
```diff
@@ -15,7 +15,7 @@ import Day from '../calendar/day/index';
 export type WeekProps = CalendarProps;
 
 const Week = (props: WeekProps) => {
-  const {theme, current, firstDay, hideExtraDays, markedDates, onDayPress, style: propsStyle, numberOfDays = 1, timelineLeftInset} = props;
+  const {theme, current, firstDay, hideExtraDays, markedDates, onDayPress, onLongPress, style: propsStyle, numberOfDays = 1, timelineLeftInset} = props;
```
Inbal-Tish, line 18: The name of the props is wrong - you need Calendar's `onDayLongPress`

maks-io: Thanks for the review @Inbal-Tish - i lack knowledge about the sematic difference of onLongPress vs onLongDayPress, what i did though was modifying it in my local repository where i had react-native-calendars installed as a dependency (directly modified it in the node_modules), and the thing i wanted to achieve only worked, when i modified it as in the PR. When i use onDayLongPress instead, it - for some reason - does not get triggered if i long press on the week component's day elements.
Maybe i am missing something, can u help me investigate this?

Inbal-Tish:
@maks-io See PR #1940

maks-io:
> @maks-io See PR #1940

Thanks @Inbal-Tish , that makes sense, thanks for tackling it!

PR #1929, opened 2022-07-11
Fix/remove generic extract prop

Inbal-Tish:


src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -139,8 +134,8 @@ class WeekCalendar extends Component<WeekCalendarProps, State> {
   };
 
   renderItem = ({item}: any) => {
-    const {style, onDayPress, markedDates, firstDay, ...others} = extractComponentProps(Week, this.props);
-    const {context} = this.props;
+    const {allowShadow, context, ...calendarListProps} = this.props;
+    const {style, onDayPress, markedDates, firstDay = 1, ...others} = extractCalendarProps(calendarListProps);
```
ethanshar, line 138: Just make sure setting default as `1` is not a problem

ethanshar (approved):

src/expandableCalendar/WeekCalendar/index.tsx
Inbal-Tish: Fixed

PR #1931, opened 2022-07-12
General bugs and performance issues

ethanshar:
I went over ExpandableCalendar code, there are a few issues I found and In general the context communication with all of the component might have bugs.

- I marked with TODOs places I think might introduce a bug
- I added a debounce in AgendaList for sync scrolls that happen to often (IMO the scrolling is pretty heavy) 
- In the Expandable Provider, we should get rid all the presenter usages - I added a specific comment there. Let me know what u think. I personally think that without that it's impossible to refactor the code
- I suggest working only on Android where performance are less... good. In general I found a bug that I think might explain some of the problems. In ExpandableCalendar screen, try moving a few months fast with the arrows. after you'll stop you'll see it'll stuck in a loop of moving between two months - my theory is that we have something that trigger extra scrolls

### DIFF
--- a/src/expandableCalendar/agendaList.tsx
+++ b/src/expandableCalendar/agendaList.tsx
@@ -2,6 +2,7 @@ import get from 'lodash/get';
 import map from 'lodash/map';
 import isFunction from 'lodash/isFunction';
 import isUndefined from 'lodash/isUndefined';
+import debounce from 'lodash/debounce';
 
 import PropTypes from 'prop-types';
 import XDate from 'xdate';
@@ -94,7 +95,7 @@ const AgendaList = (props: AgendaListProps) => {
   useEffect(() => {
     if (date !== _topSection.current) {
       setTimeout(() => {
-        scrollToSection();
+        scrollToSection(date);
       }, 500);
     }
   }, []);
@@ -102,7 +103,7 @@ const AgendaList = (props: AgendaListProps) => {
   useEffect(() => {
     // NOTE: on first init data should set first section to the current date!!!
     if (updateSource !== UpdateSources.LIST_DRAG && updateSource !== UpdateSources.CALENDAR_INIT) {
-      scrollToSection();
+      scrollToSection(date);
     }
   }, [date]);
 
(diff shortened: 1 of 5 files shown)


src/expandableCalendar/index.tsx
```diff
@@ -530,6 +533,7 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
     );
   };
 
+  // TODO: Shouldn't this be wrapped with useCallback?
```
Inbal-Tish, line 536: Will do

src/expandableCalendar/index.tsx
```diff
@@ -281,11 +281,13 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
     if (!horizontal) {
       calendar?.current?.scrollToDay(new XDate(date), 0, true);
     } else if (getYear(date) !== visibleYear.current || getMonth(date) !== visibleMonth.current) {
+      // TODO: Investigate why this is being called when moving months with arrow press
       // don't scroll if the month is already visible
       calendar?.current?.scrollToMonth(date);
     }
   };
 
+  // TODO: Why we're not wrapping this with useCallback?
```
Inbal-Tish, line 290: Will do

src/expandableCalendar/index.tsx
```diff
@@ -562,6 +566,7 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
             horizontal={horizontal}
             firstDay={firstDay}
             calendarStyle={calendarStyle}
+            // TODO: others might container unwanted props, we should use extractCalendarListProps
             {...others}
```
Inbal-Tish, line 570: Actually the 'others' here are already extracting all the Expandable's props and more

src/expandableCalendar/index.tsx
```diff
@@ -351,6 +353,7 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
     bounceToPosition();
   };
 
+  // TODO: We should consider wrapping this with ref/useMemo cause it's being created on each render
```
Inbal-Tish, line 356: Will do

src/expandableCalendar/agendaList.tsx
```diff
@@ -158,8 +159,8 @@ const AgendaList = (props: AgendaListProps) => {
     return sectionTitle;
   };
 
-  const scrollToSection = () => {
-    const sectionIndex = scrollToNextEvent ? getNextSectionIndex(date) : getSectionIndex(date);
+  const scrollToSection = useCallback(debounce((d) => {
```
Inbal-Tish, line 162: Moved to a separate PR

PR #1932, opened 2022-07-12
CalendarContextProvider - use with pure functions

Inbal-Tish:
Move `setDate` and `setDisabled` back to Provider and removing the Presenter's relevant tests

### DIFF
--- a/src/expandableCalendar/Context/Provider.tsx
+++ b/src/expandableCalendar/Context/Provider.tsx
@@ -1,17 +1,15 @@
-import PropTypes from 'prop-types';
 import XDate from 'xdate';
 
 import React, {useEffect, useRef, useState, useCallback, useMemo} from 'react';
 import {Animated, TouchableOpacity, View, ViewStyle, ViewProps, StyleProp} from 'react-native';
 
-import {toMarkingFormat} from '../../interface';
+import {sameMonth} from '../../dateutils';
+import {xdateToData} from '../../interface';
 import {Theme, DateData} from '../../types';
-import styleConstructor from '../style';
 import {UpdateSources} from '../commons';
+import styleConstructor from '../style';
 import CalendarContext from './index';
 import {
-  setDate,
-  setDisabled,
   shouldAnimateTodayButton,
   shouldAnimateOpacity,
   getButtonIcon,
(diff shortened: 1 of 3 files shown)


src/expandableCalendar/Context/Provider.tsx
```diff
@@ -65,20 +63,19 @@ const CalendarProvider = (props: CalendarContextProviderProps) => {
...
+  const today = useRef(getTodayFormatted());
   const wrapperStyle = useMemo(() => {
     return [style.current.contextWrapper, propsStyle];
   }, [style, propsStyle]);
 
+  const [prevDate, setPrevDate] = useState(date);
+  const [currentDate, setCurrentDate] = useState(date);
+  const [updateSource, setUpdateSource] = useState(UpdateSources.CALENDAR_INIT);
+  const [isDisabled, setIsDisabled] = useState(false);
+  const [buttonIcon, setButtonIcon] = useState(getButtonIcon(date, showTodayButton));
```
ethanshar, line 77: Move useState's before useMemo's

src/expandableCalendar/Context/Provider.tsx
```diff
@@ -89,24 +86,27 @@ const CalendarProvider = (props: CalendarContextProviderProps) => {
...
+  /** Context */
+
   const _setDate = (date: string, updateSource: UpdateSources) => {
-    const updateState = (buttonIcon: number) => {
-      setPrevDate(currentDate);
-      setCurrentDate(date);
-      setUpdateSource(updateSource);
-      setButtonIcon(buttonIcon);
-    };
+    setPrevDate(currentDate);
```
ethanshar, line 92: I wonder of these calls trigger multiple renders

ethanshar (approved):

src/expandableCalendar/Context/Provider.tsx
Inbal-Tish: When I log the provider, I see one call after `setDate`.

src/expandableCalendar/Context/Provider.tsx
Inbal-Tish: Done

PR #1939, opened 2022-07-22
ExpandableCalendar - close calendar with knob press

Inbal-Tish:
Also: reorganize code, remove propTypes, and wrap const with useMemo and useCallback

### DIFF
--- a/src/expandableCalendar/index.tsx
+++ b/src/expandableCalendar/index.tsx
@@ -1,10 +1,8 @@
 import first from 'lodash/first';
-import values from 'lodash/values';
 import isFunction from 'lodash/isFunction';
 import isNumber from 'lodash/isNumber';
 import throttle from 'lodash/throttle';
 
-import PropTypes from 'prop-types';
 import XDate from 'xdate';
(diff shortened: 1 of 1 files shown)


src/expandableCalendar/index.tsx
```diff
@@ -403,6 +405,15 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
     }
   };
 
+  const closeCalendar = () => {
```
ethanshar, line 408: You need to wrap with useCallback 
Unless we can move `isOpen` to be a ref. 

I noticed you're using isOpen in a lot of methods, but not sure if it actually affects the UI. 
If it's just a flag, do you think we can use a ref for it?

Inbal-Tish: What do you mean `isOpen` doesn't affect the UI? Many styles depend on this check...

ethanshar: right, so nvm

ethanshar (approved):

PR #1942, opened 2022-07-22
ContextProvider - wrap callbacks

Inbal-Tish:


src/expandableCalendar/Context/Provider.tsx
```diff
@@ -89,26 +91,26 @@ const CalendarProvider = (props: CalendarContextProviderProps) => {
...
 
-    props.onDateChanged?.(date, updateSource);
+    onDateChanged?.(date, updateSource);
 
     if (!sameMonth(new XDate(date), new XDate(currentDate))) {
-      props.onMonthChange?.(xdateToData(new XDate(date)), updateSource);
+      onMonthChange?.(xdateToData(new XDate(date)), updateSource);
     }
-  };
+  }, [currentDate, onDateChanged, onMonthChange]);
```
ethanshar, line 105: I think the main issue we'll have here, is the fact that `currentData` will change too often and will re-create the `_setDate` callback. 
This is btw and known React issue, and they have plans to release a new hook called `useEvent` 
You can read about it [here](https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md)

Anyway, for now I will approve the PR. 
We can maybe think of a way to avoid re-creation of the callback for now, maybe keep the `currentDate` as a ref value as well

ethanshar (approved):

src/expandableCalendar/Context/Provider.tsx
Inbal-Tish: Yes, I'm familiar with `useEvent`. I removed the `prevDate` from the state, as there is no need for that, but I don't see an option removing the `currentDate` from the state as the UI needs to react to its changes (today button, selected date, etc.)

Inbal-Tish: oh. Got it.

ethanshar (approved):

PR #1950, opened 2022-07-25
Week - improve performance

yuvalsho:


src/expandableCalendar/WeekCalendar/new.tsx
```diff
@@ -76,7 +82,7 @@ const WeekCalendar = (props: WeekCalendarProps) => {
 
   const renderItem = useCallback(
     (_type: any, item: string) => {
-      const {allowShadow, ...calendarListProps} = props;
+      const {...calendarListProps} = props;
```
Inbal-Tish, line 85: `allowShadow` was extracted since it's not a CalendarList prop

src/expandableCalendar/week.tsx
```diff
@@ -51,27 +50,23 @@ const Week = (props: WeekProps) => {
         />
       </View>
     );
-  };
+  }, [props]);
```
Inbal-Tish, line 53: What good is the wrap with useCallback here if the dependency is `props`?

Inbal-Tish:
@yuvalsho Hi. I see you worked on the "new" `WeekCalendar`. Currently, the ExpandableCalendar is using the old one. It's worth checking out or seeing if we can change the usage from the old to the new with no side effects.

src/expandableCalendar/index.tsx
```diff
@@ -29,7 +29,7 @@ import WeekDaysNames from '../commons/WeekDaysNames';
 import Calendar from '../calendar';
 import CalendarList, {CalendarListProps} from '../calendar-list';
 import Week from './week';
-import WeekCalendar from './WeekCalendar';
+import WeekCalendar from './WeekCalendar/new';
```
Inbal-Tish, line 32: Let's remove this change

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -149,74 +85,77 @@ class WeekCalendar extends Component<WeekCalendarProps, State> {
...
+  );
+};
+
+WeekCalendar.displayName = 'WeekCalendar';
+WeekCalendar.defaultProps = {
+  firstDay: 0,
+  allowShadow: true
+};
+
+function shouldUpdate(prevProps: WeekCalendarProps, nextProps: WeekCalendarProps) {
```
Inbal-Tish, line 151: If we migrated to functional component, why not do this in `useEffect`?

src/expandableCalendar/WeekCalendar/presenter.ts
```diff
@@ -62,6 +62,9 @@ class Presenter {
   };
 
   shouldComponentUpdate = (context: any, prevContext: any) => {
+    if (!context) {
```
Inbal-Tish, line 65: I would love to get rid of this function and on a more general note - I would love to get rid of this presenter (or at least to have it for pure functions alone).

src/expandableCalendar/asCalendarConsumer.tsx
```diff
@@ -4,7 +4,7 @@ import hoistNonReactStatic from 'hoist-non-react-statics';
 import CalendarContext from './Context';
 
 
-function asCalendarConsumer<PROPS>(WrappedComponent: React.ComponentType<any>): React.ComponentClass<PROPS> {
+function asCalendarConsumer<PROPS>(WrappedComponent: React.ComponentType<any>, functional = false): React.ComponentClass<PROPS> {
```
Inbal-Tish, line 7: If we warp the functional component with `forwardRef` I think we won't need this, can you check?

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,144 +1,80 @@
...
-  };
-
-  onScroll = (event: NativeSyntheticEvent<NativeScrollEvent>) => {
+const WeekCalendar = (props: WeekCalendarProps) => {
+  const {calendarWidth, firstDay, hideDayNames, current, markedDates, allowShadow, context, theme, style} = props;
+  const defaultStyle = useMemo(() => styleConstructor(theme), [theme]);
+  const presenter = new Presenter();
+  let page = NUMBER_OF_PAGES;
+  const [items, setItems] = useState(presenter.getDatesArray(props));
+  const {dayHeader} = defaultStyle;
```
Inbal-Tish, line 37: Why this is not wrapped with `useMemo`?

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,144 +1,80 @@
...
+  }, [context]);
+
+  const onScroll = (event: NativeSyntheticEvent<NativeScrollEvent>) => {
     const x = event.nativeEvent.contentOffset.x;
-    const {onScroll} = this.presenter;
-    const {context} = this.props;
-    const {items} = this.state;
-    const {containerWidth, page} = this;
+    const {onScroll} = presenter;
+    const {context} = props;
```
Inbal-Tish, line 50: no need to extract from props again. You have it at the top...

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,144 +1,80 @@
...
 
     const updateState = (newData: string[], newPage: number) => {
-      this.page = newPage;
-      this.setState({items: [...newData]});
+      page = newPage;
+      setItems([...newData]);
```
Inbal-Tish, line 54: is the spread necessary?

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,144 +1,80 @@
...
     };
 
     onScroll({context, updateState, x, page, items, width: containerWidth});
   };
 
-  onMomentumScrollEnd = () => {
-    const {items} = this.state;
-    const {onMomentumScrollEnd} = this.presenter;
+  const getWeekStyle = useMemo(() => {
```
Inbal-Tish, line 60: once you use memo, the name of the const should be `weekStyle` and not "get"

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,144 +1,80 @@
...
-  });
-
-  onDayPress = (value: DateData) => {
-    this.presenter.onDayPress(this.props.context, value);
-  };
-
-  onScroll = (event: NativeSyntheticEvent<NativeScrollEvent>) => {
+const WeekCalendar = (props: WeekCalendarProps) => {
+  const {calendarWidth, firstDay, hideDayNames, current, markedDates, allowShadow, context, theme, style} = props;
+  const defaultStyle = useMemo(() => styleConstructor(theme), [theme]);
```
Inbal-Tish, line 33: As a convention in the library we use this as `style` and the style from props we rename `propsStyle`

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,144 +1,80 @@
...
+  const presenter = new Presenter();
```
Inbal-Tish, line 34: I personally don't like the presenter and would love to see it as a service for pure functions (not a class), while moving all the events back to the component. What do you think?

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -149,74 +85,77 @@ class WeekCalendar extends Component<WeekCalendarProps, State> {
...
+        style={getWeekStyle}
         markedDates={markedDates}
         onDayPress={onDayPress}
         context={currentContext}
         numberOfDays={context.numberOfDays}
         timelineLeftInset={context.timelineLeftInset}
       />
     );
-  };
+  },[]);
```
Inbal-Tish, line 96: No dependencies array?? not even `getWeekStyle`? Also, Let's move the `others` logic out so we won't do it on every item render

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -149,74 +85,77 @@ class WeekCalendar extends Component<WeekCalendarProps, State> {
...
 
-  renderWeekDaysNames = () => {
+  const renderWeekDaysNames = useMemo(() => {
     return (
-      <WeekDaysNames 
-        firstDay={this.props.firstDay} 
-        style={this.style.dayHeader} 
+      <WeekDaysNames
+        firstDay={firstDay}
+        style={dayHeader}
```
Inbal-Tish, line 104: We usually don't consider the default style as dynamic so once you created the style ref it shouldn't change, thus the `dayHeader` should stay `style?.current?.dayHeader` and be removed from the dependencies array.

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -149,74 +85,77 @@ class WeekCalendar extends Component<WeekCalendarProps, State> {
...
 
-  render() {
-    const {allowShadow, firstDay, hideDayNames, current, context} = this.props;
-    const {items} = this.state;
-    const extraData = Map({
-      current,
-      date: context.date,
-      firstDay
-    });
+  const extraData = Map({
```
Inbal-Tish, line 109: Should we wrap in `useMemo` here?

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -149,74 +85,77 @@ class WeekCalendar extends Component<WeekCalendarProps, State> {
...
-          getItemLayout={this.getItemLayout}
-          onScroll={this.onScroll}
-          onMomentumScrollEnd={this.onMomentumScrollEnd}
-        />
-      </View>
-    );
+  return (
+    <View
+      testID={props.testID}
+      style={[allowShadow && defaultStyle.containerShadow, !hideDayNames && defaultStyle.containerWrapper]}
```
Inbal-Tish, line 118: Let's wrap with `useMemo`

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -149,74 +85,77 @@ class WeekCalendar extends Component<WeekCalendarProps, State> {
...
+    >
+      {!hideDayNames && (
+        <View style={[defaultStyle.week, defaultStyle.weekCalendar]}>
```
Inbal-Tish, line 121: Let's wrap with `useMemo`

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -149,74 +85,77 @@ class WeekCalendar extends Component<WeekCalendarProps, State> {
...
+          {renderWeekDaysNames}
+        </View>
+      )}
+          <FlashList
+            ref={presenter.list as unknown as LegacyRef<FlashList<string>>}
+            data={items}
+            extraData={extraData}
```
Inbal-Tish, line 128: The `extraData` will cause a re-render whenever the context's `date` changes, but also the `onScroll` is setting the context date (setDate) and setting the state's `items`. It's a mess (legacy I know) but let's change it if we're refactoring here. We should only have a `useEffect` for the context's `date` prop that changes the `items`. All other sources of updates feel redundant. What do you think?

src/expandableCalendar/WeekCalendar/index.tsx
yuvalsho: because i wont have the prevProps there, and i wouldn't know what the previous date was

Inbal-Tish: If the useEffect with `context` dependency is called, you know the context change. Second, the previous date you can keep in `useRef` and check `sameWeek` against it - update it only after.

PR #1951, opened 2022-07-25
Make moment actually optional

liamjones:
Fix for PR #1446


liamjones:
I've just realised something while discussing this with a colleague - the existing code _should_ be enough but for some reason Metro (e.g. the RN packager) wants to resolve the require at bundle time, even though it's in a function body?

We've got `inlineRequires` enabled so I'm not sure why it's doing this...

Inbal-Tish (approved):

PR #1961, opened 2022-08-07
Week calendar refactor

yuvalsho:


src/expandableCalendar/week.tsx
```diff
@@ -31,7 +31,7 @@ const Week = (props: WeekProps) => {
   const renderDay = (day: XDate, id: number) => {
     const dayProps = extractDayProps(props);
     const currXdate = parseDate(current);
-
```
Inbal-Tish, line 34: Please remove the changes here. There is no need to include this file if there are no actual changes here

src/infinite-list/index.tsx
```diff
@@ -44,7 +44,7 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
     scrollViewProps,
     positionIndex = 0
   } = props;
-
```
Inbal-Tish, line 47: Please remove the changes here. There is no need to include this file if there are no actual changes here

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,222 +1,239 @@
-import memoize from 'memoize-one';
```
Inbal-Tish, line 1: Why change the order of imports? We have an order - js libs, react and RN libs, internal services, and tools, internal components, component-specific imports

Inbal-Tish:
@yuvalsho Can you please check the conflicts and the failing build?

Inbal-Tish:
@yuvalsho Hi. Please pull master and resolve conflicts

Inbal-Tish:
@yuvalsho Why there are 18 files changed? Can we please keep this PR clean only for the WeekCalendar changes?

src/calendar/day/index.tsx
```diff
@@ -12,13 +9,14 @@ import {SELECT_DATE_SLOT} from '../../testIDs';
 import {DateData} from '../../types';
 import BasicDay, {BasicDayProps} from './basic';
 import PeriodDay from './period';
+import {isEqual, omit, some} from 'lodash';
```
Inbal-Tish, line 12: No. The reason we import each separately is to control the import package size. This is how we do it all through the lib, please don't change it.

src/calendar/index.tsx
```diff
@@ -46,9 +46,9 @@ export interface CalendarProps extends CalendarHeaderProps, DayProps {
   /** Always show six weeks on each month (only when hideExtraDays = false) */
   showSixWeeks?: boolean;
   /** Handler which gets executed on day press */
-  onDayPress?: (date: DateData) => void;
+  onDayPress?: (date?: DateData) => void;
```
Inbal-Tish, line 49: No no. You can't just change the return type. This is not part of this PR scope and it's a breaking change

src/day-state-manager.ts
```diff
@@ -1,10 +1,11 @@
+import {DayState} from './types';
 const {isToday, isDateNotInRange, sameMonth} = require('./dateutils');
 const {toMarkingFormat} = require('./interface');
 
 
-export function getState(day: XDate, current: XDate, props: any) {
+export function getState(day: XDate, current: XDate, props: any): DayState {
```
Inbal-Tish, line 6: Please separate to a diferent PR

example/src/mocks/AgendaItem.tsx
```diff
@@ -39,7 +40,7 @@ const AgendaItem = (props: ItemProps) => {
       </View>
     </TouchableOpacity>
   );
-};
+}, isEqual);
```
Inbal-Tish, line 43: This is expensive, why do you need this extra comparison?

example/src/screens/calendarListScreen.tsx
```diff
@@ -35,9 +35,11 @@ const CalendarListScreen = (props: Props) => {
       }
     };
   }, [selected]);
-  
-  const onDayPress = useCallback((day: DateData) => {
-    setSelected(day.dateString);
+
+  const onDayPress = useCallback((day?: DateData) => {
```
Inbal-Tish, line 39: Please move to a different PR

src/calendar/index.tsx
```diff
@@ -46,9 +46,9 @@ export interface CalendarProps extends CalendarHeaderProps, DayProps {
...
   /** Handler which gets executed on day long press */
-  onDayLongPress?: (date: DateData) => void;
+  onDayLongPress?: (date?: DateData) => void;
```
Inbal-Tish, line 51: Same here

src/expandableCalendar/WeekCalendar/new.tsx
```diff
@@ -1,160 +0,0 @@
-import React, {useCallback, useContext, useEffect, useMemo, useRef, useState} from 'react';
```
Inbal-Tish, line 1: Why are you deleting this file???

src/expandableCalendar/index.tsx
```diff
@@ -418,8 +418,8 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
     scrollPage(true);
   }, [onPressArrowRight, scrollPage]);
 
-  const _onDayPress = useCallback((value: DateData) => {
-    if (numberOfDaysCondition) {
+  const _onDayPress = useCallback((value?: DateData) => {
+    if (numberOfDaysCondition && value) {
```
Inbal-Tish, line 422: No. `value` is not optional

src/expandableCalendar/week.tsx
```diff
@@ -7,27 +7,48 @@ import {parseDate, toMarkingFormat} from '../interface';
 import {getState} from '../day-state-manager';
 import {extractDayProps} from '../componentUpdater';
 import styleConstructor from './style';
-import Calendar, {CalendarProps} from '../calendar';
+import {CalendarProps} from '../calendar';
 import Day from '../calendar/day/index';
-// import BasicDay from '../calendar/day/basic';
+import {CalendarContextProps} from './Context';
+import {isEmpty, isEqual} from 'lodash';
```
Inbal-Tish, line 13: Please separate to two imports for each function

src/index.ts
```diff
@@ -7,7 +7,7 @@ export {default as Agenda} from './agenda';
 export type {AgendaProps} from './agenda';
 export {default as ExpandableCalendar} from './expandableCalendar';
 export type {ExpandableCalendarProps} from './expandableCalendar';
-export {default as WeekCalendar} from './expandableCalendar/WeekCalendar/new';
```
Inbal-Tish, line 10: We don't need to change it for everybody, not in this PR anyways...

src/expandableCalendar/index.tsx
```diff
@@ -530,7 +530,7 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
           firstDay={firstDay}
           {...others}
           allowShadow={disableWeekScroll ? undefined : false}
-          current={date}
+          current={disableWeekScroll ? date : undefined}
```
Inbal-Tish, line 533: When adding `disableWeekScroll` prop to the ExpandableCalendar the arrows stop working.

src/expandableCalendar/week.tsx
```diff
@@ -7,27 +7,48 @@ import {parseDate, toMarkingFormat} from '../interface';
...
+    ...prevProps.context,
+  };
+
+  const nextPropsToCheck = {
+    ...nextProps.markedDates,
+    ...nextProps.context,
+  };
 
-const Week = (props: WeekProps) => {
+  return isEmpty(nextProps.context) || isEqual(prevPropsToCheck, nextPropsToCheck);
```
Inbal-Tish, line 34: You're already checking `isEmpty(nextProps.context)` at the top

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,221 +1,225 @@
...
+  const visibleWeek = useRef(date);
+  const style = useRef(styleConstructor(theme));
+  const [items, setItems] = useState(getDatesArray(current ?? date, firstDay, numberOfDays));
+  const changedItems = useRef(constants.isRTL);
+  const list = useRef<FlatList>(null);
+
+  useEffect(() => {
+    if (updateSource !== UpdateSources.WEEK_SCROLL) {
+      const pageIndex = items.findIndex(item => sameWeek(item, date, firstDay));
+      const oldIndex = items.findIndex(item => sameWeek(item, prevDate, firstDay));
```
Inbal-Tish, line 52: Why not store the `oldIndex` on a ref and save this calc for every date change?

src/expandableCalendar/week.tsx
```diff
@@ -7,27 +7,48 @@ import {parseDate, toMarkingFormat} from '../interface';
...
+import {isEmpty, isEqual} from 'lodash';
 
 
-export type WeekProps = CalendarProps;
+export type WeekProps = CalendarProps & {
+  context?: CalendarContextProps;
+};
+
+const propsAreEqual = (prevProps: WeekProps, nextProps: WeekProps) => {
+  if (isEmpty(nextProps.context)) {
```
Inbal-Tish, line 21: This is blocking updates from the compoent when it is not use from the WeekCalendar, meaning, when there is no `context` passed. In the case of ExpandableCalendar, passing `disableWeekScroll` prop will render the Week component directly and it will not work properly - see my comment above.

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,221 +1,225 @@
...
+  }, [calendarWidth]);
 
-  getDate(weekIndex: number) {
-    const {current, context, firstDay = 0} = this.props;
-    const d = new XDate(current || context.date);
-    // get the first day of the week as date (for the on scroll mark)
-    let dayOfTheWeek = d.getDay();
-    if (dayOfTheWeek < firstDay && firstDay > 0) {
-      dayOfTheWeek = 7 + dayOfTheWeek;
+  const onDayPressCallback = useCallback((value?: DateData) => {
```
Inbal-Tish, line 66: No need for the `Callback` suffix here

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,221 +1,225 @@
...
+    if (onDayPress) {
+      onDayPress(value);
+    } else if (value) {
```
Inbal-Tish, line 69: `value` is not optional

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,221 +1,225 @@
...
+    d.addDays(firstDay - dayOfTheWeek);
   }
+
+  const newDate = numberOfDays && numberOfDays > 1 ? generateDay(d, weekIndex * numberOfDays) : d.addWeeks(weekIndex);
+
+  return toMarkingFormat(newDate);
 }
 
-export default asCalendarConsumer<WeekCalendarProps>(WeekCalendar);
+function getDatesArray (date: string, firstDay: number, numberOfDays?: number) {
```
Inbal-Tish, line 217: remove space between function name and parameters parentheses.

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,221 +1,225 @@
...
 
     return (
       <Week
         {...others}
         key={item}
         current={item}
         firstDay={firstDay}
-        style={this.getWeekStyle(this.containerWidth, style)}
-        onDayPress={onDayPress}
+        style={getWeekStyle}
```
Inbal-Tish, line 87: Let's change the name from `getWeekStyle` to `WeekStyle` since it's memoized const and not a method.

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,221 +1,225 @@
...
+  const getItemLayout = useCallback((_, index: number) => {
+    return {
+      length: containerWidth,
+      offset: containerWidth * index,
+      index
+    };
+  }, [containerWidth]);
+
+  const onViewableItemsChanged = useCallback(({viewableItems}: { viewableItems: Array<ViewToken>}) => {
+    const newDate = viewableItems[0]?.item;
```
Inbal-Tish, line 131: You can move this line after the check since you don't need it in case you return here...

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,221 +1,225 @@
...
+        }
+      } else {
+        visibleWeek.current = newDate;
+        setDate(newDate, UpdateSources.WEEK_SCROLL);
+        if (visibleWeek.current === items[0]) {
+          onEndReached();
+        }
+      }
+    }
+  }, [items]);
```
Inbal-Tish, line 154: Don't you need `onEndReached` as a dependency here as well?

Inbal-Tish:
@yuvalsho Overall looks and feels good. Just a few comments about this PR's scope and some code standards. 
Also, I did find some little bugs, not only in the ExpandableCalendarScreen but also in other screens, like WeekCalendarScreen and TimelineCalendarScreen (with 'numberOfDays` passed). Need to make sure we're not breaking anything in the process...

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,221 +1,225 @@
...
+    };
+  }, [containerWidth]);
+
+  const onViewableItemsChanged = useCallback(({viewableItems}: { viewableItems: Array<ViewToken>}) => {
+    const newDate = viewableItems[0]?.item;
+    if (changedItems.current) {
+      changedItems.current = false;
+      return;
+    }
+    if (newDate !== visibleWeek.current) {
```
Inbal-Tish, line 136: Don't you need to check if the two dates are not in the same week before updating?

src/expandableCalendar/index.tsx
yuvalsho: it already exists in the code, all i did was to pass current only for single week view instead of WeekCalendar (it is required for week but in WeekCalednar we get the date from the context which means that this prop updates for no reason

src/expandableCalendar/WeekCalendar/index.tsx
yuvalsho: newDate is the item name/date (since this is function is only for week scroll), since each item is a different week then by definition they are not in the same week

src/expandableCalendar/index.tsx
Inbal-Tish: Currently, `disableWeekScroll` prop makes the calendar not really functional. This is a regression we should address here.

src/expandableCalendar/WeekCalendar/index.tsx
Inbal-Tish: Sure, only that `viewableItems` is an array that can contain the current week (depends on the viewability config), and since the weeks are actually marked as day dates it is possible that the `newDate` is different from the `visibleWeek.current` (One is Sunday and the other Wednesday) and still on the same week.

Inbal-Tish:
@yuvalsho When passing `disableWeekScroll` prop the calendar not working properly.

src/expandableCalendar/week.tsx
```diff
@@ -1,33 +1,51 @@
...
+    ...prevProps.context,
+  };
+
+  const nextPropsToCheck = {
+    ...nextProps.markedDates,
+    ...nextProps.context,
+  };
 
-const Week = (props: WeekProps) => {
+  return isEqual(prevPropsToCheck, nextPropsToCheck);
```
Inbal-Tish, line 31: What if these props are equal but some other prop changed, like 'firstDay'? It will still return true and there won't be an update to the component.

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,221 +1,236 @@
...
 import Week from '../week';
-import Presenter from './presenter';
+import {UpdateSources} from '../commons';
 import constants from '../../commons/constants';
+import {extractCalendarProps} from '../../componentUpdater';
+import CalendarContext from '../Context';
 
-const NUMBER_OF_PAGES = 2; // must be a positive number
-const applyAndroidRtlFix = constants.isAndroid && constants.isRTL;
+const NUMBER_OF_PAGES = 6; // must be a positive number
```
Inbal-Tish, line 18:
```suggestion
const NUMBER_OF_PAGES = 6;
```

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,221 +1,236 @@
...
+  const {date, numberOfDays, updateSource, setDate, timelineLeftInset} = context;
+  const visibleWeek = useRef(date);
+  const style = useRef(styleConstructor(theme));
+  const items = useRef<string[]>(getDatesArray(current ?? date, firstDay, numberOfDays));
+  const [listData, setListData] = useState(items.current);
+  const changedItems = useRef(constants.isRTL);
+  const list = useRef<FlatList>(null);
+  const currentIndex = useRef(NUMBER_OF_PAGES);
+
+  useEffect(() => {
```
Inbal-Tish, line 51: Do we need to do this on load as well or just on update? Maybe consider changing to `useDidUpdate` hook instead...

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -1,221 +1,236 @@
...
-          initialScrollIndex={NUMBER_OF_PAGES}
-          getItemLayout={this.getItemLayout}
-          onScroll={this.onScroll}
-          onMomentumScrollEnd={this.onMomentumScrollEnd}
-        />
+  const containerStyle = useMemo(() => {
+    return [style.current.week, style.current.weekCalendar];
+  }, []);
+
+  const flatListContainerStyle = useMemo(() => {
```
Inbal-Tish, line 123: Why are you memoizing here? This is just a style from a StyleShhet...

src/expandableCalendar/WeekCalendar/index.tsx
yuvalsho: i tested it many times, using sameWeek will only cause issues when passing number of days (as one week can have 2/3 different items

Inbal-Tish: oh ok. Got it. Thanks

src/expandableCalendar/week.tsx
```diff
@@ -68,20 +72,13 @@ const Week = (props: WeekProps) => {
...
-    return [style.current.partialWeek, {paddingLeft: timelineLeftInset}];
-  }, [timelineLeftInset]);
-
   return (
     <View style={style.current.container}>
       <View style={[style.current.week, numberOfDays > 1 ? partialWeekStyle : undefined, propsStyle]}>{renderWeek()}</View>
     </View>
   );
-};
+}, isEqual);
```
Inbal-Tish, line 80: I see you removed your deep comparison of the `merkedDates` and `context` props, but 'isEqual' will not deep check them now and will only compaier their ref (it will deep check prop object only). Is this what you mean to do?

yuvalsho: according to the docs isEqual is deep comparison
Screen Shot 2022-09-12 at 10 05 30

src/expandableCalendar/week.tsx
```diff
@@ -17,6 +17,13 @@ export type WeekProps = CalendarProps & {
   context?: CalendarContextProps;
 };
 
+function arePropsEqual(prevProps: WeekProps, nextProps: WeekProps) {
+  const {context: prevContext, markedDates: prevMarkings, ...prevRest} = prevProps;
+  const {context: nextContext, markedDates: nextMarkings, ...nextRest} = nextProps;
+
+  return isEqual(prevContext, nextContext) && isEqual(prevMarkings, nextMarkings) && isEqual(prevRest, nextRest);
```
Inbal-Tish, line 24: Yes. That's good. Just a rename as we use 'others' and not 'rest', so `prevOthers` and `nextOthers` or simply `prevProps` and `nextProps`.

Inbal-Tish (approved):

Inbal-Tish:
@yuvalsho Hi. Can you please resolve conflict and let's merge this?

PR #1962, opened 2022-08-07
improve week component performance

yuvalsho:


src/expandableCalendar/week.tsx
```diff
@@ -55,23 +53,16 @@ const Week = (props: WeekProps) => {
 
   const renderWeek = () => {
     const dates = numberOfDays > 1 ? getPartialWeekDates(current, numberOfDays) : getWeek(current);
-    let week: any[] = [];
-  
+    const week: JSX.Element[] = [];
+    const todayIndex = dates?.indexOf(parseDate(new Date())) || -1;
```
Inbal-Tish, line 57: You can move `todayIndex` inside the if (dates) as you're not using it outside of it

Inbal-Tish (approved):

PR #1970, opened 2022-08-10
Use template string literals for string interpolation

gachez:
Replaced current string interpolation with template string literals for eligibility and best coding practice.


Inbal-Tish:
@gachez Hi. Thanks for the PR. I like where this is going but it feels incomplete... Can you fix that across the lib please?

gachez:
Hello @Inbal-Tish I went ahead and fixed across the entire lib.

PR #1971, opened 2022-08-10
Utils directory for all shared utils

Inbal-Tish:
Don't be alert - many files but 97% only imports path changes.
Move 'MarkedDates' type to `types` and change usage accordingly.
Move `VelocityTracker` to the Agenda directory.
Move all tools to `utils` directory.
Moving 'interface' tools and tests to 'dateutils' and removing 'interface' files.


src/utils/dateutils.ts
```diff
@@ -1,6 +1,4 @@
 const XDate = require('xdate');
-const {parseDate, toMarkingFormat} = require('./interface');
-const {getDefaultLocale} = require('./services');
```
Inbal-Tish, line 3: To avoid required cycle

Inbal-Tish:
error will be fixed when merging : https://github.com/wix/react-native-calendars/pull/1973

PR #1972, opened 2022-08-11
Replaced some if else statements with ternary operators

gachez:
So this is a small one. All I did was replace some if else statements with conditional ternary operator to avoid redanduncy and also renamed a variable to describe itself correctly

### DIFF
--- a/src/calendar/index.js
+++ b/src/calendar/index.js
@@ -279,11 +279,7 @@ class Calendar extends Component {
     }
 
     const dates = this.props.markedDates[day.toString('yyyy-MM-dd')] || EmptyArray;
-    if (dates.length || dates) {
-      return dates;
-    } else {
-      return false;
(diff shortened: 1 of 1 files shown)


src/calendar/index.js
```diff
@@ -335,11 +331,10 @@ class Calendar extends Component {
       week.push(this.renderDay(day, id2));
     }, this);
 
-    if (this.props.showWeekNumbers) {
-      week.unshift(this.renderWeekNumber(days[days.length - 1].getWeek()));
-    }
-
-    return (<View style={this.style.week} key={id}>{week}</View>);
+    this.props.showWeekNumbers ?
```
Inbal-Tish, line 334: This one doesn't work. You need to return the view with the weeks, there is no if-else here and you're missing a return statement

PR #1979, opened 2022-08-11
dateutils - parseDate - reduce usage

Inbal-Tish:
dateutils - parseDate - reduce usage


ethanshar (approved):
Approved
But why do you prefer to use XDate directly instead of parseDate? 

Technically, if you use something like parseDate it encapsulated XDate and gives you the option to change XDate in a single place.

Inbal-Tish:
> Approved But why do you prefer to use XDate directly instead of parseDate?
> 
> Technically, if you use something like parseDate it encapsulated XDate and gives you the option to change XDate in a single place.

If I know that I can use XDate directly, like in the case of date strings, I want to avoid the multiple checks that the parseDate() is going through to figure out the type and determine how to turn it to XDate. In some cases, parseDate will be called many times and I'll save accumulated time on its checks.

ethanshar:
👍

PR #1980, opened 2022-08-11
ExpandableCalendar - fix - day set when arrow press while closed

Inbal-Tish:
ExpandableCalendar - fix - when calendar closed arrow press doesn't set the first day of the week (sets the dame day, e.g. Wednesday)


lidord-wix (commented):
This fixes the arrow issue, but now I can't press on a different day in the same week or open the calendar using the knob

Inbal-Tish:
> This fixes the arrow issue, but now I can't press on a different day in the same week or open the calendar using the knob

Thanks! fixed.

lidord-wix (approved):

PR #1981, opened 2022-08-14
ExpandableCalendar - day selection

Inbal-Tish:
ExpandableCalendar - selection made from Day.state prop instead of `markedDates`

### DIFF
--- a/src/expandableCalendar/WeekCalendar/index.tsx
+++ b/src/expandableCalendar/WeekCalendar/index.tsx
@@ -138,7 +138,7 @@ class WeekCalendar extends Component<WeekCalendarProps, State> {
 
   renderItem = ({item}: any) => {
     const {allowShadow, context, ...calendarListProps} = this.props;
-    const {style, onDayPress = this.onDayPress, markedDates, firstDay = 0, ...others} = extractCalendarProps(calendarListProps);
+    const {style, onDayPress = this.onDayPress, firstDay = 0, ...others} = extractCalendarProps(calendarListProps);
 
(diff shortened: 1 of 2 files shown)


src/expandableCalendar/index.tsx
```diff
@@ -172,6 +166,11 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
   const numberOfWeeks = useRef(getNumberOfWeeksInMonth(date));
 
   /** Position */
+  const [position, setPosition] = useState(initialPosition);
+  
+  const isOpen = useMemo(() => {
+    return position === Positions.OPEN;
+  }, [position]);
```
ethanshar, line 173: No need to memoize boolean values

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -150,7 +150,6 @@ class WeekCalendar extends Component<WeekCalendarProps, State> {
         current={item}
         firstDay={firstDay}
         style={this.getWeekStyle(this.containerWidth, style)}
-        markedDates={markedDates}
```
ethanshar, line 153: I see u remove markDates completely, don't we still need them for general markings?

Inbal-Tish: They will be passed in `...others`

src/expandableCalendar/index.tsx
Inbal-Tish: Fixed

ethanshar (approved):

PR #1983, opened 2022-08-15
Calendar header - fix day names style

Inbal-Tish:
Following PR: #1885 There is a bug in the Calendar and CalendarList header style - extra horizontal padding for the day names.
Also, cleaned the timeline screen and moved stuff to mock file.

### DIFF
--- a/src/calendar-list/style.ts
+++ b/src/calendar-list/style.ts
@@ -30,7 +30,8 @@ export default function getStyle(theme: Theme = {}) {
       left: 0,
       right: 0,
       top: 0,
-      backgroundColor: appStyle.calendarBackground
+      backgroundColor: appStyle.calendarBackground,
+      paddingHorizontal: 15
     },
     // @ts-expect-error
     ...(theme['stylesheet.calendar-list.main'] || {})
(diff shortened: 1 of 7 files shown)


src/calendar-list/style.ts
```diff
@@ -30,7 +30,8 @@ export default function getStyle(theme: Theme = {}) {
       left: 0,
       right: 0,
       top: 0,
-      backgroundColor: appStyle.calendarBackground
+      backgroundColor: appStyle.calendarBackground,
+      paddingHorizontal: 15
```
lidord-wix, line 34: This extra padding breaks the layout of the timeline calendar - try to uncomment the `numberOfDays={3}` in the `timelineCalendarScreen.tsx` and see the padding of the days' names

lidord-wix: Also, the days' names don't change when moving to the next 3 days

Inbal-Tish: Header alignment fixed

Inbal-Tish: There is an issue with the dates but I do see the day names are changing...

lidord-wix (approved):

PR #1989, opened 2022-08-17
ExpandableCalendar - fix fast list navigation

Inbal-Tish:
Fix for fast list navigation - scroll / arrow press:
1) Separate update source for user scroll (PAGE_SCROLL) and arrow press that invokes list scroll (ARROW_PRESS).
2) When context.date changes, only scroll when the update source is ARROW_PRESS.
3) On PAGE_SCROLL - debounce setDate call.
4) On ARROW_PRESS - throttle auto list scroll.


src/expandableCalendar/index.tsx
```diff
@@ -242,6 +242,10 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
     return {height: deltaY.current};
   }, [deltaY.current]);
 
+  const numberOfDaysCondition = useMemo(() => {
+    return !numberOfDays || numberOfDays && numberOfDays <= 1; 
+  }, [numberOfDays]);
```
ethanshar, line 247: No need to memoize boolean values

src/expandableCalendar/index.tsx
```diff
@@ -255,29 +259,35 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
...
-  }, [date]);
-
   const handleScreenReaderStatus = (screenReaderEnabled: any) => {
     setScreenReaderEnabled(screenReaderEnabled);
   };
 
   /** Scroll */
+  const autoScroll = useRef(false);
+
+  const scrollToDate = throttle((date: string) => {
```
ethanshar, line 269: Creating a throttled function in a function component is tricky. 
The "throttled" function is being recreated each render and basically losing the throttle feature. 
You will need to wrap this with useCallback to preserve the method.

src/expandableCalendar/index.tsx
```diff
@@ -255,29 +259,35 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
...
     }
-  };
+  }, 100, {leading: false, trailing: true});
+
+  /** Update Date */
 
-  const scrollPage = useCallback((next: boolean) => {
+  const newTempDate = useRef(date);
+
+  const _setDate = debounce((updateSource = updateSources.PAGE_SCROLL) => {
```
ethanshar, line 284: Same here, the debounce method should be wrapped with useCallback

src/expandableCalendar/index.tsx
```diff
@@ -255,29 +259,35 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
...
-    // date was changed from AgendaList, arrows or scroll
-    scrollToDate(date);
-  }, [date]);
-
   const handleScreenReaderStatus = (screenReaderEnabled: any) => {
     setScreenReaderEnabled(screenReaderEnabled);
   };
 
   /** Scroll */
+  const autoScroll = useRef(false);
```
ethanshar, line 267: Move to the top of the function, near the other refs

src/expandableCalendar/index.tsx
```diff
@@ -255,29 +259,35 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
...
 
-  const scrollToDate = (date: string) => {
     if (!horizontal) {
       calendarList?.current?.scrollToDay(date, 0, true);
     } else if (getYear(date) !== visibleYear.current || getMonth(date) !== visibleMonth.current) {
       // don't scroll if the month is already visible
       calendarList?.current?.scrollToMonth(date);
     }
-  };
+  }, 100, {leading: false, trailing: true});
```
ethanshar, line 278: I think it's better to set leading to true and increase the throttle time. So the first press will take into account and quick presses after that will be throttled.

src/expandableCalendar/index.tsx
```diff
@@ -255,29 +259,35 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
...
+
+  /** Update Date */
 
-  const scrollPage = useCallback((next: boolean) => {
+  const newTempDate = useRef(date);
```
ethanshar, line 282: This may lead to bugs. 
By creating a fork of the date value here we're losing the single source of truth principal. 

I see you update its value in `updatePage`, but technically the date value can be changed in many other places. 
This will probably lead to date issues in the future. 

I know we did something similar in `CalendarProvider`, which I'm also not a fan of, but at least the provider is where we maintain the original date value.
Once more sub components start to fork this value, we'll have a mess. 

If the reason you need this is for the debouncing the setDate, consider providing a `setDateDebounced` method from the context.
Components that need an immediate update will use `setDate`, others can use `setDateDebounced` 
In any case, it will be maintained by the provider

src/expandableCalendar/index.tsx
Inbal-Tish: Done

src/expandableCalendar/index.tsx
Inbal-Tish: The refs are placed by related section, not only on top. I thought it will be more readable this way...

src/expandableCalendar/index.tsx
Inbal-Tish: Done

src/expandableCalendar/index.tsx
Inbal-Tish: Done

src/expandableCalendar/index.tsx
Inbal-Tish: Done

src/expandableCalendar/index.tsx
Inbal-Tish: Yes, I'm using the `newTempDate` ref to debounce the `setDate`. I need the date I'm passing `setDate` to be updated according to the new page without relaying on the date change by the`setDate` (context.date). It's a way to keep an updated date that I can increment/decrement (and scroll to it) even when I'm not really setting the new date to the context. I agree it can lead to bugs but I think we do use it often when we want to keep updated value without updating the state (and rerendering as a result). Even if I'll add a `debouncedSetDate` on the provider I will still need to keep the `newTempDate` ref locally as I need it for scrolling and moving the list items.

ethanshar: But isn't it better to let the Provider handle it? 
Let's say another component will need to setDate with debounce, will we implement the logic again? 

Also, there's already a bug. I assume it's because you now maintain another copy of the date value. 

In the Expandable Calendar example. Scroll few weeks forward with gestures (no arrows), now scroll forward with the arrows. The date goes back, the dates are not synced.

ethanshar (changes requested):

src/expandableCalendar/index.tsx
Inbal-Tish: There is an issue with the WeekCalendar. I'll wait for the new implementation then...

Inbal-Tish: There is an issue with the WeekCalendar. I'll wait for the new implementation then...

src/expandableCalendar/index.tsx
ethanshar: I see. 
I mention it before, but you should consider moving each section's code to a hook. 
You'll see how it makes the code even more readable 
You can have useScrollHandlers, useGesturesHandler, etc.. 
Anyway, not related to this PR

src/expandableCalendar/index.tsx
ethanshar: But if the issue is caused by the new date temp ref, why not fix that?

ethanshar (commented):
@Inbal-Tish 
One more thing. 
Moving a single month forward/backward with a scroll/press has a feel of a delayed effect. 
The month scrolls fine, but it take a "long" time till the new day of the month is selected, it feels weird. Maybe we need to fine tune the call of setDate

src/expandableCalendar/index.tsx
Inbal-Tish: I'm not sure it's related

PR #1995, opened 2022-08-21
CalendarList - fix - update months doesn't apply to the initial month

Inbal-Tish:
CalendarList - fix - update months doesn't apply to the initial month

### DIFF
--- a/src/calendar-list/index.tsx
+++ b/src/calendar-list/index.tsx
@@ -161,8 +161,10 @@ const CalendarList = (props: CalendarListProps, ref: any) => {
       }
     }
 
-    // @ts-expect-error
-    list?.current?.scrollToOffset({offset: scrollAmount, animated});
+    if (scrollAmount != 0) {
+      // @ts-expect-error
+      list?.current?.scrollToOffset({offset: scrollAmount, animated});
+    }
   };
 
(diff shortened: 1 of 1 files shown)


src/calendar-list/index.tsx
```diff
@@ -171,12 +168,10 @@ const CalendarList = (props: CalendarListProps, ref: any) => {
   const scrollToMonth = useCallback((date: XDate | string) => {
     const scrollTo = parseDate(date);
     const diffMonths = Math.round(initialDate?.current?.clone().setDate(1).diffMonths(scrollTo?.clone().setDate(1)));
-    if (diffMonths !== 0) {
-      const scrollAmount = calendarSize * pastScrollRange + diffMonths * calendarSize;
-      
-      // @ts-expect-error
-      list?.current?.scrollToOffset({offset: scrollAmount, animated: animateScroll});
-    }
+    const scrollAmount = calendarSize * pastScrollRange + diffMonths * calendarSize;
```
ethanshar, line 171: scrolling 0 pixels is still a scroll action and can be heavy.
I think it's better to avoid the scroll all together if the scrollAmount is 0

Inbal-Tish: Agree. Fixed

ethanshar (approved):

PR #1996, opened 2022-08-21
[Don't Merge] testIDs refactor

ethanshar:
**Don't merge yet, just review**
Change all nested testIDs
Fixed existing tests and drivers
Added tests to expandable knob behavior


src/calendar-list/index.tsx
```diff
@@ -216,13 +216,16 @@ const CalendarList = (props: CalendarListProps, ref: any) => {
     return false;
   }, [currentMonth]);
 
-  const renderItem = useCallback(({item}: any) => {
+  const renderItem = useCallback(({item}: {item: XDate}) => {
+    const dateString = toMarkingFormat(item);
+    const [year, month] = dateString.split('-');
```
Inbal-Tish, line 221: If `item` is of XDate type you don't need to move it to be a string and split it, you can just use its methods, like `item.getFullYear()` and `item.getMonth()`.

src/calendar-list/index.tsx
```diff
@@ -216,13 +216,16 @@ const CalendarList = (props: CalendarListProps, ref: any) => {
...
+    const _testID = `${testID}.item_${year}-${month}`;
```
Inbal-Tish, line 222: const in a function scope are not supposed to be marked with an underscore... you can simply name it `testId` instead.

src/calendar/index.tsx
```diff
@@ -268,11 +269,11 @@ const Calendar = (props: CalendarProps) => {
     const ref = customHeader ? undefined : header;
     const CustomHeader = customHeader;
     const HeaderComponent = customHeader ? CustomHeader : CalendarHeader;
-    
+
     return (
       <HeaderComponent
         {...headerProps}
-        testID={testID}
+        testID={`${testID}.header`}
```
Inbal-Tish, line 276: This is what might collide with the CalndarList's static header (comment above)

src/calendar-list/item.tsx
```diff
@@ -30,7 +30,7 @@ const CalendarListItem = React.memo((props: CalendarListItemProps) => {
     headerStyle,
     onPressArrowLeft,
     onPressArrowRight,
-    visible
+    visible,
```
Inbal-Tish, line 33: Why the extra comma?

src/calendar-list/index.tsx
```diff
@@ -239,7 +242,7 @@ const CalendarList = (props: CalendarListProps, ref: any) => {
       return (
         <CalendarHeader
           {...headerProps}
-          testID={STATIC_HEADER}
+          testID={`${testID}.header`}
```
Inbal-Tish, line 245: This id might cause duplications, as it is an additional header on top of the Calendar header (CalendarListItem). We should consider calling it `staticHeader` or have a default value to the `testID` prop as "calendarList".

src/calendar/index.tsx
```diff
@@ -194,7 +192,7 @@ const Calendar = (props: CalendarProps) => {
           marking={weekNumberMarking.current}
           // state='disabled'
           theme={theme}
-          testID={`${WEEK_NUMBER}-${weekNumber}`}
+          testID={`${testID}.week_${weekNumber}`}
```
Inbal-Tish, line 195: Why is it called "week" and not "week_number"?

src/expandableCalendar/week.tsx
```diff
@@ -36,14 +50,16 @@ const Week = (props: WeekProps) => {
...
       }
     }
+    const dayString = toMarkingFormat(day);
 
     return (
       <View style={style.current.dayContainer} key={id}>
         <Day
           {...dayProps}
-          date={toMarkingFormat(day)}
+          testID={`${testID}_week_${dayString}`}
```
Inbal-Tish, line 59: The testID pattern used so far is `${testID}**.**week_${dayString}`. Also, why is it named "week" and not "week_day"?

src/expandableCalendar/index.tsx
```diff
@@ -581,15 +583,15 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
...
+          testID={`${testID}.calendarAccessible`}
           {...others}
           theme={themeObject}
           onDayPress={_onDayPress}
           hideExtraDays
           renderArrow={_renderArrow}
         />
       ) : (
-        <Animated.View ref={wrapper} style={wrapperStyle} {...panResponder.panHandlers}>
+        <Animated.View testID={`${testID}.expandableHeader`} ref={wrapper} style={wrapperStyle} {...panResponder.panHandlers}>
```
Inbal-Tish, line 594: This is not a header... This is the "expandable" calendar/wrapper

Inbal-Tish: The header for the vertical state is the animated header in `renderAnimatedHeader()`. Maybe worth adding a testID there as well...

src/calendar/header/index.tsx
```diff
@@ -207,7 +199,7 @@ const CalendarHeader = forwardRef((props: CalendarHeaderProps, ref) => {
         <Text
           allowFontScaling={false}
           style={style.current.monthText}
-          testID={testID ? `${HEADER_MONTH_NAME}-${testID}` : HEADER_MONTH_NAME}
+          testID={`${testID}.label`}
```
Inbal-Tish, line 202: I would call it "title" instead of "label"...

src/expandableCalendar/week.tsx
ethanshar: This is actually your change, not mine. 
I'll fix it

src/expandableCalendar/driver.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (!element) throw 'Element is missing';
+    this.renderTree = render(element);
+    return this.renderTree;
+  }
+
+  getKnob() {
+    return this.renderTree?.getByTestId(`${this.testID}.knob`);
+  }
+
+  getExpandableHeader() {
```
Inbal-Tish, line 34: Please change the name of the method as well

src/expandableCalendar/driver.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  getDay(date: string) {
+    return this.renderTree?.getByTestId(this.getDayTestID(date));
+  }
+
+  toggleKnob() {
+    fireEvent(this.getKnob(), 'onPress');
+  }
+
+  isHeaderExpanded() {
```
Inbal-Tish, line 46: `isCalendarExpanded`...

src/expandableCalendar/driver.ts
```diff
@@ -0,0 +1,54 @@
...
+    const headerHeight = this.getExpandableHeader().props?.style?.height;
```
Inbal-Tish, line 47: `calendarHeight`

src/expandableCalendar/week.tsx
ethanshar: Hmm now the testIDs are like this 
for calendarList
`expandableCalendar.calendarList.item_2022-08.day_2022-09-01`
for weekCalendar
`expandableCalendar.weekCalendar.day_2022-09-10`

Inbal-Tish:
@ethanshar Awsome! Small comments...

src/calendar/index.tsx
ethanshar: Because it's a week_[number], it's like I do with day_[date]
Let me know if you still prefer weekNumber

src/expandableCalendar/week.tsx
Inbal-Tish: Make sense, except the calendarList item's date doesn't match the day date...

src/calendar-list/index.tsx
ethanshar: Fixed

src/calendar/index.tsx
Inbal-Tish: I prefer "week_number" or "weekNumber" since we have a `Week` component and because it's actually on a day component, so it can be confusing.

src/calendar-list/item.tsx
ethanshar: Fixed

src/calendar/index.tsx
ethanshar: I fixed the static header id, here we'll keep it as header

src/calendar-list/index.tsx
ethanshar: > const in a function scope are not supposed to be marked with an underscore
Why? 

Do you prefer to have a const that called `testId` and not `_testID`?
The former is more confusing IMO, moreover we usually use `_` prefix for internal variables with a similar name to a prop/argument, it's more explicit and easy to spot.

src/calendar-list/index.tsx
ethanshar: I did it at first, but the `getMonth` returns a zero based month number so I had to add +1 and it also without the zero padding..
So at the end it was just cleaner to use markingFormat

src/calendar/index.tsx
Inbal-Tish: Yes. That's what I meant. Thanks

src/calendar-list/index.tsx
Inbal-Tish: Yes, I prefer `testId`. 
The prefix `_` is used for private (class) variables, inside the function they are all private, so it's wired...

src/calendar/index.tsx
ethanshar: Fair enough, fixed

src/calendar-list/index.tsx
Inbal-Tish: Is it possible this is the reason you get testID with `item_**month-year**.day_**date**` that don't match?

src/expandableCalendar/week.tsx
ethanshar: That's on purpose, the calendar list item represents a year and a month, not a day. 
The day tends to change because it's based on the `current` value, it can be the first day sometimes or a the mid day of the month. 
When a user wants to fetch a specific calendar item, they should fetch it by year+month

Inbal-Tish: I meant this: `expandableCalendar.calendarList.item_**2022-08**.day_**2022-09**-01`

src/calendar-list/index.tsx
ethanshar: What do you don't match? 
All the tests pass. 

Not sure I follow

src/calendar-list/index.tsx
ethanshar: I will change. 
just FYI, we used it all the time for function components so it's not that different

src/expandableCalendar/week.tsx
ethanshar: Is it possible it's because we render days from nearby months as disabled days?

Inbal-Tish: Do you mean because of the extra days? mmm... if one of these days is selected the calendar is moved to that day's month, and if it is the 'current' it should show that day's month... I think haha

src/calendar-list/index.tsx
Inbal-Tish: Yes, we do, in the Component scope (which can be a function, for sure...)

src/calendar-list/index.tsx
Inbal-Tish: I meant this: expandableCalendar.calendarList.item_**2022-08**.day_**2022-09**-01 - the day is in September and the item is in August

Inbal-Tish:
@ethanshar Can we remove `src/testIDs.js` file?

ethanshar:
> @ethanshar Can we remove `src/testIDs.js` file?

I see it's still used in some places, so for now we can't

Inbal-Tish (approved):

PR #1999, opened 2022-08-24
Move 'TodayButton' to a separate component

Inbal-Tish:
Moving all TodayButton's code from the Context provider to a separate component (and deleting the presenter).


src/expandableCalendar/Context/Provider.tsx
```diff
@@ -158,18 +123,15 @@ const CalendarProvider = (props: CalendarContextProviderProps) => {
...
-        </TouchableOpacity>
-      </Animated.View>
+      <TodayButton
+        state={isToday(currentDate) ? 0 : isPastDate(currentDate) ? -1 : 1}
+        disabled={disabledButton}
+        disabledOpacity={disabledOpacity}
+        onPress={onTodayPress}
+        margin={todayBottomMargin}
+        style={todayButtonStyle}
+        theme={theme}
```
ethanshar, line 133: Consider moving ALL of the logic to TodayButton. 
You can listen to the context inside the component and handle everything from there. 
It will make the Provider code way more clearer and simple

ethanshar (commented):
Great change. 
The more we break big components into small ones it will easier to maintain and refactor the code. 

See my comment. I think we can move more logic to the TodayButton code.

src/expandableCalendar/Context/Provider.tsx
Inbal-Tish: Not sure I understand what logic you mean... Can you be more specific?

ethanshar: - The `state` prop. The TodayButton can listen to currentDate and handle it internally
- Same for the onPress
- I was thinking also about the `disabled` functionality, but I see you include it in the context

Inbal-Tish: So you're saying basically to pass the TodayButton the context, right?

ethanshar: you don't need to pass it, you can consume it with `useContext` inside the TodayButton code

ethanshar (approved):

PR #2008, opened 2022-08-26
Update README.md

Jinho1011:
Remove renderList from Agenda props beause it is not available


Inbal-Tish:
@Jinho1011 What do you mean "it's not available"?

PR #2015, opened 2022-09-01
Fix/theme issues

dlariosuniandes:
Reverted props extraction as they seem to mess with the Theme for the Agenda

### DIFF
--- a/src/calendar-list/index.tsx
+++ b/src/calendar-list/index.tsx
@@ -5,7 +5,7 @@ import XDate from 'xdate';
 import React, {forwardRef, useImperativeHandle, useRef, useEffect, useState, useCallback, useMemo} from 'react';
 import {FlatList, View, ViewStyle, FlatListProps} from 'react-native';
 
-import {extractHeaderProps, extractCalendarProps} from '../componentUpdater';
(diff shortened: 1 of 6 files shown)


dlariosuniandes:
Hi is there any way I can test this build ?

bakwas247:
Yeah, if you compile the files using "tsc", then use the newly emitted files as a node module, you can test it. They also already have a set of tests prepared, that the docs tell you how to run.

Inbal-Tish:
@dlariosuniandes Please describe the bug you wish to solve. Reverting to the previous usage of `extractComponentProps` is not recommended for performance. We'll need to find how to fix `extractHeaderProps` instead, so please add a description of the bug and a code snippet so we can try to reproduce it.

Saikedo:
Are there any plans to address this ? Currently it is impossible to modify a lot of the Theme props for the agenda.

Inbal-Tish:
@Saikedo Hi. See the PR #2041

Saikedo:
@Inbal-Tish Seems like `backgroundColor` is still not working after the PR #2041.

I managed to make it work with `reservationsBackgroundColor` but I am not sure if this is appropriate since `reservationsBackgroundColor` is not even listed in the Theme.

Am I missing something? Is there a better way of handling this?

PR #2028, opened 2022-09-12
fixed: ReservationList.theme no work

akakaze:
Need to pass prop theme down

### DIFF
--- a/src/componentUpdater.ts
+++ b/src/componentUpdater.ts
@@ -209,6 +209,7 @@ export function extractReservationListProps(props: AgendaProps) {
   const {
     // ReservationList props
     items,
+    theme,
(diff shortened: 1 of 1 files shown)


walteribeiro (approved):
I need this too 👍🏻

src/componentUpdater.ts
```diff
@@ -209,6 +209,7 @@ export function extractReservationListProps(props: AgendaProps) {
   const {
     // ReservationList props
     items,
+    theme,
```
Inbal-Tish, line 212: You're extracting it but not passing it...

Inbal-Tish:
@akakaze Please see fix on PR #2041

Issue #2017, opened 2022-09-02
Agenda not rendering marks and custom colors


andrelramos:
## Description

Marks and custom colors not rendering in last release version ([1.1288.2](https://github.com/wix/react-native-calendars/releases/tag/1.1288.2)).

The last version working with custom colors was [1.1287.0](https://github.com/wix/react-native-calendars/releases/tag/v1.1287) and working with marks was [1.1285.0](https://github.com/wix/react-native-calendars/releases/tag/v1.1285.0).

## Expected Behavior

Show day marks on Agenda's calendar and custom color for today and selected day

## Observed Behavior

The Agenda is ignoring these parameters and rendering like default Agenda

## Environment

* `npm ls react-native-calendars`: react-native-calendars@1.1288.2
* `npm ls react-native`: react-native@0.64.3
* `npm ls expo`:expo@44.0.6

Also specify:

1. Device/emulator/simulator & OS version: iPhone 8 Plus emulator (iOS 15.5)

## Reproducible Demo

Code to reproduce:
```
<Agenda 
        items={{
          "2022-09-06": [
            {
              "day": "2022-09-06",
              "height": 75,
              "name": "2661",
            },
          ]
        }}
        markingType={"multi-dot"}
        markedDates={{
          "2022-09-06": {
            "dots": [
              {
                "color": "grey",
                "key": "Waiting",
              },
              {
                "color": "green",
                "key": "Confirmed",
              }
            ]
          }
        }}
        theme={{
          selectedDayBackgroundColor: Colors.purple,
          todayTextColor: Colors.purple,
        }}
        loadItemsForMonth={loadItems}
        renderItem={renderItem}
        renderEmptyDate={() => <NothingToShowText>Nenhum agendamento disponível</NothingToShowText>}
      />
``` 

## Screenshots

The code above using version [1.1285.0](https://github.com/wix/react-native-calendars/releases/tag/v1.1285.0): 

https://user-images.githubusercontent.com/10236206/188246927-045354c9-3020-4d7e-a82c-f0d912ff0c68.mov

Using version [1.1288.2](https://github.com/wix/react-native-calendars/releases/tag/1.1288.2):

https://user-images.githubusercontent.com/10236206/188246991-bd28accd-1c52-4098-a42a-a16c3c0de04f.mov

Saikedo:
Facing the same issue. I used to provide a theme to agenda like this

```
const agendaTheme: Theme = useMemo(() => {
        return {
[...]

PR #2041, opened 2022-09-26
Added fix for Agenda theme prop bug

harshsiri110:
Fix: #2039 

This PR solves the issue of changing the date color in Agenda items via props.

Change date color in Agenda items fix PR

PR #2042, opened 2022-09-26
Fixed Agenda marks and custom colors not rendering issue

mohitcharkha:
Fix: #2017

This PR fixes Agenda not rendering marks and custom colors issue.

### DIFF
--- a/src/componentUpdater.ts
+++ b/src/componentUpdater.ts
@@ -234,6 +234,7 @@ export function extractReservationListProps(props: AgendaProps) {
     renderDay,
     renderItem,
     renderEmptyDate,
+    ...others
   } = props;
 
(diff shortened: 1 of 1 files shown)


src/componentUpdater.ts
```diff
@@ -234,6 +234,7 @@ export function extractReservationListProps(props: AgendaProps) {
     renderDay,
     renderItem,
     renderEmptyDate,
+    ...others
```
Inbal-Tish, line 237: No. `...others` contains all the props we don't want to pass. By passing `others` you're actually passing ALL the props.

Inbal-Tish:
@mohitcharkha Please see fix in PR #2041

PR #2060, opened 2022-10-11
Fix week days cut off in Portuguese and Spanish

adids1221:
Fix weekdays cut off in Portuguese and Spanish.
Jira: 3101

### DIFF
--- a/src/calendar/header/index.tsx
+++ b/src/calendar/header/index.tsx
@@ -184,7 +184,7 @@ const CalendarHeader = forwardRef((props: CalendarHeaderProps, ref) => {
       }
 
       return (
-        <Text allowFontScaling={false} key={index} style={dayStyle} numberOfLines={1} accessibilityLabel={''}>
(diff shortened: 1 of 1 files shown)


src/calendar/header/index.tsx
```diff
@@ -184,7 +184,7 @@ const CalendarHeader = forwardRef((props: CalendarHeaderProps, ref) => {
       }
 
       return (
-        <Text allowFontScaling={false} key={index} style={dayStyle} numberOfLines={1} accessibilityLabel={''}>
+        <Text allowFontScaling={false} key={index} style={[dayStyle, {flex: 1}]} numberOfLines={1} accessibilityLabel={''}>
```
Inbal-Tish, line 187: This is bad for performance

Inbal-Tish: The `dayStyle` defined the width to be 32, flexing them can result in miss-alignment as they need to be alight to the day numbers

PR #2080, opened 2022-11-07
Timeline calendar num of days issues

yuvalsho:
fixed issues:
1) selected day marking appears as opposed to the guideline which doesn't show the selected date
2) when scrolling week from the timeline sometimes it doesn't change the date
3) when updating numOfSays the component doesn't re-render properly

### DIFF
--- a/src/dateutils.ts
+++ b/src/dateutils.ts
@@ -23,7 +23,7 @@ export function sameDate(a?: XDate, b?: XDate) {
   }
 }
 
-export function onSameDAteRange({
+export function onSameDateRange({
   firstDay,
   secondDay,
   numberOfDays,
--- a/src/dateutils.ts
+++ b/src/dateutils.ts
@@ -23,10 +23,20 @@ export function sameDate(a?: XDate, b?: XDate) {
   }
 }
 
-export function sameWeekRange(a: string, b: string, numberOfDays: number, firstDateInWeek: string){
-  const aDate = new XDate(a);
-  const bDate = new XDate(b);
-  const firstDayDate = new XDate(firstDateInWeek);
+export function sameDayRange({
+  firstDay,
+  secondDay,
+  numberOfDays,
+  firstDateInRange,
+}: {
+  firstDay: string;
+  secondDay: string;
+  numberOfDays: number;
+  firstDateInRange: string;
+}){
+  const aDate = new XDate(firstDay);
+  const bDate = new XDate(secondDay);
+  const firstDayDate = new XDate(firstDateInRange);
   const aDiff = aDate.getTime() - firstDayDate.getTime();
   const bDiff = bDate.getTime() - firstDayDate.getTime();
   const aTotalDays = Math.ceil(aDiff / (1000 * 3600 * 24));
--- a/src/dateutils.ts
+++ b/src/dateutils.ts
@@ -23,7 +23,7 @@ export function sameDate(a?: XDate, b?: XDate) {
   }
 }
 
-export function sameDayRange({
+export function onSameDAteRange({
   firstDay,
   secondDay,
   numberOfDays,
--- a/src/dateutils.ts
+++ b/src/dateutils.ts
@@ -23,10 +23,10 @@ export function sameDate(a?: XDate, b?: XDate) {
   }
 }
 
-export function sameWeekRange(a: string, b: string, numberOfDays: number, firstDayOfWeek: string){
+export function sameWeekRange(a: string, b: string, numberOfDays: number, firstDateInWeek: string){
   const aDate = new XDate(a);
   const bDate = new XDate(b);
-  const firstDayDate = new XDate(firstDayOfWeek);
+  const firstDayDate = new XDate(firstDateInWeek);
   const aDiff = aDate.getTime() - firstDayDate.getTime();
   const bDiff = bDate.getTime() - firstDayDate.getTime();
   const aTotalDays = Math.ceil(aDiff / (1000 * 3600 * 24));
--- a/src/dateutils.ts
+++ b/src/dateutils.ts
@@ -20,7 +20,20 @@ export function sameDate(a?: XDate, b?: XDate) {
     return false;
   } else {
     return a?.getFullYear() === b?.getFullYear() && a?.getMonth() === b?.getMonth() && a?.getDate() === b?.getDate();
-  } 
+  }
+}
+
+export function sameWeekRange(a: string, b: string, numberOfDays: number, firstDayOfWeek: string){
+  const aDate = new XDate(a);
+  const bDate = new XDate(b);
+  const firstDayDate = new XDate(firstDayOfWeek);
+  const aDiff = aDate.getTime() - firstDayDate.getTime();
+  const bDiff = bDate.getTime() - firstDayDate.getTime();
+  const aTotalDays = Math.ceil(aDiff / (1000 * 3600 * 24));
+  const bTotalDays = Math.ceil(bDiff / (1000 * 3600 * 24));
+  const aWeek = Math.floor(aTotalDays / numberOfDays);
+  const bWeek = Math.floor(bTotalDays / numberOfDays);
+  return aWeek === bWeek;
 }
 
 export function sameWeek(a: string, b: string, firstDayOfWeek: number) {
--- a/src/day-state-manager.ts
+++ b/src/day-state-manager.ts
@@ -3,10 +3,10 @@ const {toMarkingFormat} = require('./interface');
 
 
 export function getState(day: XDate, current: XDate, props: any) {
-  const {minDate, maxDate, disabledByDefault, context, disableSelection} = props;
+  const {minDate, maxDate, disabledByDefault, context} = props;
   let state = '';
 
-  if (((context?.date ?? toMarkingFormat(current)) === toMarkingFormat(day))) {
+  if ((context?.date ?? toMarkingFormat(current)) === toMarkingFormat(day)) {
     state = 'selected';
   } else if (isToday(day)) {
     state = 'today';
--- a/src/day-state-manager.ts
+++ b/src/day-state-manager.ts
@@ -3,10 +3,10 @@ const {toMarkingFormat} = require('./interface');
 
 
 export function getState(day: XDate, current: XDate, props: any) {
-  const {minDate, maxDate, disabledByDefault, context} = props;
+  const {minDate, maxDate, disabledByDefault, context, disableSelection} = props;
   let state = '';
 
-  if ((context?.date ?? toMarkingFormat(current)) === toMarkingFormat(day)) {
+  if (!disableSelection && ((context?.date ?? toMarkingFormat(current)) === toMarkingFormat(day))) {
     state = 'selected';
   } else if (isToday(day)) {
     state = 'today';
--- a/src/day-state-manager.ts
+++ b/src/day-state-manager.ts
@@ -2,8 +2,8 @@ const {isToday, isDateNotInRange, sameMonth} = require('./dateutils');
 const {toMarkingFormat} = require('./interface');
 
 
-export function getState(day: XDate, current: XDate, props: any) {
-  const {minDate, maxDate, disabledByDefault, context, disableDaySelection} = props;
+export function getState(day: XDate, current: XDate, props: any, disableDaySelection?: boolean) {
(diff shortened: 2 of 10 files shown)


src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -23,6 +23,7 @@ const APPLY_ANDROID_FIX = constants.isAndroid && constants.isRTL;
 export interface WeekCalendarProps extends CalendarListProps {
   /** whether to have shadow/elevation for the calendar */
   allowShadow?: boolean;
+  disableDaySelection?: boolean;
```
Inbal-Tish, line 26: Is `disableDaySelection` prop only relevant to the Week and WeekCalendar components? Can't it be relevant to the CalendarProps?

src/expandableCalendar/index.tsx
```diff
@@ -221,7 +221,7 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
         paddingRight: isNumber(rightPaddings) ? rightPaddings + 6 : DAY_NAMES_PADDING
       }
     ];
-  }, [calendarStyle, numberOfDays]);
```
Inbal-Tish, line 224: Wasn't this a fix for the style when moving from 3-days to a week?

src/expandableCalendar/week.tsx
```diff
@@ -60,15 +62,15 @@ const Week = React.memo((props: WeekProps) => {
...
+    const markings = markedDates?.[dayString];
     return (
       <View style={style.current.dayContainer} key={id}>
         <Day
           {...dayProps}
           testID={`${testID}.day_${dayString}`}
           date={dayString}
           state={getState(day, currXdate, props)}
-          marking={markedDates?.[dayString]}
+          marking={disableDaySelection ? {...markings, disableTouchEvent: true} : {markings}}
```
Inbal-Tish, line 73: Let's memoize here so we won't hurt performance

src/expandableCalendar/WeekCalendar/index.tsx
yuvalsho: our usage of display day selection is only relevant for weekly display with custom number of days per week, in the display the knob is hidden and you cant toggle between week view and month view

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -52,7 +54,10 @@ const WeekCalendar = (props: WeekCalendarProps) => {
 
   useDidUpdate(() => {
     if (updateSource !== UpdateSources.WEEK_SCROLL) {
-      const pageIndex = items.current.findIndex(item => sameWeek(item, date, firstDay));
+      const pageIndex = items.current.findIndex(
+        item => (numberOfDays && numberOfDays > 1) ?
+          sameWeekRange(item, date, numberOfDays, item) :
```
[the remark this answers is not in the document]
yuvalsho: I should name this better, I pass the date for the first day in the custom week (for example in a 3-day view it would be a 3 days week where the current date is the first day in one of those weeks) because those weeks have custom length we can't rely on any fixed first day of the week

src/expandableCalendar/week.tsx
```diff
@@ -52,6 +55,18 @@ const Week = React.memo((props: WeekProps) => {
...
+    if (!markedDates) {
+      return markedDates;
+    }
+    return Object.keys(markedDates).reduce((acc, day) => {
+      const dayMarking = markedDates?.[day];
+      return {
+        ...acc,
+        [day]: disableDaySelection ? {...dayMarking, disableTouchEvent: true} : dayMarking};
+    }, {} as MarkedDates);
+  }, []);
```
Inbal-Tish, line 68: `markedDates ` and `disableDaySelection` should be in the dep array

src/expandableCalendar/week.tsx
```diff
@@ -52,6 +55,18 @@ const Week = React.memo((props: WeekProps) => {
   const dayProps = extractDayProps(props);
   const currXdate = useMemo(() => parseDate(current), [current]);
 
+  const markings = useMemo(() => {
+    if (!markedDates) {
+      return markedDates;
+    }
+    return Object.keys(markedDates).reduce((acc, day) => {
```
Inbal-Tish, line 62: What's `acc`? Can we name this better?

src/expandableCalendar/index.tsx
```diff
@@ -539,6 +539,7 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
           onDayPress={_onDayPress}
           accessibilityElementsHidden // iOS
           importantForAccessibility={'no-hide-descendants'} // Android
+          disableDaySelection={!!numberOfDays && numberOfDays > 1}
```
Inbal-Tish, line 542: Why this condition can't be in the WeekCalendar instead of passing it from the ExpandableCalendar? This way 'disableDaySelection' won't be a public prop but an internal one, that the WeekCalendar can pass the Week component (which is a private component).

src/expandableCalendar/WeekCalendar/index.tsx
Inbal-Tish: Then it shouldn't be a public prop. See my comment below

src/dateutils.ts
```diff
@@ -20,7 +20,20 @@ export function sameDate(a?: XDate, b?: XDate) {
     return false;
   } else {
     return a?.getFullYear() === b?.getFullYear() && a?.getMonth() === b?.getMonth() && a?.getDate() === b?.getDate();
-  } 
+  }
+}
+
+export function sameWeekRange(a: string, b: string, numberOfDays: number, firstDateInWeek: string){
```
Inbal-Tish, line 26: I would probably name this method `sameDayRange` and the last pharam 'firstDateInRange'

src/expandableCalendar/WeekCalendar/index.tsx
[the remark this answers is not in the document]
Inbal-Tish: Yes, probably a renaming issue - see my comment above

src/expandableCalendar/index.tsx
```diff
@@ -174,12 +174,17 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
     return CLOSED_HEIGHT + (WEEK_HEIGHT * (numberOfWeeks.current - 1)) + (hideKnob ? 12 : KNOB_CONTAINER_HEIGHT) + (constants.isAndroid ? 3 : 0);
   };
   const openHeight = useRef(getOpenHeight());
-  const closedHeight = useRef(CLOSED_HEIGHT + (hideKnob || Number(numberOfDays) > 1 ? 0 : KNOB_CONTAINER_HEIGHT));
+  const closedHeight = useMemo(() => CLOSED_HEIGHT + (hideKnob || Number(numberOfDays) > 1 ? 0 : KNOB_CONTAINER_HEIGHT), [numberOfDays]);
```
Inbal-Tish, line 177: missing dep 'hideKnob'

src/expandableCalendar/index.tsx
```diff
@@ -174,12 +174,17 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
...
 
-  const startHeight = isOpen ? openHeight.current : closedHeight.current;
+  const startHeight = useMemo(() => isOpen ? openHeight.current : closedHeight, [closedHeight]);
```
Inbal-Tish, line 179: missing dep 'isOpen'

src/timeline-list/index.tsx
```diff
@@ -60,21 +60,24 @@ const TimelineList = (props: TimelineListProps) => {
...
+
+    if (updateSource !== UpdateSources.LIST_DRAG) {
+      if (isOutOfRange(datePageIndex)) {
+        updateSource === UpdateSources.DAY_PRESS ? resetPages(date) : resetPagesDebounce(date);
+      } else {
+        scrollToPageDebounce(datePageIndex);
       }
+    }
+    prevDate.current = date;
+  }, [pagesRef.current, updateSource]);
```
Inbal-Tish, line 74: pagesRef.current is not dep...

Inbal-Tish:
@yuvalsho I don't understand how the changes in Expandable/index regarding the height and closed mechanized relate to this PR? Are there other issue you're fixing here?

src/timeline-list/index.tsx
yuvalsho: why not? it changes when the number of days changes (since when numOfDay = undefined we have a page for each day and when the number of days = 3 we have a page for every 3 days), when the scroll doesn't adjust to it we get an issue with the scrolling (it thinks that it scrolled 3 pages instead of 1 because of the pagesRef old value being used

Inbal-Tish: `Mutable values like 'something.current' aren't valid dependencies because mutating them doesn't re-render the component.`

src/dateutils.ts
```diff
@@ -20,7 +20,30 @@ export function sameDate(a?: XDate, b?: XDate) {
     return false;
   } else {
     return a?.getFullYear() === b?.getFullYear() && a?.getMonth() === b?.getMonth() && a?.getDate() === b?.getDate();
-  } 
+  }
+}
+
+export function sameDayRange({
```
Inbal-Tish, line 26: 'same**Date**Range', no?

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -50,9 +50,23 @@ const WeekCalendar = (props: WeekCalendarProps) => {
...
+    items.current = getDatesArray(date, firstDay, numberOfDays);
+    setListData(items.current);
+    visibleWeek.current = date;
+  }, [numberOfDays]);
+
   useDidUpdate(() => {
     if (updateSource !== UpdateSources.WEEK_SCROLL) {
-      const pageIndex = items.current.findIndex(item => sameWeek(item, date, firstDay));
+      const pageIndex = items.current.findIndex(
+        item => (numberOfDays && numberOfDays > 1) ?
```
Inbal-Tish, line 62: Better move this condition (numberOfDays && numberOfDays > 1) to a const since you're using the same one later in line 250

src/expandableCalendar/index.tsx
Inbal-Tish: Actualy I'm not sure memoizing this is a good idea as it effects other non-memoized consts

src/day-state-manager.ts
```diff
@@ -3,10 +3,10 @@ const {toMarkingFormat} = require('./interface');
 
 
 export function getState(day: XDate, current: XDate, props: any) {
-  const {minDate, maxDate, disabledByDefault, context} = props;
+  const {minDate, maxDate, disabledByDefault, context, disableDaySelection} = props;
```
Inbal-Tish, line 6: Why is `disableDaySelection` part of the 'props' if it's not a prop? It's confusing, why not add an option object or another pharam?

src/expandableCalendar/index.tsx
yuvalsho: added the dep

src/expandableCalendar/index.tsx
yuvalsho: it affects memoized consts or is used as an initializer for useRef references, also added the missing dep

src/dateutils.ts
yuvalsho: renamed

src/timeline-list/index.tsx
yuvalsho: removed unnecessary dependency

src/day-state-manager.ts
yuvalsho: i agree, moved to be a different parameter

src/dateutils.ts
```diff
@@ -20,7 +20,30 @@ export function sameDate(a?: XDate, b?: XDate) {
     return false;
   } else {
     return a?.getFullYear() === b?.getFullYear() && a?.getMonth() === b?.getMonth() && a?.getDate() === b?.getDate();
-  } 
+  }
+}
+
+export function onSameDAteRange({
```
Inbal-Tish, line 26: The "A" on Date is capitalized... typo I guess

yuvalsho: fixed

Inbal-Tish (approved):

PR #2081, opened 2022-11-07
add integration tests on expandable calendar

yuvalsho:


src/expandableCalendar/__test__/expandableCalendarTestKit.tsx
```diff
@@ -0,0 +1,41 @@
...
+  ExpandableCalendarProps
+} from 'react-native-calendars';
+import {toMarkingFormat} from '../../interface';
+
+const XDate = require('xdate');
+
+const today = new XDate();
+export const testIdExpandableCalendar = 'myExpandableCalendar';
+
+export const expandableCalendarTestIDs = (testId: string) => {
```
Inbal-Tish, line 15: Can't we move these to the driver? Why do we need another file?

src/expandableCalendar/__test__/index.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+import {CalendarContextProviderProps} from 'react-native-calendars';
+
+const XDate = require('xdate');
+
+enum Direction {
+  LEFT = 'left',
+  RIGHT = 'right',
+}
+
+const testIdExpandable = 'myExpandableCalendar';
```
Inbal-Tish, line 19: There's another one in the testKit file, it's confusing...

src/expandableCalendar/__test__/index.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+    });
+
+    it('should expand expandable header ', () => {
+      driver.toggleKnob();
+      jest.runAllTimers();
+
+      expect(driver.isCalendarExpanded()).toBe(true);
+    });
+
+    it('should day press close expandable header', () => {
```
Inbal-Tish, line 121: I would add another test to check if 'closeOnDayPress' disables this feature

src/expandableCalendar/__test__/index.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+        expect(onDateChanged).toHaveBeenNthCalledWith(2, toMarkingFormat(expectedDate), UpdateSources.PAGE_SCROLL);
+        expect(onMonthChange).toHaveBeenNthCalledWith(2, xdateToData(expectedDate), UpdateSources.PAGE_SCROLL);
+      });
+
+      it(`should navigate 6 months ahead and back successfully`, () => {
+        driver.toggleKnob();
+        jest.runAllTimers();
+        times(6, () => {
+          driver.pressOnHeaderArrow({left: false});
+          jest.runAllTimers();
```
Inbal-Tish, line 180: Do we need to run all times after every press? Can't we move it to the end of the loop?

src/expandableCalendar/__test__/index.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+        });
+        const expectedFutureDate = today.clone().addMonths(6).setDate(1);
+        expect(onDateChanged).toHaveBeenNthCalledWith(6, toMarkingFormat(expectedFutureDate), UpdateSources.PAGE_SCROLL);
+        expect(onMonthChange).toHaveBeenNthCalledWith(6, xdateToData(expectedFutureDate), UpdateSources.PAGE_SCROLL);
+        times(6, () => driver.pressOnHeaderArrow({left: true}));
```
Inbal-Tish, line 185: Here I see you didn't runAllTimes...

src/expandableCalendar/__test__/index.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+        expect(onDateChanged).toHaveBeenCalled();
+        // expect(onMonthChange).toHaveBeenCalled();
+      });
+    });
+
+    describe('arrow press', () => {
+      beforeEach(() => {
+        driver.render();
+      });
+      it.each([[Direction.LEFT],[Direction.RIGHT]])(`should call onDateChanged and onMonthChanged to next month first day when pressing the %s arrow`, (direction: Direction) => {
```
Inbal-Tish, line 154: You have a % sign in the string... 3 more in the file.

src/expandableCalendar/__test__/index.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+      });
+      driver = new ExpandableCalendarDriver(testIdExpandableCalendar, component);
+      driver.render();
+      const todayButton = driver.getTodayButton();
+      isToday ?
+        expect(todayButton).not.toBeDefined() :
+        expect(todayButton).toBeDefined();
+    });
+
+    it('should call onDateChanged with current date when today button pressed', () => {
```
Inbal-Tish, line 266:
```suggestion
    it('should call onDateChanged with today's date when today button pressed', () => {
```

src/expandableCalendar/__test__/index.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+          date: toMarkingFormat(today.clone().addDays(5)),
+        },
+      });
+      driver = new ExpandableCalendarDriver(testIdExpandableCalendar, component);
+      driver.render();
+      driver.pressOnTodayButton();
+      expect(onDateChanged).toHaveBeenCalledWith(toMarkingFormat(today), UpdateSources.TODAY_PRESS);
+    });
+
+    it('should call onMonthChanged with current month when today button pressed while current date is not in current month', () => {
```
Inbal-Tish, line 280:
```suggestion
    it('should call onMonthChanged with current month when today button pressed while initial date is not in current month', () => {
```

src/expandableCalendar/__test__/index.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+        const expectedDate = today.clone().setDate(today.getDate() + 7 * diff - today.getDay());
+        times(diff, () => driver.pressOnHeaderArrow({left: false}));
+        expect(onMonthChange).toHaveBeenCalledWith(xdateToData(expectedDate), UpdateSources.PAGE_SCROLL);
+      });
+    });
+  });
+
+  describe.skip('today button', () => {
+    //check if button appear even on today
+    it.each([['', `isn't`, false], ['not ', 'is', true]])('should %sappear when the current date %s today', (_message1, _message2, isToday) => {
```
Inbal-Tish, line 250:
```suggestion
    it.each([['', `isn't`, false], ['not ', 'is', true]])('should %sappear when the initial date %s today', (_message1, _message2, isToday) => {
```

src/expandableCalendar/__test__/index.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+        driver.render();
+      });
+      it.each([[Direction.LEFT], [Direction.RIGHT]])(`should call onDateChanged to next week first day when pressing %s arrow`, (direction) => {
+        const currentDay = today.getUTCDay();
+        const expectedDate = today.clone().addDays(direction === Direction.LEFT ? - (currentDay + 7) : (7 - currentDay));
+        driver.pressOnHeaderArrow({left: direction === Direction.LEFT});
+        expect(onDateChanged).toHaveBeenCalledWith(toMarkingFormat(expectedDate), UpdateSources.PAGE_SCROLL);
+      });
+
+      it(`should call onDateChanged for first day of current week when changing to current week`, () => {
```
Inbal-Tish, line 223: Not sure I understand this test... you pressing the right arrow, then the left arrow, and expect the date returned to be the first day of the current week, when the initial date was not the first day of the week. Right?

src/expandableCalendar/__test__/index.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+      it.each([[Direction.LEFT],[Direction.RIGHT]])(`should call onDateChanged and onMonthChanged to next month first day when pressing the %s arrow`, (direction: Direction) => {
+        driver.toggleKnob();
+        jest.runAllTimers();
+        const expectedDate = today.clone().addMonths(direction === Direction.RIGHT ? 1 : -1).setDate(1);
+        driver.pressOnHeaderArrow({left: direction === Direction.LEFT});
+        expect(onDateChanged).toHaveBeenCalledWith(toMarkingFormat(expectedDate), UpdateSources.PAGE_SCROLL);
+        expect(onMonthChange).toHaveBeenCalledWith(xdateToData(expectedDate), UpdateSources.PAGE_SCROLL);
+      });
+
+      it(`should call onDateChanged and onMonthChanged for first day in current month when changing to current month`, () => {
```
Inbal-Tish, line 163:
```suggestion
      it(`should call onDateChanged and onMonthChanged for first day in initial month when changing back to initial month`, () => {
```

src/expandableCalendar/__test__/expandableCalendarTestKit.tsx
yuvalsho: I believe it's easier to read when the driver itself (accessing components and performing actions on them) should be separated from the test kit (where we create all the mocks and the component we want to render in the driver)

src/expandableCalendar/__test__/index.spec.ts
yuvalsho: %s is used in it.each to read the string from the param in the test name (like in java string formatting)

src/expandableCalendar/__test__/index.spec.ts
yuvalsho: yes, when I initially wrote the test I remembered we returned to today when we scrolled to this week, but after we talked I understood this was not the expected behavior and changed to the correct behavior

src/expandableCalendar/__test__/index.spec.ts
Inbal-Tish: oh, I didn't know that! nice!

src/expandableCalendar/__test__/index.spec.ts
yuvalsho: removed the duplicate

src/expandableCalendar/__test__/index.spec.ts
yuvalsho: added

src/expandableCalendar/__test__/index.spec.ts
yuvalsho: fixed

Inbal-Tish (approved):

Issue #51, opened 2017-06-05
Disable days before today


lucasgdev:
There is a simple away to disable days before the current day?
Ex.: Disable all days before today

Or there is a milestone to implement that?  Thanks.

sibelius:
use maxDate https://github.com/wix/react-native-calendars/blob/master/src/calendar/index.js#L34

tautvilas:
I think minDate would be more appropriate ;)

lucasgdev:
Thanks guys. Makes totally sense. :)

PR #2119, opened 2022-12-16
Add marking.fillerColor prop

ice-orion:
Adds an ability to set a filler color for a specific marked point. Possible use case is when you need to set a different background color for day-start, day-end and filler points. Like so:

Here is an example of how it could be achieved:
```
marked[startDate] = {
      startingDay: true,
      color: COLORS.primaryLight,
      fillerColor: COLORS.secondaryFaint,
};
marked[fillerDate] = {
      color: COLORS.secondaryFaint,
};
marked[endDate] = {
      endingDay: true,
      color: COLORS.primaryLight,
      fillerColor: COLORS.secondaryFaint,
};
```

### DIFF
--- a/src/calendar/day/period/index.tsx
+++ b/src/calendar/day/period/index.tsx
@@ -128,9 +128,9 @@ const PeriodDay = (props: PeriodDayProps) => {
     const end = markingStyle.endingDay;
 
     if (start && !end) {
-      rightFillerStyle.backgroundColor = markingStyle.startingDay?.backgroundColor;
+      rightFillerStyle.backgroundColor = marking?.fillerColor ?? markingStyle.startingDay?.backgroundColor;
     } else if (end && !start) {
-      leftFillerStyle.backgroundColor = markingStyle.endingDay?.backgroundColor;
+      leftFillerStyle.backgroundColor = marking?.fillerColor ?? markingStyle.endingDay?.backgroundColor;
     } else if (markingStyle.day) {
       leftFillerStyle.backgroundColor = markingStyle.day?.backgroundColor;
       rightFillerStyle.backgroundColor = markingStyle.day?.backgroundColor;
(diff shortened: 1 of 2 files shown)


Inbal-Tish:
@ice-orion Hi. Thank you for the PR. Not sure I understand the need to pass 'fillerColor' when you can simply use 'color'. For example: 
`markedDates={{
  ['2023-07-06']: {startingDay: true, color: 'blue'},
  ['2023-07-07']: {color: 'lightblue'},
  ['2023-07-08']: {endingDay: true, color: 'blue'}
}}`

ice-orion:
@Inbal-Tish If we set just color, it would be like this:
image
So `color` controls the background color of a day (circle) + color of that day's fillers. So currently there is no way to set different colors for those.

Inbal-Tish:
> @Inbal-Tish If we set just color, it would be like this: image So `color` controls the background color of a day (circle) + color of that day's fillers. So currently there is no way to set different colors for those.

Not sure I understand what you can't set... The only difference I see is that in your implementation the start and end days get the whole circle () instead of the given design which is half a circle attached to the middle days |). Is that what you're trying to achieve?

ice-orion:
The goal is to be able to set different colors for start / end days + keep them as circles.

src/calendar/day/period/index.tsx
```diff
@@ -128,9 +128,9 @@ const PeriodDay = (props: PeriodDayProps) => {
     const end = markingStyle.endingDay;
 
     if (start && !end) {
-      rightFillerStyle.backgroundColor = markingStyle.startingDay?.backgroundColor;
+      rightFillerStyle.backgroundColor = marking?.fillerColor ?? markingStyle.startingDay?.backgroundColor;
```
Inbal-Tish, line 131: this is not the place I would check for the attribute. A better one is in the definition of `markingStyle` where we set the `.startingDay` and `.endingDay` (lines #51 and 54). For example:
```
 defaultStyle.startingDay = {backgroundColor: marking.fillerColor ?? marking.color};

PR #2126, opened 2022-12-20
Calendar - fix double selection

Inbal-Tish:
Calendar - fix double selection - generated from both 'markedDates' and 'state' props.

In controlled components, like Calendar and CalendarList, the selection is marked by the 'markedDates' passed by the user.
In an uncontrolled component, like ExpandableCalender, the selection is made by the 'state' passed by the component.
To fix the double selection I'm passing 'context' to the Calendar component from ExpandableCalendar, thus indicating that this is an uncontrolled component and the selection should be made from the 'state' prop. 
Otherwise, it will disable the selection through the 'state' prop passing **true** to getState() function's 'disableDaySelection' parameter.

### DIFF
--- a/src/expandableCalendar/index.tsx
+++ b/src/expandableCalendar/index.tsx
@@ -583,6 +583,7 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
         numberOfDays={numberOfDays}
         headerStyle={_headerStyle}
         timelineLeftInset={timelineLeftInset}
+        context={useContext(Context)}
       />
     );
   };
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,6 +1,9 @@
 import {ViewStyle, TextStyle} from 'react-native';
 import {MarkingProps} from './calendar/day/marking';
 
+export type ContextProp = {
+  context?: any;
+}
 export type MarkingTypes = 'dot' | 'multi-dot' | 'period' | 'multi-period' | 'custom';
 export type MarkedDates = {
   [key: string]: MarkingProps;
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,8 +1,9 @@
 import {ViewStyle, TextStyle} from 'react-native';
 import {MarkingProps} from './calendar/day/marking';
+import {CalendarContextProps} from './expandableCalendar/Context';
 
 export type ContextProp = {
-  context?: any;
+  context?: CalendarContextProps;
 }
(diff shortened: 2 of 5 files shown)


src/expandableCalendar/index.tsx
```diff
@@ -583,6 +583,7 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
         numberOfDays={numberOfDays}
         headerStyle={_headerStyle}
         timelineLeftInset={timelineLeftInset}
+        context={useContext(Context)}
```
ethanshar, line 586: Why are you passing context like this and not using Context provider? 
You can create a new Context just for this behavior

src/types.ts
```diff
@@ -1,6 +1,9 @@
 import {ViewStyle, TextStyle} from 'react-native';
 import {MarkingProps} from './calendar/day/marking';
 
+export type ContextProp = {
```
ethanshar, line 4: If you know the type, why are you using any here?

src/expandableCalendar/index.tsx
Inbal-Tish: Why create a new context for this?

src/types.ts
Inbal-Tish: Right. Fixed

src/expandableCalendar/index.tsx
ethanshar: You expect the user to tell if it's controlled/uncontrolled? 

Isn't it inherited by the components the user use? 
I mean if they use Expandable, you can tell it's uncontrolled and the other way around. 

It's a little weird to have the user decide on this, it's an internal implementation detail

Inbal-Tish: The user does NOT decide that at all. It is exactly how you described it - if the user uses ExpandableCalendar it is controlled since the Expandable is passing the 'context' prop. Other components do not thus they are uncontrolled. This is why the 'context' prop is private and not exposed to the user.

ethanshar: So am I missing something? Where in the code do you pass the `isControlled` in the context?

Inbal-Tish: `isControlled` is just a local const (I named it such to indicate that this is controlled/uncontrolled component. I could name it 'hasContext'...), if there is 'context' I pass 'disableDaySelection' to the 'getState()' function.

ethanshar (approved):

PR #2130, opened 2022-12-26
WeekCalendar - Fix day item not re-rendered when using `dayComponent`

jagamypriera:
There is problem with current `renderItem` on `WeekCalendar/new.tsx` and `WeekCalendar/index.tsx` where it doesn’t get updated when using custom `dayComponent`. This PR removes `useCallback` to fix it.


channeladam:
I think this might be a subset of #2140

Issue #2138, opened 2023-01-06
Performance issues with React 18 only on iOS


ashok3:
## Description

I'm trying to use the Agenda component in our app using React 18 and React Native 0.70, but I have noticed that the Calendar part is updating very slowly and even incorrectly when we scroll the Reservations part fast on iOS. 

To diagnose this, I tried the examples in the master branch and I noticed that it's working perfectly, but that you are using React 17 and React Native 0.68 for the examples.

Just to double check, I created a new React Native project using React 17 and React Native 0.68, and it worked like your example. However, I also created a new React Native project using React 18 and React Native 0.70 and the problem I describe above happens. 

All this happens only on iOS. It seems fine on Android. I've provided both projects files here for you to compare this and see the issue clearly.

## Expected Behavior

I expect it to work the same or even better in React 18 than React 17.

## Observed Behavior

I observed that the Calendar part of the Agenda component was updating the selected date very slowly when using React 18.

## Environment

"react": "18.1.0",
"react-native": "0.70.6",
"react-native-calendars": "^1.1293.0"

Testing on simulator: iPhone 14 Pro (iOS 16)

## Reproducible Demo

Demo using React 18:
[MyAppReact18.zip](https://github.com/wix/react-native-calendars/files/10363666/MyAppReact18.zip)

Demo using React 17:
[MyAppReact17.zip](https://github.com/wix/react-native-calendars/files/10363624/MyAppReact17.zip)

## Screenshots
React 17  ✅
https://user-images.githubusercontent.com/716309/211112127-78ac7d73-bed7-4b32-8758-ec1d0b0295f1.mp4

React 18 ❌
https://user-images.githubusercontent.com/716309/211112967-dc1c5a16-87da-4adf-a85c-b8d81bc7eba5.mp4

muhamdfikrii:
facing same issue

GideonAgboba:
facing same issue

devsales:
I'm having a performance issue with the Calendar & CalendarList components. It takes a long time to render and re-render (see video below). I haven't tested it on a real device though, only on the simulator.

"react": "18.1.0",
"react-native": "0.70.5",
"react-native-calendars": "^1.1293.0",

https://user-images.githubusercontent.com/47062834/212470336-21f7631d-2f8f-4ed5-9b06-931b0d7a2e47.mov

muhamdfikrii:
[...]

PR #2139, opened 2023-01-06
Agenda - Added ability to hide the calendar part of the Agenda component

ashok3:
This change adds the option to disable/hide the calendar on top of the Agenda component.

### DIFF
--- a/src/agenda/index.tsx
+++ b/src/agenda/index.tsx
@@ -27,11 +27,12 @@ import WeekDaysNames from '../commons/WeekDaysNames';
 import CalendarList, {CalendarListProps, CalendarListImperativeMethods} from '../calendar-list';
 import ReservationList, {ReservationListProps}  from './reservation-list';
 
+
 const HEADER_HEIGHT = 104;
 const KNOB_HEIGHT = 24;
 
 export type AgendaProps = CalendarListProps & ReservationListProps & {
-    /** the list of items that have to be displayed in agenda. If you want to render item as empty date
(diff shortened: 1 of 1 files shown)


Inbal-Tish:
@ashok3 Thank you for the PR. Can you please remove the formatting so I can review the relevant changes?

ashok3:
@Inbal-Tish - thanks for your reply. The formatting what are you referring to exactly? If you're talking about ESLint issues, they're already fixed in this commit on this PR: https://github.com/wix/react-native-calendars/pull/2139/commits/b6bc217b462c3a053498bd058440c77eed27267f and i think it passed all checks.

Let me know!

Also, we ended up using this and have a fully working version of this library in a soon to be production app now. Using the update here in this PR solves the issue we were seeing also here: https://github.com/wix/react-native-calendars/issues/2138 which seems to be a pretty critical issue. We almost didn't use the library. It seems like having this Calendar part on the Agenda causes the rendering issue which leads to that performance problem on latest versions of React Native. Just hiding it like we do in this PR seems to "fix" that or avoid the additional renders which works in our case. So it may be worth looking into why that's happening to solve that other issue.

Inbal-Tish:
@ashok3 Hi. By "formatting" I mean all the line deletions, indentations, commas, and line breaks - all the formatting changes make the review cumbersome. I would like to see only the relevant changes and avoid changing the formatting according to each dev's config

ashok3:
@Inbal-Tish - I've removed the formatting now, can you check? I think this causes the eslint checks to fail on GitHub though.

PR #2140, opened 2023-01-10
Fixed React Hook dependencies and TypeScript errors globally, and fixed ExpandableCalendar tests and a duplicate key error

channeladam:


src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -178,14 +183,14 @@ const WeekCalendar = (props: WeekCalendarProps) => {
         }
       }
     }
-  }, [onEndReached]);
+  }, [onEndReached, setDate]);
```
channeladam, line 186: I need to change this so there are no dependencies - otherwise there might be an invariant error from flatlist.

channeladam: Done

adrianurs (approved):

channeladam:
Is there any active maintainer of this repo that can review this?

Inbal-Tish:
@channeladam Hi. Thank you for the PR but I'm afraid this is impossible to review. 
First, no description explains what and why the changes are needed. Second, there are too many different issues that are more appropriate for separate PRs: lint, ts, dependencies array, etc. 
Also, I saw you're renaming methods to use PascalCase which is incorrect since you can't name a component with an action (renderXXX is a method name and not a component name). 
I'll be happy to review your PR but please separate it into a few PR, each responsible for one issue, otherwise there is no way for me to test it all together.
Thanks again!

PR #2155, opened 2023-02-06
Make calendar react to theme changes

jnowakow:
Calendar component doesn't react to changes of theme prop because it's using useRef for memonization. By replacing it with useMemo you achieve desired effect and allow component to react properly on changes. 

Before: 

https://user-images.githubusercontent.com/56261019/216993978-c80fd314-a143-4b46-89bb-4644791650e1.mov

After:

https://user-images.githubusercontent.com/56261019/216994019-fcf3aea2-9419-4686-906d-a7fdb06fe0b1.mov

### DIFF
--- a/src/calendar/day/basic/index.tsx
+++ b/src/calendar/day/basic/index.tsx
@@ -1,4 +1,4 @@
-import React, {Fragment, useCallback, useRef} from 'react';
+import React, {Fragment, useCallback, useMemo} from 'react';
 import {TouchableOpacity, Text, View, ViewProps} from 'react-native';
 
 import {xdateToData} from '../../../interface';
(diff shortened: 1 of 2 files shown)


Inbal-Tish:
@jnowakow Hi. Thank you for the PR. By design, the 'theme' prop was not meant to be a controlled prop but a configuration prop that you're setting once as changing it recreates stylesheets and rerenders the entire component.
Also, If we want to allow the theme to change we can't do it only for 2 components (Calendar and Day) it makes the code inconsistent and confusing. Consider adding this support to all components and be minded for performance.

pzatorski:
I don't think removing the `useRef` implementation is the best way of fixing this problem.

To achieve the same behaviour you could use the `key` props.

```
return (
	<Calendar key={darkTheme ? "dark" : "light"} />
)
```

leons1767:
> I don't think removing the `useRef` implementation is the best way of fixing this problem.
> 
> To achieve the same behaviour you could use the `key` props.
> 
> ```
> return (
> 	<Calendar key={darkTheme ? "dark" : "light"} />
> )
> ```

Nice workaround, cheers!

PR #2161, opened 2023-02-19
Week not updating when markings change

yuvalsho:


src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -92,16 +93,34 @@ const WeekCalendar = (props: WeekCalendarProps) => {
...
+    }, {})
+  }, [])
+
   const weekStyle = useMemo(() => {
     return [{width: containerWidth}, propsStyle];
   }, [containerWidth, propsStyle]);
 
   const renderItem = useCallback(({item}: {item: string}) => {
     const currentContext = sameWeek(date, item, firstDay) ? context : undefined;
+    const markings = getCurrentWeekMarkings(item, markedDates);
```
Inbal-Tish, line 118: Here you're getting the 'markedDates' for this week's item and passing it to the Week component. The Week component gets this object and extracts the specific date marking, but it does it anyways... so why is it important to pass the specific week's markings? Passing the marked date to the Week component and adding it to the dependencies array was not the fix?

yuvalsho: it was, but if I pass all the markings then all the weeks will rerender when a single marking was changed, this way only relevant weeks are re-rendered

Inbal-Tish (approved):

PR #2179, opened 2023-03-07
Added month change with ref

kanimetov:


simone989 (approved):

Inbal-Tish:
@kanimetov Hi. Thank you for the PR. Please add a description explaining why do you need to expose the addMonth method via ref. What is the use case?

simone989:
Hi @Inbal-Tish , I opened a similar Pull request a few months ago. Passing `initialDate` to the component works, but in my opinion, it's incorrect. The prop name is wrong.

> @kanimetov Hi. Thank you for the PR. Please add a description explaining why do you need to expose the addMonth method via ref. What is the use case? Also, if you need to change month why not pass the new month in the `initialDate` prop instead?

Inbal-Tish:
> Hi @Inbal-Tish , I opened a similar Pull request a few months ago. Passing `initialDate` to the component works, but in my opinion, it's incorrect. The prop name is wrong.
> 
> > @kanimetov Hi. Thank you for the PR. Please add a description explaining why do you need to expose the addMonth method via ref. What is the use case? Also, if you need to change month why not pass the new month in the `initialDate` prop instead?

Not sure I understand why. When you pass `initialDate` the Calendar is set (init) to this new date, which is the new "initial" date.
Other than the naming, is there an issue with this prop when trying to initiate the Calendar with a new date?

simone989:
> > Hi @Inbal-Tish , I opened a similar Pull request a few months ago. Passing `initialDate` to the component works, but in my opinion, it's incorrect. The prop name is wrong.
> > > @kanimetov Hi. Thank you for the PR. Please add a description explaining why do you need to expose the addMonth method via ref. What is the use case? Also, if you need to change month why not pass the new month in the `initialDate` prop instead?
> 
> Not sure I understand why. When you pass `initialDate` the Calendar is set (init) to this new date, which is the new "initial" date. Other than the naming, is there an issue with this prop when trying to initiate the Calendar with a new date?

The suffix "Initial" suggests that the variable initializes only the initial value, but in reality, it can also affect the displayed month of the calendar. 
It could be replaced with displayedMonth to make the function of the prop clearer.

PR #2201, opened 2023-03-29
Prevent date overflow on pressing arrows ( if currentDate > days in month)

jodrescher:
Prevent date overflow when the current date is greater than days in the next month (for example current 30 or 31 and switch from January to February, with only 28 or 29 days) (familiar to #2051)


Inbal-Tish:
@jodrescher Not sure I understand the issue. How is the current date greater than the last day of the month?

PR #2026, opened 2022-09-09
fix date.day text not vertically centered

skam22:
date.day (number) was not vertically centered on a marked period.

PR #2207, opened 2023-04-07
Fix date number not centered vertically

haeniya:
When using a custom font, the date number of the selected date in the standard calendar is not vertically centered.

calendar_before

calendar_after

Same thing was already done for the "period" styling in a different PR, see: https://github.com/wix/react-native-calendars/pull/2026

### DIFF
--- a/src/calendar/day/basic/style.ts
+++ b/src/calendar/day/basic/style.ts
@@ -13,7 +13,8 @@ export default function styleConstructor(theme: Theme = {}) {
     base: {
       width: 32,
       height: 32,
-      alignItems: 'center'
+      alignItems: 'center',
+      justifyContent: 'center'
     },
     text: {
       marginTop: constants.isAndroid ? 4 : 6,


Inbal-Tish (approved):

Inbal-Tish:
@haeniya Hi. Sorry, but I had to reverse this commit. Please see ExpandableCalendar screen with your code and without it: 
Simulator Screen Shot - iPhone 13 - 2023-05-21 at 13 08 30
Simulator Screen Shot - iPhone 13 - 2023-05-21 at 13 08 39

haeniya:
@Inbal-Tish thanks for checking, I wasn't aware of that, sorry. Any idea how we could fix the issue with the not centered date in the standard calendar? I see that there's a separate file for the expandable calendar styling: https://github.com/wix/react-native-calendars/blob/master/src/expandableCalendar/style.ts. Maybe we can tweak something there?

Inbal-Tish:
I can suggest that you override the theme `stylesheet.day.basic.base` styling through the 'theme' prop, for example:
```
{
'stylesheet.day.basic': {
    base: {
     width: 32,
      height: 32,
      alignItems: 'center',
      justifyContent: 'center'
    }
  }
}
```

haeniya:
Yes, that's what I already did for now. Just thought it might make sense to fix it in the source as others might experience the same issue. But it works for me currently so I'm fine with keeping it as-is. Thanks.

choijiho0021:
Thank you for the issue that helped me a lot.

Issue #2209, opened 2023-04-11
Doesn't exist a feature of dynamic height on CalendarListItem


freirepaulo:
## Description

I've been using the react-native-calendars in my project to show the list of one year in the calendar list component. Unfortully until the beginner I prepared if the first mount didn't start on the first day, for example, on April 15, the height of the calendar will be the same then the end of the mouth will be empty (for example on observed behavior)

## Observed Behavior

at the moment what happens is the month component always be the same height, independent of how many days the month has. So like the example above if the calendar starts on the 11th this empty space under the last line will be shown and if the calendar has fewer days then this behavior be worse, if thee month start in 29th then the month has only 3 days so the empty space will be much bigger.

<div>
<img src='https://user-images.githubusercontent.com/33518714/231136396-28d22d4b-e957-4775-90f3-e46f3752a18c.png' height='500' />
<img src='https://user-images.githubusercontent.com/33518714/231136407-671866b6-89d9-498d-8fbd-74b9e79030da.jpg' height='500' />
</div>

## Expected Behavior

for me, the calendar height needs to be dynamic based on the initial date of the month. So if the month is full like starting on the 1st the height will be full but if the month started on the 15th the height need to be half of the full month.

I took the liberty to change the library code to suggest a solution.
here we have these examples of the same screen, one with the current code and the other with my code.

<div>
<img src='https://user-images.githubusercontent.com/33518714/231153348-1fdc8724-e3d2-403a-b5c0-c00a60a72946.png' height='500' />
<img src='https://user-images.githubusercontent.com/33518714/231153358-950e3bef-3cfc-4821-9b0b-b3c3e4e0aac9.png' height='500' />
</div>

## Possible Solution

this is the diff between my changes and the current code. I will open a PR with these changes to see how we can improve it to be able to be merged.
 
<img width="1473" alt="Screenshot 2023-04-11 at 09 01 32" src="https://user-images.githubusercontent.com/33518714/231156669-997fc75d-5a68-470b-b5db-f9133cbdfe9d.png">

## Environment

Please run these commands in the project folder and fill in their results:

* react-native-calendars@1.1293.0
* react-native@0.70.0

Also, specify:

based on my test, this behavior happens in all devices and all versions.

## Reproducible Demo

to reproduce this issue you need to:
- create a React Native project
- add react-native-calendars
- use CalendarList with futureScrollRange of 12 and pastScrollRange of 0
- current of CalendarList needs to be on the 15th of the month

PR #2210, opened 2023-04-11
Dynamic height on CalendarListItem

freirepaulo:
### Issue related: Doesn't exist a feature of dynamic height on CalendarListItem #2209 

I've been using the react-native-calendars in my project to show the list of one year in the calendar list component. Unfortully until the beginner I prepared if the first mount didn't start on the first day, for example, on April 15, the height of the calendar will be the same then the end of the mouth will be empty (example above)

<div>

</div>

## Propose Solution
for me, the CalendarListItem height needs to be dynamic based on the initial date of the month. So if the month is full like starting on the 1st the height will be full but if the month started on the 15th the height need to be half of the full month.

I made a little change to the code and works for me, however, I would like to see with this PR if I need to consider more use cases of that and which improvement I need to do for this PR to be merged.

here we have these examples of the same screen, one with the current code and the other with my code.

<div>

</div>
<br>
To give a breath explanation of the code, I'm basically using the index of the renderItem to look for the first element of the calendar. after that, I take how many days this month has in total, and then after already knowing the initial date I can subtract the initial date of the total days to have the latest days of the month. With that value, I can divide it by 7 (a week) to get how many lines the calendar will have and then multiply by the CALENDAR_LINE_HEIGHT to get the height of this month.

### DIFF
--- a/src/calendar-list/index.tsx
+++ b/src/calendar-list/index.tsx
@@ -20,6 +20,7 @@ const CALENDAR_WIDTH = constants.screenWidth;
 const CALENDAR_HEIGHT = 360;
 const PAST_SCROLL_RANGE = 50;
 const FUTURE_SCROLL_RANGE = 50;
+const CALENDAR_LINE_HEIGHT = 70;
 
 export interface CalendarListProps extends CalendarProps, Omit<FlatListProps<any>, 'data' | 'renderItem'> {
   /** Max amount of months allowed to scroll to the past. Default = 50 */
@@ -223,10 +224,16 @@ const CalendarList = (props: CalendarListProps & ContextProp, ref: any) => {
     return false;
   }, [currentMonth]);
 
-  const renderItem = useCallback(({item}: {item: XDate}) => {
+  const renderItem = useCallback(({item, index}: {item: XDate, index: number}) => {
     const dateString = toMarkingFormat(item);
     const [year, month] = dateString.split('-');
     const testId = `${testID}.item_${year}-${month}`;
+    let dynamicHeight;
+    if(index === 0) {
+      const daysInCurrentMonth = new Date(Number(year), Number(month), 0).getDate();
+      const latestDays = daysInCurrentMonth - initialDate.current.getDate();
+      dynamicHeight = latestDays / 7 * CALENDAR_LINE_HEIGHT;
+    };
     return (
       <CalendarListItem
         {...calendarProps}
@@ -237,7 +244,7 @@ const CalendarList = (props: CalendarListProps & ContextProp, ref: any) => {
         // @ts-expect-error - type mismatch - ScrollView's 'horizontal' is nullable
         horizontal={horizontal}
         calendarWidth={calendarWidth}
-        calendarHeight={calendarHeight}
+        calendarHeight={dynamicHeight || calendarHeight}
         scrollToMonth={scrollToMonth}
         visible={isDateInRange(item)}
       />
--- a/src/calendar-list/index.tsx
+++ b/src/calendar-list/index.tsx
@@ -232,7 +232,7 @@ const CalendarList = (props: CalendarListProps & ContextProp, ref: any) => {
     if(index === 0) {
       const daysInCurrentMonth = new Date(Number(year), Number(month), 0).getDate();
       const latestDays = daysInCurrentMonth - initialDate.current.getDate();
-      dynamicHeight = latestDays / 7 * CALENDAR_LINE_HEIGHT;
+      dynamicHeight = Math.round(latestDays / 7) * CALENDAR_LINE_HEIGHT
     };
     return (
       <CalendarListItem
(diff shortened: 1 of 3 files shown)


src/calendar-list/index.tsx
```diff
@@ -223,10 +224,16 @@ const CalendarList = (props: CalendarListProps & ContextProp, ref: any) => {
     return false;
   }, [currentMonth]);
 
-  const renderItem = useCallback(({item}: {item: XDate}) => {
+  const renderItem = useCallback(({item, index}: {item: XDate, index: number}) => {
     const dateString = toMarkingFormat(item);
     const [year, month] = dateString.split('-');
     const testId = `${testID}.item_${year}-${month}`;
+    let dynamicHeight;
```
Inbal-Tish, line 231: This calculation looks like something that should be optional and passed to the component, otherwise, it will break current behavior - displaying an entire calendar month - that is what most users want.

freirepaulo: completely agree! So I created a new commit that isolates the functionality of getting the calendar height and made wrapped by a condition that took into consideration a component prop that by default is false but can be controlled by the user - Considering that the prop comes true, we have one more condition to check the index and know that it is the first item in the calendar and only then apply the dynamic height.

freirepaulo: @Inbal-Tish Any update on it?

stevengoldberg:
I need this feature so I tried this change out but it didn't work for my use case. I'm using a `CalendarList` that needs to scroll over the maximum range, plus a custom day component. I poked around in the code and haven't quite been able to get it working but I think an approach like this might work.

Define a `getCalendarHeight` function which determines the number of weeks in a month, then computes the calendar height based on a passed-in `dayHeight` prop + a `HEADER_HEIGHT` constant (although this would break if the header style is overridden, so `headerHeight` should probably be an optional prop as well). e.g.:

```
    const getCalendarHeight = useCallback((month, year) => {
        if (calendarHeightDynamic) {
            const numDays = new Date(year, month, 0).getDate()
            const firstDay = new Date(year, month - 1, 1).getDay()
            const lastDay = new Date(year, month - 1, numDays).getDay()

            let numWeeks = Math.floor((numDays + firstDay) / 7) // Number of full weeks
            if (lastDay < 6) { // If the last day is not a Sunday, count the partial week
                numWeeks += 1
            }
            return (dayHeight * numWeeks) + HEADER_HEIGHT
        }
        return calendarHeight
    }, [])
```

That's pretty straightforward — the other part will be calculating the correct offsets in `getItemLayout`, `scrollToDay`, and `scrollToMonth`. I think this could be accomplished by constructing a map with keys in the form of `${month}-${year}` and heights as values, then using that to compute the offset or scroll range for a given month.

I have to move on to something else for now but if no one else gets to it I'll try to implement later.

freirepaulo:
Those last commits were made to implement the changes suggested by @stevengoldberg . They involve considering the day and header heights to calculate the dynamic height of the calendar.

I need to add new props to the calendar list component: `dayHeight` and `headerHeight`. These props will provide information about the component because we are unable to retrieve the heights from the `dayComponent` prop or the `customHeader` prop. By using these new props and setting default values for them, we can ensure that nothing breaks if the developer uses the default day and header components. With this approach, the developer will need to create two constants: `BASIC_DAY_HEIGHT` and `HEADER_HEIGHT`, which will set the default height for these components (I have used the size that we are already using).

Note: I tried to obtain the heights dynamically from the custom components and even from the themes, but it was not possible. Therefore, we need to find a way to set a default value for the day height and header height because, if the developer uses the default components, these values will be undefined.

I believe this approach covers all use cases and is not a breaking change, so @Inbal-Tish can review it again to see if the merge is now possible.

Note: @stevengoldberg, I think the second part of your suggestion is more of a separate issue. We are waiting for @Inbal-Tish's response regarding this new code and will reevaluate if those additional changes make sense for this pull request.

stevengoldberg:
Those commits look good to me, but I think what's missing is handling [calendarSize](https://github.com/wix/react-native-calendars/pull/2210/files#diff-35a50640948fbbb26c395046b6f6b3b756d41cc3fc93b4c2c82e46c380f48c7cR118), which is used in `getItemLayout` and the `scrollTo` methods. Right now `calendarSize` is still static, so I don't see how this can work if the actual calendar size is dynamic.

freirepaulo:
@stevengoldberg, I know it's still missing, but I'm waiting for a response from someone at Wix to confirm that these changes are approved so that I can look into the scrolling. However, if I have some time this week, I'll work on it. And, of course, feel free to implement this feature as well.

kioltk:
I also experience inconvenience with current behavior. Static height can only benefit the horizontal calendars, despite static width that may benefit both, since the week length is not dynamic and can be calculated in advance.

This is how it looks for me:
Screenshot 2023-08-30 at 21 23 33


For those who wanders trying to resolve it, I'd recommend using `patch-package`, just grab the code from this pull request into the `node_modules` and create a patch

freirepaulo:
@kioltk Yes bro, I solved my problem using the `patch-package`, but the best solution will be to resolve this internally in the library, unfortunately, the Wix team doesn't give us attention to be able to merge those changes.

PR #2067, opened 2022-10-23
Added support for custom placeholder for CalendarList

leonardorib:
This just adds the possibility to render a custom placeholder on CalendarList while the Calendar item on the flatlist is not visible yet.

This can be useful when fast scrolling through months is a concern.

This problem is discussed here: https://github.com/wix/react-native-calendars/issues/325

Issue #325, opened 2018-01-31
CalendarList - blank space instead of month when scrolling fast


mharrisweb:
## Description

When scrolling from month to month, using the CalendarList, If you scroll quickly the month is displayed in large bold text, with an empty calendar. 

## Expected Behavior

If scrolling quickly through CalendarList. Month appears as blank space.

## Observed Behavior

No error, just blank space instead of month,.

## Environment

* `npm ls react-native-calendars`: react-native-calendars@1.16.1
* `npm ls react-native`: react-native@0.50.3

 Samsung s8 Android V7

## Reproducible Demo

![videotogif_2018 01 31_15 52 55](https://user-images.githubusercontent.com/8245572/35632881-cc336326-069f-11e8-89b3-d570f826e0e5.gif)

tautvilas:
Hi, this is the result of calendars optimization. Calendar days are not rendered until calendar scrolls into view. If you scroll really fast, rendering lags and for a short time you see placeholder that is visible until calendars are loaded (the other alternative would be to have blank calendars until they are rendered)

mharrisweb:
Hi @tautvilas 

So it cant be prevented? Could I use 'onVisibleMonthsChange' to just render the current/ next and previous months?

Also, would I be able to put a loading spinner in instead of the placeholder?

tautvilas:
Currently you can not customize placeholder. This is the place in code that you can change in your fork to display something else:

https://github.com/wix/react-native-calendars/blob/master/src/calendar-list/item.js#L45

mharrisweb:
Can I setup a pull request to add that functionaility?

tautvilas:
Yes

Mahdi-Farahani:
i have this issue how can i fix it? @tautvilas

leonardorib:
I would also like to be able to customize this placeholder.

I'm going to submit a pull request.

pfcodes:
Placeholder much needed

MursiDirect:
Any Solution yet?

PauloBarroso85:
Are you planning to merge the PR soon? 🤔

PR #2216, opened 2023-04-16
Placeholder for the CalenderList component

hainsdominic:
> All the credit goes to @leonardorib - I created the PR because https://github.com/wix/react-native-calendars/pull/2067 seems abandoned

Issue: https://github.com/wix/react-native-calendars/issues/325

Enables the user to define a renderPlaceholder prop in order to render something while the component is loading.

### DIFF
--- a/src/calendar-list/item.tsx
+++ b/src/calendar-list/item.tsx
@@ -1,6 +1,6 @@
 import XDate from 'xdate';
 import React, {useRef, useMemo, useCallback} from 'react';
-import {Text} from 'react-native';
+import {Text, View} from 'react-native';
 import {Theme} from '../types';
 import {toMarkingFormat} from '../interface';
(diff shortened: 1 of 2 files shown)


src/calendar-list/item.tsx
```diff
@@ -82,6 +84,11 @@ const CalendarListItem = React.memo((props: CalendarListItemProps) => {
   }, [onPressArrowRight, scrollToMonth]);
 
   if (!visible) {
+    if (renderPlaceholder) {
+      const year = item.getFullYear();
+      const month = item.getMonth();
+      return <View style={calendarStyle}>{renderPlaceholder(year, month)}</View>;
```
Inbal-Tish, line 90: Why not simply pass 'item' so the users will do the extraction if they need it

Issue #1228, opened 2020-07-13
React Native Calendar Agenda week view not rendering properly


aik117:
I started using react native calendars for my app. I used Agenda in one of my screens. There is a strange issue however when selecting dates. I have set pastScrollRange and futureScrollRange. The problem is whenever i select any date from the first month in calendar, it's showing me the last week of next month. For example if January 2017 is the first month in calendar, after selecting any date it shows me the last week of February 2017. Similarly when i select any date from the last month in calendar, it shows me the first week of that same month. I have been trying to figure out the solution but nothing seems to help. Any guidance will be highly appreciated.

Kindly look at the below screenshots. In the below screenshots December 2017 is the first month in calendar and December 2021 is the last month in calendar.

![image](https://user-images.githubusercontent.com/17512745/87301719-8261cd00-c529-11ea-9f6b-82ad9e85f6cd.png)
![image](https://user-images.githubusercontent.com/17512745/87301801-a02f3200-c529-11ea-9e79-3ac739d55ae6.png)

Here is my code.

import React, { useState, useEffect } from "react";
import { View, Text, Image, TouchableOpacity, SafeAreaView, StyleSheet, Platform } from "react-native";
import { Agenda } from "react-native-calendars";
import Toast from "react-native-simple-toast";
import moment from "moment";
import "../../../Utils/Globals";
import { SERVICES_FOR_DATE_URL, CLIENT_ID_KEY } from "../../../Config/Routes";
import { requestData } from "../../../Api/ApiRequests";
import ServiceEvent from "./ServiceEvent";
//import { checkLocationPermission, requestLocationPermission, getDistance } from "../../../Location/LocationManager";
import Geolocation from "react-native-geolocation-service";
import Icon from "react-native-vector-icons/FontAwesome";

let watchId = "";

const ServicesCalendarScreen = (props) => {

    //let date = moment();
    //date = moment().format("MMMM YYYY");
    //date = moment().month() + moment().year();
    //Toast.show(""+date, Toast.LONG);

    //events state hook for selected day events
    const [events, setEvents] = useState(null);

    //set selected month and year
    const [date, setDate] = useState(moment().format("MMMM YYYY"));

    //hook for setting selected date to pass to onRefresh prop of agenda to refresh the selected date contents
    const [dateToRefresh, setDateToRefresh] = useState(moment().format("YYYY-MM-DD"));

    const getSelectedDayEvents = (date) => {

        let serviceDate = moment(date);
[...]

Issue #519, opened 2018-06-07
Agenda's CalendarList Header is invisible on initial render, visible only after scrolling or opening CalendarList


lizalopez:
## Description
Agenda component's CalendarList header is not displaying dates at the top, on initial render. The header dates only become visible when opening the CalendarList and manually selecting a date, or scrolling down the Agenda components days until I scroll to a day on the next row of weeks (when the srcollAmount is adjusted in CalendarList's scrollToDay.

This occurs in our app on iOS and android devices, but does not occur in the Agenda example listed. 

## Expected Behavior

The Agenda component should render the CalendarList's header on initial load, without any user interaction.

## Observed Behavior

CalendarList is not rendering until specific user interaction- whether the current day has an event or is an empty array.

Likely similar to #405  or #485 , but while issue was marked as solved in v1.19.2,  it persists and the temporary solution does not work.

## Environment

* `npm ls react-native-calendars`:
└── react-native-calendars@1.19.3
* `npm ls react-native`:
└── react-native-calendars@0.47.2

1. Phone/emulator/simulator & version:
Simulator iOS 11.2 - iPhone 6s

### On inital load:
<img width="409" alt="screen shot 2018-06-07 at 3 27 38 pm" src="https://user-images.githubusercontent.com/10860957/41129547-736cf172-6a67-11e8-8aa9-cddd92e4f42e.png">

### After scrolling to new week:
<img width="411" alt="screen shot 2018-06-07 at 3 27 55 pm" src="https://user-images.githubusercontent.com/10860957/41129552-785bb8d0-6a67-11e8-9b05-a0bc49a2da6c.png">

## Reproducible Demo

Below is an example of the data we are feeding the `items` prop, for both empty and non-empty dates. Function loadScheduleItems makes a request to our database to fetch events for the current month, the previous month, and the next month-- the results are served in an object or arrays, like below:

`const schedules.events = { 
          2018-05-01: [],
          2018-05-02: [],
          2018-05-02: [ 
                  {company: 'string',
                  endDate : {Moment object},
                  id: 'string',
                  imageURL: 'string',
                  startDate: {Moment object},
          ]
   };`

    <Agenda
          items={schedules.events}
          loadItemsForMonth={this.loadScheduleItems}
          selected={startingDate.toDate()}
          renderItem={this.renderEvent}
          renderEmptyDate={this.renderEmptyDate}
          rowHasChanged={this.rowHasChanged}
      />

asaxena1237:
[...]

PR #2219, opened 2023-04-19
Fix the render issue of CalendarList in Agenda component caused by a race condition on layout change

Yupeng-li:
Sometimes the header with days in an agenda component is not displayed correctly after the initial render. For example, in the screenshot below, the selected day is 19th April, but the header shows the days from the previous week. 

The cause I found is that `onLayout` callback sometimes is called after [`onCalendarListLayout`](https://github.com/wix/react-native-calendars/blob/master/src/agenda/index.tsx#L265). The `calendarOffset` used in `onCalendarListLayout` depends on the `viewHeight` that is updated by `onLayout`. When this happens, the offset passed to `CalendarList.scrollToDay` will be incorrect, and therefore causes the rendering issue. 

The fix may help [issue 1228](https://github.com/wix/react-native-calendars/issues/1228) and [issue 519](https://github.com/wix/react-native-calendars/issues/519) too. 

Simulator Screenshot - iPhone 14 - 2023-04-19 at 17 18 25

### DIFF
--- a/src/agenda/index.tsx
+++ b/src/agenda/index.tsx
@@ -269,6 +269,7 @@ export default class Agenda extends Component<AgendaProps, State> {
   onLayout = (event: LayoutChangeEvent) => {
     this.viewHeight = event.nativeEvent.layout.height;
     this.viewWidth = event.nativeEvent.layout.width;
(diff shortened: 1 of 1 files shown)


liamjones:
Not stale

PR #2241, opened 2023-05-25
fix weekcalendar page change callback

YueLiXing:


src/expandableCalendar/WeekCalendar/new.tsx
```diff
@@ -60,11 +60,13 @@ const WeekCalendar = (props: WeekCalendarProps) => {
 
   const onPageChange = useCallback(
     (pageIndex: number, _prevPage, {scrolledByUser}) => {
+      const dateData = items[pageIndex];
       if (scrolledByUser) {
-        context?.setDate(items[pageIndex], UpdateSources.WEEK_SCROLL);
+        context?.setDate(dateData, UpdateSources.WEEK_SCROLL);
       }
+      props.onDayPress?.(xdateToData(dateData));
```
Inbal-Tish, line 67: onDayPress is called when a day is pressed while onPageChange is called when there is a scroll on the list. We can not mix them

PR #2242, opened 2023-05-28
CalendarList - fix staticHeader moves month out of range

Inbal-Tish:
The arrow press should stop passing months in the static header when getting to the end of the past/future scroll range.

To test: in the **horizontal CalendarList screen** change the `past/futureScrollRange` prop to 0 or a small number and press the arrows to move btw months - when you get to the edges you can see the months in the header changing but the calendar is off - the months should not change at all. This PR fix this bug.

### DIFF
--- a/example/src/screens/calendarListScreen.tsx
+++ b/example/src/screens/calendarListScreen.tsx
@@ -44,8 +44,8 @@ const CalendarListScreen = (props: Props) => {
     <CalendarList
       testID={testIDs.calendarList.CONTAINER}
       current={initialDate}
-      pastScrollRange={0}
-      futureScrollRange={12}
+      pastScrollRange={RANGE}
+      futureScrollRange={RANGE}
       onDayPress={onDayPress}
       markedDates={marked}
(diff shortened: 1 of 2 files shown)


example/src/screens/calendarListScreen.tsx
```diff
@@ -44,8 +44,8 @@ const CalendarListScreen = (props: Props) => {
     <CalendarList
       testID={testIDs.calendarList.CONTAINER}
       current={initialDate}
-      pastScrollRange={RANGE}
-      futureScrollRange={RANGE}
+      pastScrollRange={0}
+      futureScrollRange={12}
```
ethanshar, line 48: Did u leave this on purpose?

ethanshar (approved):
Approved. wrote a small comment. 
Btw, the description was really helpful 🙏

example/src/screens/calendarListScreen.tsx
Inbal-Tish: No... Fixed

Issue #2224, opened 2023-05-04
ExpandableCalendar not focusing on selected date Data


vinaybn-exxat:
## Description

ExpandableCalendar, AgendaList - Not scrolling to the data based on the date selection. It always scrolls back to the first data in the array.

Also the issue with the date marking when we scroll.

## Expected Behavior

On date selection, based on the date selection, data to be scrolled to the correct index

## Observed Behavior

Not working as expected

## Environment

I have used the same example as provided in the source code
https://github.com/wix/react-native-calendars/blob/master/example/src/screens/expandableCalendarScreen.tsx

Please run these commands in the project folder and fill in their results:

* `npm ls react-native-calendars`:
└── react-native-calendars@1.1295.0

* `npm ls react-native`:

├─┬ react-native-calendars@1.1295.0
│ └─┬ recyclerlistview@3.0.5
│   └── react-native@0.71.7 deduped
├─┬ react-native-date-picker@4.2.13
│
└── react-native@0.71.7

Also specify:

1. Device/emulator/simulator & OS version:
IOS Simulator - iPhone 14Pro

## ScreenRecording

https://user-images.githubusercontent.com/107985143/236255265-08dfd50f-ba0f-49d6-a584-8bd1956e65f8.mov

kovacsni76574:
In the code files of agendaList there is a function called scrollToSection and within this function the 'scrollToLocation' fuction gets called with the following parameters:
`            list?.current.scrollToLocation({
                animated: true,
                sectionIndex: sectionIndex,
                itemIndex: 0,
                viewPosition: 0,
                viewOffset: (constants.isAndroid ? sectionHeight.current : 0) + viewOffset
            });`

I then changed the itemIndex from 0 to 1 and all of a sudden the scrolling worked. I actually have no clue as to why that fixes the scrolling issues I had with the agendaList

vinaybn-exxat:
HI @kovacsni76574 , I did the same change and it's working.

Inbal-Tish:
@vinaybn-exxat We're checking to see if it's a RN71 issue and we'll update as soon as possible.

PR #2244, opened 2023-05-30
Updated the itemIndex

vinaybn-exxat:
The change is to fix scroll issue  on date selection, the fix enables the data to scroll to the selected date

https://github.com/wix/react-native-calendars/issues/2224#issuecomment-1568329186


vinaybn-exxat:
@Inbal-Tish , please review the fix. If the fix is appropriate, do merge the changes to main.

PR #2270, opened 2023-06-28
Add Optional InfiniteList Support to AgendaList

yakirza17:
Introduces an experimental prop `infiniteListProps` to `AgendaList`. When defined, `AgendaList` switches from `SectionList` to `InfiniteList`. Be aware, as an experimental feature, it might not support all existing APIs.

Additionally, enhances `InfiniteList` to accept a `layoutProvider` for detailed item size control.

### DIFF
--- a/src/expandableCalendar/agendaListNew.tsx
+++ b/src/expandableCalendar/agendaListNew.tsx
@@ -2,61 +2,32 @@ import PropTypes from 'prop-types';
 
 import isUndefined from 'lodash/isUndefined';
 import debounce from 'lodash/debounce';
-import isEqual from 'lodash/isEqual';
 import InfiniteList from '../infinite-list';
 
 import XDate from 'xdate';
 
 import React, {useCallback, useContext, useEffect, useMemo, useRef, useState} from 'react';
 import {
-  Text,
-  SectionListProps,
   DefaultSectionT,
   SectionListData,
-  ViewStyle,
   LayoutChangeEvent,
-  TextProps, View
 } from 'react-native';
 
 import {useDidUpdate} from '../hooks';
 import {getMoment} from '../momentResolver';
 import {isGTE, isToday} from '../dateutils';
 import {getDefaultLocale} from '../services';
-import {Theme} from '../types';
 import {UpdateSources, todayString} from './commons';
 import styleConstructor from './style';
 import Context from './Context';
 import constants from "../commons/constants";
 import {parseDate} from "../interface";
-
-// const viewabilityConfig = {
-//   itemVisiblePercentThreshold: 20 // 50 means if 50% of the item is visible
-// };
-
-export interface AgendaListProps extends SectionListProps<any, DefaultSectionT> {
-  /** Specify theme properties to override specific styles for calendar parts */
-  theme?: Theme;
-  /** day format in section title. Formatting values: http://arshaw.com/xdate/#Formatting */
-  dayFormat?: string;
-  /** a function to custom format the section header's title */
-  dayFormatter?: (arg0: string) => string;
-  /** whether to use moment.js for date string formatting
-   * (remember to pass 'dayFormat' with appropriate format, like 'dddd, MMM D') */
-  useMoment?: boolean;
-  /** whether to mark today's title with the "Today, ..." string. Default = true */
-  markToday?: boolean;
-  /** style passed to the section view */
-  sectionStyle?: ViewStyle;
-  /** whether to block the date change in calendar (and calendar context provider) when agenda scrolls */
-  avoidDateUpdates?: boolean;
-  /** offset scroll to section */
-  viewOffset?: number;
-  /** enable scrolling the agenda list to the next date with content when pressing a day without content */
-  scrollToNextEvent?: boolean;
-}
+import {LayoutProvider} from "recyclerlistview/dist/reactnative/core/dependencies/LayoutProvider";
+import {AgendaListProps, AgendaSectionHeader} from "./agendaList";
+import {ScrollEvent} from "recyclerlistview/src/core/scrollcomponent/BaseScrollView";
 
 /**
- * @description: AgendaList component
+ * @description: AgendaList component that use InfiniteList to improve performance
  * @note: Should be wrapped with 'CalendarProvider'
  * @extends: SectionList
  * @example: https://github.com/wix/react-native-calendars/blob/master/example/src/screens/expandableCalendar.js
@@ -65,29 +36,24 @@ const AgendaList = (props: AgendaListProps) => {
   const {
     theme,
     sections,
-    // scrollToNextEvent,
+    scrollToNextEvent,
     viewOffset = 0,
-    // avoidDateUpdates,
-    // onScroll,
-    // onMomentumScrollBegin,
-    // onMomentumScrollEnd,
-    // onScrollToIndexFailed,
+    avoidDateUpdates,
+    onScroll,
     renderSectionHeader,
     sectionStyle,
-    // keyExtractor,
     dayFormatter,
     dayFormat = 'dddd, MMM d',
     useMoment,
     markToday = true,
-    // onViewableItemsChanged,
   } = props;
 
-  const {date, updateSource} = useContext(Context);
+  const {date, updateSource, setDate} = useContext(Context);
 
   const style = useRef(styleConstructor(theme));
   const list = useRef<any>();
   const _topSection = useRef(sections[0]?.title);
-  // const didScroll = useRef(false);
+  const didScroll = useRef(false);
   const sectionScroll = useRef(false);
   const sectionHeight = useRef(0);
   const [data, setData] = useState([] as any[]);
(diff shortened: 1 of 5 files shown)


src/expandableCalendar/agendaListNew.tsx
```diff
@@ -0,0 +1,258 @@
...
+      list.current?.scrollToIndex(sectionIndex, true);
+    }
+  }, 1000, {leading: false, trailing: true}), [ sections]);
+
+  const layoutProvider = useMemo(
+    () => new LayoutProvider(
+      (index) => data[index]?.isTitle ? 'title': 'page',
+      (type, dim) => {
+        dim.width = constants.screenWidth;
+        dim.height = type === 'title' ? 60 : 120;
```
yakirza17, line 147: should see what the right size to pass

chenei: @ethanshar @Inbal-Tish please give your thoughts

src/infinite-list/index.tsx
```diff
@@ -157,13 +162,17 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
...
       dataProvider={dataProvider}
-      layoutProvider={layoutProvider.current}
+      layoutProvider={props.layoutProvider ?? layoutProvider.current}
       extendedState={extendedState}
       initialRenderIndex={initialPageIndex}
       renderAheadOffset={5 * pageWidth}
       onScroll={onScroll}
       style={style}
       scrollViewProps={scrollViewPropsMemo}
+      onMomentumScrollEnd={scrollViewProps?.onMomentumScrollEnd}
```
Inbal-Tish, line 172: scrollViewPropsMemo should include 'onMomentumScrollEnd', why separate them?

src/infinite-list/index.tsx
```diff
@@ -67,13 +68,17 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
   const reloadPagesDebounce = useCallback(debounce(reloadPages, 500, {leading: false, trailing: true}), [reloadPages]);
 
   useEffect(() => {
+    if (props.onEndReached) {
```
Inbal-Tish, line 71: Why this change? Even if you pass 'onEndReached' callback you need to scroll the data in case it changes... this is not clear

src/expandableCalendar/agendaListNew.tsx
```diff
@@ -0,0 +1,259 @@
...
+import Context from './Context';
+import constants from "../commons/constants";
+import {parseDate} from "../interface";
+import {LayoutProvider} from "recyclerlistview/dist/reactnative/core/dependencies/LayoutProvider";
+import {AgendaListProps, AgendaSectionHeader} from "./agendaList";
+
+/**
+ * @description: AgendaList component that use InfiniteList to improve performance
+ * @note: Should be wrapped with 'CalendarProvider'
+ * @extends: SectionList
```
Inbal-Tish, line 31: Doesn't extend SectionList, but Infinite list

src/expandableCalendar/agendaListNew.tsx
```diff
@@ -0,0 +1,259 @@
...
+ * @example: https://github.com/wix/react-native-calendars/blob/master/example/src/screens/expandableCalendar.js
+ */
+const AgendaList = (props: AgendaListProps) => {
```
Inbal-Tish, line 34: Rename the component AgendaListNew or InfiniteAgendaList

src/expandableCalendar/agendaListNew.tsx
```diff
@@ -0,0 +1,259 @@
...
+    if (props.renderItem) {
+      return props.renderItem({item} as any);
+    }
+
+    return <></>;
+  }, [props.renderItem]);
+
+  const _onEndReached = useCallback(() => {
+    if (props.onEndReached) {
+      props.onEndReached({distanceFromEnd: 0});
```
Inbal-Tish, line 228: Why distanceFromEnd is hard-coded to 0?

src/expandableCalendar/agendaListNew.tsx
```diff
@@ -0,0 +1,259 @@
...
+  }, [data]);
+
+  const _onMomentumScrollEnd = useCallback(() => {
+    sectionScroll.current = false;
+  }, []);
+
+  const headerTextStyle = useMemo(() => [style.current.sectionText, sectionStyle], [sectionStyle]);
+
+  const onHeaderLayout = useCallback((event: LayoutChangeEvent) => {
+    sectionHeight.current = event.nativeEvent.layout.height;
```
Inbal-Tish, line 200: Where are you using the sectionHeight.current?

src/expandableCalendar/agendaListNew.tsx
```diff
@@ -0,0 +1,259 @@
...
+  }, []);
+
+  const scrollToSection = useCallback(debounce((d) => {
+    const sectionIndex = scrollToNextEvent ? getNextSectionIndex(d) : getSectionIndex(d);
+    if (isUndefined(sectionIndex)) {
+      return;
+    }
+
+    if (list?.current && sectionIndex !== undefined) {
+      sectionScroll.current = true; // to avoid setDate() in onViewableItemsChanged
```
Inbal-Tish, line 135: You're not using onViewableItemsChanged here so why onViewableItemsChanged still set?

Inbal-Tish, line 135:
```suggestion
      sectionScroll.current = true; // to avoid setDate() in _onVisibleIndicesChanged
```

src/expandableCalendar/agendaListNew.tsx
Inbal-Tish: In the current example, it should be 80 instead of 120, but anyways you'll have to let the user pass the item's height to set it there.

Inbal-Tish:
@yakirza17 @chenei Overall looks good. Left a few comments

src/expandableCalendar/agendaListNew.tsx
yakirza17: removed

src/expandableCalendar/agendaListNew.tsx
yakirza17: The `onEndReached` of `RecyclerListView` is not pass any argument, so I don't have this information

src/expandableCalendar/agendaListNew.tsx
yakirza17: Renamed to `InfiniteAgendaList`

src/infinite-list/index.tsx
yakirza17: I don't want to scroll in the case of `onEndReached` I want to keep the scroll position in the same place (without this change, the view jump to the first item on any array update)
In case of new data changes, we have `scrollToIndex` on `InfiniteAgendaList`

src/infinite-list/index.tsx
yakirza17: removed

src/expandableCalendar/agendaListNew.tsx
yakirza17: Created `infiniteListProps` in the oldAgendaList, with an option to define these values (and load the new component in case it pass - even an empty object)

yakirza17: @Inbal-Tish It's ok to put values like that? will it work in any view? tablet view? landscape?

src/infinite-list/index.tsx
Inbal-Tish: You want to block the scroll when onEndReached is called or implemented?

yakirza17: If it exists

src/expandableCalendar/agendaListNew.tsx
yakirza17: It’s required on the onEndReach on the agenda list props, and I don’t want to change existing props

src/infinite-list/index.tsx
Inbal-Tish: But if the user wants to change the screen's background color every time the onEndReach is called, he needs to implement it, it doesn't mean he doesn't want the scroll to happen...

Inbal-Tish:
@yakirza17 Can you add a code snippet for the AgendaList so I can see on what example you're working on?

src/infinite-list/index.tsx
yakirza17: Ok, I will add a new props on infinite List that will disable this scroll

yakirza17: BTW, the infinite List does not supports onEndReached before this PR

yakirza17:
@Inbal-Tish, add `infiniteListProps={{}}` in the AgendaList on `example/src/screens/expandableCalendarScreen.tsx`
I also test it with our real-use repo (by changing the node modules code)

src/expandableCalendar/agendaListNew.tsx
Inbal-Tish: It will stay the same unless you change it when the orientation change. Regarding tablets, it depends on the design of course... Once you give the user the option to set these values it can be dynamic. 'pageWidth' and 'pageHeight' covers that

src/expandableCalendar/agendaListNew.tsx
Inbal-Tish: It a FlatList prop: onEndReached?: ((info: { distanceFromEnd: number }) => void) | null | undefined; So you can return undefined

yakirza17: I want to apply this CB for users that want to know when the end is reached, for applying this CB I must provide an info parameter (it's not optional), the null | undefined part just says that it's not required CB

image

Inbal-Tish: Ok... Add a comment about that please

src/infinite-list/index.tsx
Inbal-Tish: Yes, I noticed

src/infinite-list/index.tsx
```diff
@@ -67,13 +69,17 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
   const reloadPagesDebounce = useCallback(debounce(reloadPages, 500, {leading: false, trailing: true}), [reloadPages]);
 
   useEffect(() => {
+    if (props.disableScrollOnDataChange) {
```
Inbal-Tish, line 72: instead of extracting using props.xxx you can add it above where we extract all props. Also for the prop you're passing to the RecyclerListView

src/infinite-list/index.tsx
```diff
@@ -157,13 +164,16 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
       isHorizontal={isHorizontal}
       rowRenderer={renderItem}
       dataProvider={dataProvider}
-      layoutProvider={layoutProvider.current}
+      layoutProvider={props.layoutProvider ?? layoutProvider.current}
```
Inbal-Tish, line 167: here too

src/infinite-list/index.tsx
```diff
@@ -157,13 +164,16 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
...
       extendedState={extendedState}
       initialRenderIndex={initialPageIndex}
       renderAheadOffset={5 * pageWidth}
       onScroll={onScroll}
       style={style}
       scrollViewProps={scrollViewPropsMemo}
+      onEndReached={props?.onEndReached}
```
Inbal-Tish, line 174: the 3 here too

Inbal-Tish (approved):

PR #2272, opened 2023-07-04
Fix ReactImageView warning

Hiroenzo:
Remove WARN on console:  ReactImageView: Image source "null" doesn't exist


zkteco-home (commented):
good

geoffcfchen:
nice solution. I just create a patch to solve this.

PR #2283, opened 2023-07-16
Add support for refreshControl in the infiniteAgendaList.tsx

yakirza17:


src/expandableCalendar/infiniteAgendaList.tsx
```diff
@@ -46,7 +46,8 @@ const InfiniteAgendaList = (props: AgendaListProps) => {
     infiniteListProps,
     renderItem,
     onEndReached,
-    onEndReachedThreshold
+    onEndReachedThreshold,
+    refreshControl
```
Inbal-Tish, line 50: The 'refreshControl' prop is not declared anywhere. We can either add it to the component's props (extend AgendaListProps) or add the InfiniteListProps' scrollViewProps to the component's props and spread it before the overrides

yakirza17: Is declare here: `AgendaListProps extends SectionListProps` in the `SectionListProps` part

Inbal-Tish (approved):

PR #2286, opened 2023-07-18
Allow passing listStyle to the infiniteAgendaList to allow control the RecyclerListView style

yakirza17:


src/infinite-list/index.tsx
```diff
@@ -161,8 +163,8 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
   }, [onScrollBeginDrag, onMomentumScrollEnd, scrollViewProps, isHorizontal]);
 
   const style = useMemo(() => {
-    return {height: pageHeight};
-  }, [pageHeight]);
+    return {height: pageHeight, ...(listStyle ?? {})};
```
Inbal-Tish, line 166: Why no to return an array? 
`[{height: pageHeight}, listStyle]`

yakirza17: I want to allow an override of the height property if it is needed.
And the style is an object, not an array.

Inbal-Tish: if you have 'height' in listStyle it will override whether the style is an array or an object. 
Are you sure the style in RecyclerListView does not accept an array? ViewStyle does...

Inbal-Tish: What if the listStyle is a number (from a styleSheet)? You can't spread it...

src/infinite-list/index.tsx
```diff
@@ -28,6 +28,7 @@ export interface InfiniteListProps
   layoutProvider?: LayoutProvider;
   disableScrollOnDataChange?: boolean;
   renderFooter?: () => React.ReactElement | null;
+  listStyle?: object;
```
Inbal-Tish, line 31: This is not correct, for style prop the type should be StyleProp<ViewStyle>

src/infinite-list/index.tsx
```diff
@@ -28,6 +28,7 @@ export interface InfiniteListProps
   layoutProvider?: LayoutProvider;
   disableScrollOnDataChange?: boolean;
   renderFooter?: () => React.ReactElement | null;
+  listStyle?: StyleProp<ViewStyle>;
```
Inbal-Tish, line 31: RecycleListViewProp has a 'style' prop, why do we need another one here?

Inbal-Tish (approved):

PR #2310, opened 2023-08-10
fix: style prop in infinite-list

### LINKED ISSUE #2292: WeekCalendar style warning
## Description

When using a WeekCalendar i get a warning. The style gets applied correctly, but i would like the warning to go away.

## Expected Behavior

Style gets applied with no warning.

## Observed Behavior

What actually happened when you performed the above actions?

Seems like in  the infinite-list/index.js there is this code block:

    const _style = useMemo(() => {
        return [{ height: pageHeight }, style];
    }, [pageHeight, style]);

but the prop is expecting   style?: object | number;

 ERROR  Warning: Failed prop type: Invalid prop `style` supplied to `RecyclerListView`, expected one of type [object, number].
    at RecyclerListView (http://192.168.0.185:19000/node_modules/expo/AppEntry.bundle?platform=ios&dev=true&hot=false:570303:30)
    in InfiniteList (created by WeekCalendar)
    in RCTView (created by View)
    in View (created by WeekCalendar)
    in RCTView (created by View)
    in View (created by WeekCalendar)

fahyik:
Warning: Failed prop type: Invalid prop `style` supplied to `RecyclerListView`, expected one of type [object, number].

fixes #2292

### DIFF
--- a/src/infinite-list/index.tsx
+++ b/src/infinite-list/index.tsx
@@ -162,7 +162,7 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
   }, [onScrollBeginDrag, onMomentumScrollEnd, scrollViewProps, isHorizontal]);
 
   const _style = useMemo(() => {
-    return [{height: pageHeight}, style];
+    return {...style, height: pageHeight};
   }, [pageHeight, style]);
 
   return (


src/infinite-list/index.tsx
```diff
@@ -162,7 +162,7 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
   }, [onScrollBeginDrag, onMomentumScrollEnd, scrollViewProps, isHorizontal]);
 
   const _style = useMemo(() => {
-    return [{height: pageHeight}, style];
+    return {...style, height: pageHeight};
```
Inbal-Tish, line 165: In this case you're not letting the user override the height prop via style, also if the style is in array there's not need to spread it

Inbal-Tish:
@fahyik Thanks. This is a breaking change so it can't go in...

PR #2317, opened 2023-08-24
Fix android RTL issues

yuvalsho:


src/timeline-list/index.tsx
```diff
@@ -73,6 +75,12 @@ const TimelineList = (props: TimelineListProps) => {
     prevDate.current = date;
   }, [updateSource]);
 
+  const shouldApplyRTL = useMemo(() => constants.isRTL && constants.isAndroid, []);
```
Inbal-Tish, line 78:
```suggestion
  const shouldUseAndroidRTLFix = useMemo(() => constants.isRTL && constants.isAndroid, []);
```
Just to keep unified naming, or better let's move this to the 'commons/constants' file

src/expandableCalendar/WeekCalendar/index.tsx
```diff
@@ -70,14 +70,15 @@ const WeekCalendar = (props: WeekCalendarProps) => {
           }) :
           sameWeek(item, date, firstDay));
       if (pageIndex !== currentIndex.current) {
+        const adjustedIndexFrScroll = APPLY_ANDROID_FIX ? NUM_OF_ITEMS - 1 - pageIndex : pageIndex;
```
Inbal-Tish, line 73: 'shouldUseAndroidRTLFix' should be the name, and we need to take it from the constants

src/calendar-list/index.tsx
```diff
@@ -108,13 +109,16 @@ const CalendarList = (props: CalendarListProps & ContextProp, ref: any) => {
 
   const [currentMonth, setCurrentMonth] = useState(parseDate(current));
 
+  const isAndroidRTL = useMemo(() =>constants.isRTL && constants.isAndroid && horizontal, [horizontal]);
```
Inbal-Tish, line 112:
```suggestion
  const isAndroidRTL = useMemo(() => shouldUseAndroidRTLFix && horizontal, [horizontal]);
```
Need to take 'shouldUseAndroidRTLFix' from commons/constants

src/calendar-list/index.tsx
```diff
@@ -99,7 +100,7 @@ const CalendarList = (props: CalendarListProps & ContextProp, ref: any) => {
     onMomentumScrollEnd,
     /** FlatList props */
     onEndReachedThreshold,
-    onEndReached
+    onEndReached,
```
Inbal-Tish, line 103: no trailing commas please

src/infinite-list/index.tsx
```diff
@@ -68,6 +70,9 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
       }
     )
   );
+  const shouldUseAndroidRTLFix = useMemo(() => {
+    return constants.isAndroid && constants.isRTL && isHorizontal;
```
Inbal-Tish, line 74:
```suggestion
    return constants.isAndroidRTL && isHorizontal;
```

src/timeline-list/useTimelinePages.ts
```diff
@@ -26,6 +26,9 @@ const UseTimelinePages = ({date, listRef, numberOfDays}: UseTimelinePagesProps)
 
   const [pages, setPages] = useState<string[]>(pagesRef.current);
   const shouldResetPages = useRef(false);
+  const shouldUseAndroidRTLFix = useMemo(() => {
+    return constants.isAndroid && constants.isRTL;
```
Inbal-Tish, line 30:
```suggestion
    return constants.isAndroidRTL;
```

src/calendar-list/index.tsx
```diff
@@ -250,11 +254,21 @@ const CalendarList = (props: CalendarListProps & ContextProp, ref: any) => {
...
+  const onPressArrowRight = useCallback((method, month) => {
+    headerProps.onPressArrowRight?.(method, month);
+  }, [headerProps.onPressArrowRight]);
+
   const renderStaticHeader = () => {
     if (staticHeader && horizontal) {
       return (
         <CalendarHeader
           {...headerProps}
+          onPressArrowLeft={headerProps.onPressArrowLeft ? onPressArrowLeft : undefined}
```
Inbal-Tish, line 270: Why do we need to implement these callbacks? The CalendarHeader is invoking them by default..

src/infinite-list/index.tsx
```diff
@@ -82,7 +87,7 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
     }
 
     setTimeout(() => {
-      const x = isHorizontal ? Math.floor(data.length / 2) * pageWidth : 0;
+      const x = isHorizontal ? constants.isRTL && constants.isAndroid ? Math.floor(data.length / 2) + 1 : Math.floor(data.length / 2) * pageWidth : 0;
```
Inbal-Tish, line 90:
```suggestion
      const x = isHorizontal ? constants.isAndroidRTL ? Math.floor(data.length / 2) + 1 : Math.floor(data.length / 2) * pageWidth : 0;
```

src/calendar-list/index.tsx
yuvalsho: you are right, they were part of a previous attempt to fix and are obsolete now

src/calendar-list/index.tsx
```diff
@@ -304,6 +302,7 @@ const CalendarList = (props: CalendarListProps & ContextProp, ref: any) => {
       <FlatList
         // @ts-expect-error
         ref={list}
+        windowSize={constants.isAndroidRTL && horizontal ? pastScrollRange + futureScrollRange + 1 : undefined}
```
Inbal-Tish, line 305: We should use 'shouldUseAndroidRTLFix' here

Inbal-Tish (approved):

PR #2337, opened 2023-10-17
feat: hide days outside min/max date interval

FrSenpai:
You can now hide days that's outside the min/max date interval.
Usefull to reduce the height of the Calendar component and to display only required days.

Documentation & example added.


psegalen (commented):
Everything seems ok to me, I'm eager to use this new mode!

PR #2355, opened 2023-11-13
Today date state in extra day

adarshqb:


mikelgarciaurbina:
I need this fix. Any update on when it will be fixed?

Inbal-Tish:
@adarshqb Please add a description for what issue it this change trying to resolve

mikelgarciaurbina:
My problem is that when today is in another month. For example, today is August 1st, but I am viewing the month of July, August 1st appears as today and there is no way to remove the today styles

Inbal-Tish:
@mikelgarciaurbina I see what you mean but this is not a bug. The best way to handle your requirement is to open the option to customize the today marking (and in your case remove it when the visible month does not include the today), but we can't remove a feature as it a breaking change for those who don't share your specific requirement

PR #2367, opened 2023-12-08
fix(calendar-list): fix new calendar hide extra days prop

gijosso:
Fix the behavior causing `hideExtraDays`, from the `NewCalendarList`, to never be `false`.


gijosso:
As of today, the props is still unusable and it is a really simple fix.
I would love to use this feature, hopefully this fix can be integrated 🙏

PR #2368, opened 2023-12-11
feat(timeline): add testId to timeline

yakirza17:


Inbal-Tish:
@yakirza17 Why did you remove the change in TimelineList?

yakirza17:
@Inbal-Tish It will passed as part of _timelineProps

Inbal-Tish:
> @Inbal-Tish It will passed as part of _timelineProps

You only need it on the Timeline component? You don't need another testID on the TimelineList component?

yakirza17:
@Inbal-Tish no, the playable part is the scrollView inside the timeline, and the calendar component

Inbal-Tish (approved):

PR #2381, opened 2024-01-01
fix AgendaList component documentation

DominicData:
As per the code https://github.com/wix/react-native-calendars/blob/fbbdbbaa6eefd148b6df76ad5f1bf81d9bbf1fec/src/expandableCalendar/AgendaListsCommon.tsx#L6
This component is extending SectionList not FlatList and so has a different set of props


Inbal-Tish (approved):

Inbal-Tish:
@DominicData Sorry for the late review. Please resolve conflicts.

Inbal-Tish:
#rebuild

PR #2388, opened 2024-01-09
fix: fix custom style on Calendar type: period

liang54110:
Fix the bug that the backgroundColor defined in customContainerStyle will be overwritten
image
image


neboshall (approved):
I've tested this locally and it corrected the issue exactly as intended.

Inbal-Tish (approved):

PR #2391, opened 2024-01-11
fix timeline fixed width

jis93:
remove width style that prevented custom width to be applied


ammarNoon:
This is not the right way, i don't know if the devs are still active but if they are they can look at this

https://github.com/wix/react-native-calendars/pull/2490

PR #2392, opened 2024-01-11
fix: infinite-list reload data bug

XHFkindergarten:
in infinite-list/index.tsx, when onReachNearEdgeThreshold is 1, onMomentumScrollEnd will never be triggered

because case: pageIndex.current === 0 is never going to trigger
```
if (pageIndex.current) {
```

I know now `NUMBER_OF_PAGES` is 50, so we got enough space for `onNearEdge`, but actully it's a bug.

### DIFF
--- a/src/infinite-list/index.tsx
+++ b/src/infinite-list/index.tsx
@@ -137,17 +137,14 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
 
   const onMomentumScrollEnd = useCallback(
     event => {
-      if (pageIndex.current) {
-        if (isOnEdge.current) {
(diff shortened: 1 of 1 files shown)


src/infinite-list/index.tsx
```diff
@@ -137,17 +137,14 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
 
   const onMomentumScrollEnd = useCallback(
     event => {
-      if (pageIndex.current) {
```
Inbal-Tish, line 140: We can just check if isUndefined(pageIndex.current) instead of removing the condition

PR #2393, opened 2024-01-15
fix: calendar selected day text color(2386)

SanyaPillai:
(2386) - Calendar selectedDayTextColor not working as currentMonth state was not getting updated and the flag isControlled should have an not operator which determines the context is present or not.

Screenshot 2024-01-15 at 12 39 33 PM

Screenshot 2024-01-15 at 12 39 49 PM


zkteco-home (approved):

kishanbh (approved):

kishanbh (changes requested):
I think code set color for same date in all months with current month. please check again

PR #2397, opened 2024-01-19
fix horizontal loading spinner jump

mattg95:
The loading spinner causes the title to jump horizontally when it appears. This is due to the fact that an extra element is being rendered into the flex box. This PR adds a style to the loading spinner that does not cause the title to jump when the loading spinner is rendered

### Issue:
https://github.com/wix/react-native-calendars/assets/45606434/88793fa2-ce67-43a5-986c-33d2ab398fb2

### Fix:

https://github.com/wix/react-native-calendars/assets/45606434/9b8bca6d-4a99-445d-9b42-b3c4d7e6fd80

### DIFF
--- a/src/calendar/header/style.ts
+++ b/src/calendar/header/style.ts
@@ -68,6 +68,10 @@ export default function (theme: Theme = {}) {
     disabledDayHeader: {
       color: appStyle.textSectionTitleDisabledColor
     },
+    loadingIndicator: {
+      position: 'absolute',
+      right: -30,
+    },
     ...(theme['stylesheet.calendar.header'] || {})
   });
 }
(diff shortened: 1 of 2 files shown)


SamRotaCloud (approved):
Looks good to me 🚀

src/calendar/header/style.ts
```diff
@@ -68,6 +68,10 @@ export default function (theme: Theme = {}) {
     disabledDayHeader: {
       color: appStyle.textSectionTitleDisabledColor
     },
+    loadingIndicator: {
+      position: 'absolute',
+      right: -30,
```
Inbal-Tish, line 73: You need to center it vertically as well to the title

PR #2398, opened 2024-01-22
refactor: simplified isDateInRange func to improve large calendar lis…

f0wu5u:
When debugging to improve performance for a large calendar list ( eg. 15 months calendar list ), we found the `isDateInRange` not super optimize as we needed to complete some not-necessary loop action for each month each time we scroll through the calendar list.

|  Current `isDateInRange`  |  Next `isDateInRange`  |
|---|---|
|  Screenshot 2024-01-22 at 23 29 56  | Screenshot 2024-01-22 at 23 33 35 |

As you can see, we can get rid of the `not-necessary loop action`

### DIFF
--- a/src/calendar-list/index.tsx
+++ b/src/calendar-list/index.tsx
@@ -223,11 +223,10 @@ const CalendarList = (props: CalendarListProps & ContextProp, ref: any) => {
   }, []);
 
   const isDateInRange = useCallback((date) => {
-    for(let i = -range.current; i <= range.current; i++) {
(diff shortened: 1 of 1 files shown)


zkteco-home (approved):

stianjensen:
Great! Currently having issues rendering a calendar with more than 3 months of lookahead, without getting framedrops. So anything that makes re-renders faster would be really helpful!

jerone:
Any performance improvements are still relevant.

jerone:
Any performance improvements are still relevant.

PR #2410, opened 2024-02-13
Secure isLTE and isGTE functions

Inbal-Tish:
Secure isLTE and isGTE functions
Fix error: Cannot read property 'getTime' of undefined

### DIFF
--- a/src/dateutils.ts
+++ b/src/dateutils.ts
@@ -78,11 +78,15 @@ export function isToday(date?: XDate | string) {
 }
 
 export function isGTE(a: XDate, b: XDate) {
-  return b.diffDays(a) > -1;
+  if (a && b) {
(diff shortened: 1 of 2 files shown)


src/dateutils.ts
```diff
@@ -78,11 +78,15 @@ export function isToday(date?: XDate | string) {
 }
 
 export function isGTE(a: XDate, b: XDate) {
-  return b.diffDays(a) > -1;
```
ethanshar, line 81: What was returned before? 
Did it throw an error or return some value

ethanshar (approved):

src/dateutils.ts
Inbal-Tish: It throw en error: Cannot read property 'getTime' of undefined

PR #2416, opened 2024-02-22
typo

pvinis:


Inbal-Tish:
@pvinis Thanks. Just merged a fix for that

PR #2428, opened 2024-03-11
fix custom stylesheet for commonjs

suneox:
For common js can not get custom style with the current syntax

### DIFF
--- a/src/agenda/reservation-list/style.ts
+++ b/src/agenda/reservation-list/style.ts
@@ -1,5 +1,4 @@
 import {StyleSheet} from 'react-native';
-import get from 'lodash/get';
 import * as defaultStyle from '../../style';
 import {Theme} from '../../types';
 
@@ -38,6 +37,6 @@ export default function styleConstructor(theme: Theme = {}) {
     indicator: {
       marginTop: 80
     },
-    ...get(theme, 'stylesheet.agenda.list', {})
+    ...theme?.stylesheet?.agenda?.list
   });
 }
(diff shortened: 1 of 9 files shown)


src/agenda/reservation-list/style.ts
```diff
@@ -37,6 +38,6 @@ export default function styleConstructor(theme: Theme = {}) {
     indicator: {
       marginTop: 80
     },
-    ...(theme['stylesheet.agenda.list'] || {})
+    ...get(theme, 'stylesheet.agenda.list', {})
```
Inbal-Tish, line 41: Why 'get' and not optional chaining?

Inbal-Tish: The theme is a typed object, you can do `...theme?.stylesheet?.agenda?.main`

abraaoz:
Thank you @suneox. I am using your fix through patch-package, it works!
[react-native-calendars+1.1307.0.patch](https://github.com/user-attachments/files/17787729/react-native-calendars%2B1.1307.0.patch)

src/agenda/reservation-list/style.ts
```diff
@@ -37,6 +37,6 @@ export default function styleConstructor(theme: Theme = {}) {
     indicator: {
       marginTop: 80
     },
-    ...(theme['stylesheet.agenda.list'] || {})
```
Inbal-Tish, line 40: removing these is a breaking change. 
You can only add the optional chaining to allow passing nested object but we must keep the old one as it will stop working for users who still use this pattern.

PR #2435, opened 2024-03-19
fix calendar theme override

jrafflenbeul:
This pull request ensures uniform use of the theme overrides within the calendar component.


Inbal-Tish:
@jrafflenbeul Thanks. This is a very partial change as we use this pattern in almost every style file...

PR #2437, opened 2024-03-20
Add day component props for memo equality control

brianangulo:
Hi, @Inbal-Tish @ethanshar I would like to hear your thoughts on this. Thank you for all your work contributing to open source.

**The Problems:**

- The Day component will often cause performance issues coming from it not properly memoizing leading to unnecessary rerenders.
- Passing additional data to a custom day component would require declaring it as a function on the parent component and passing it to the Calendar. This is inconvenient and inefficient.

**Solutions:**

- Adding a prop `dayComponentMemoEqualityFn` that allows the user to override the default equality function thus giving the consumers more fine control over memoization of the component
- Adding a prop `extraData` inspired by react-native Flatlist's [extraData](https://reactnative.dev/docs/flatlist#extradata) to give the consumers a better and guided way of passing data

### DIFF
--- a/docsRNC/docs/Components/Calendar.md
+++ b/docsRNC/docs/Components/Calendar.md
@@ -180,6 +180,16 @@ Replace default title with custom element
 Replace default day with custom day rendering component  
 <span style={{color: 'grey'}}>JSX.Element</span>
 
+### dayComponentMemoEqualityFn
+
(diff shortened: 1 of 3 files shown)


brianangulo:
Could you take a peek please @M-i-k-e-l 🙏 ?

PR #2441, opened 2024-04-02
dateUtil - getMonthTitle fix

Inbal-Tish:
fix for wrong month at the beginning of the month


M-i-k-e-l:
@Inbal-Tish how can I reproduce this locally?

Inbal-Tish:
> @Inbal-Tish how can I reproduce this locally?

Pass instead of 'today' '2024-04-01'

Inbal-Tish:
> @Inbal-Tish I was unable to repro locally, were you able to repro locally?

Not any more... Let's leave it for next month :)

Issue #2350, opened 2023-11-04
"showSixWeeks" show  5 weeks or 7 weeks when firstDay != 0


ng-ha:
## Description

With option "showSixWeeks" calendar shows  5 weeks or 7 weeks  when firstDay != 0.

## Expected Behavior

"showSixWeeks" should always show 6 weeks so calendar's height doesn't change. 

## Environment

Please run these commands in the project folder and fill in their results:

* `npm ls react-native-calendars`: react-native-calendars@1.1302.0
* `npm ls react-native`: react-native@0.71.4

TobbeLundberg:
Came here to report the same issue
![image](https://github.com/wix/react-native-calendars/assets/87008627/0ecd359f-d7b1-4395-89e8-21ae87161bcf)

ThomasGuenard:
I was struggling with this issue too.

I created a pull request that solve this behavior

The first week calculation wasn't taking into account the first day of the week

Here is my patch:

`diff --git a/node_modules/react-native-calendars/src/dateutils.js b/node_modules/react-native-calendars/src/dateutils.js
index 34ac965..6918af8 100644
--- a/node_modules/react-native-calendars/src/dateutils.js
+++ b/node_modules/react-native-calendars/src/dateutils.js
@@ -101,7 +101,7 @@ export function page(date, firstDayOfWeek = 0, showSixWeeks = false) {
     const ldow = (fdow + 6) % 7;
     firstDayOfWeek = firstDayOfWeek || 0;
     const from = days[0].clone();
-    const daysBefore = from.getDay();
+    const daysBefore = (from.getDay() + 7 - fdow) % 7;
     if (from.getDay() !== fdow) {
         from.addDays(-(from.getDay() + 7 - fdow) % 7);
     }
`

matyasproch:
This is still a problem

PR #2444, opened 2024-04-05
fix: fdow taken into account when calculating daysBefore

ThomasGuenard:
fix for [this issue](https://github.com/wix/react-native-calendars/issues/2350)

The first week wasn't well calculated when the first day of week isn't sunday

### DIFF
--- a/src/dateutils.ts
+++ b/src/dateutils.ts
@@ -131,25 +131,25 @@ export function page(date: XDate, firstDayOfWeek = 0, showSixWeeks = false) {
   let after: XDate[] = [];
 
   const fdow = (7 + firstDayOfWeek) % 7 || 7;
-  const ldow = (fdow + 6) % 7;
-
+    const ldow = (fdow + 6) % 7;
+  
   firstDayOfWeek = firstDayOfWeek || 0;
-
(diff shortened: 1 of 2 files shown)


src/dateutils.ts
```diff
@@ -131,25 +131,25 @@ export function page(date: XDate, firstDayOfWeek = 0, showSixWeeks = false) {
...
   firstDayOfWeek = firstDayOfWeek || 0;
-
+  
   const from = days[0].clone();
-  const daysBefore = from.getDay();
+  const daysBefore = (from.getDay() + 7 - fdow) % 7;
 
-  if (from.getDay() !== fdow) {
+    if (from.getDay() !== fdow) {
     from.addDays(-(from.getDay() + 7 - fdow) % 7);
```
Inbal-Tish, line 142:
```suggestion
    from.addDays(-daysBefore);
```

ThomasGuenard (commented):
Well spotted!

ThomasGuenard:
@Inbal-Tish  maybe you know what was missing?

Why this PR that resolves an actual bug was set to stale?

PR #2470, opened 2024-05-17
Implement hook for change week on scroll (or tab button)

Dietermartens:
We were missing a hook for the scroll action in the WeekCalendar. This is very handy to handle because when the user scrolls, the parent component does not know the selected date of the week view.


Inbal-Tish:
@Dietermartens Hi. Thank you for the PR. I wonder why the 'onDateChanged' on the ContextProvider not providing you with the info you need. It is called when the date change and returns the new date and the updateSource so you can tell what action caused this update, for example 'weekScroll'.

PR #2484, opened 2024-06-21
fix: XDate uses wrong format for formatting date in 12 h format in timeline EventBlock.js

luckykenlin:
This PR addresses an issue with the 12-hour format display on the timeline event component. Previously, the component did not correctly format the time when using a 12-hour clock, which caused confusion for users relying on AM/PM distinctions.

Screenshots:
image

[XDate docs](https://arshaw.com/xdate/#Formatting)


luckykenlin:
any one review this?

PR #2488, opened 2024-06-27
feat: 

hanseulhee:
hello! I really like this library and am using it well.

If you find a part during use that you would like to see modified or added, send a PR.

**correction part**
This was corrected because monthNames was duplicated twice in README.

**additional part**
We also added Korea to LocaleConfig.locales to help provide convenience to Korean users. Since it is used by people all over the world, we cautiously suggest that it would be a good idea to provide multiple languages ​​as constants in the library.

thank you! 😀

### DIFF
--- a/example/src/app.tsx
+++ b/example/src/app.tsx
@@ -1,9 +1,9 @@
+import { Navigation } from 'react-native-navigation';
 import './wdyr'; // <--- must be first import
-import {Navigation} from 'react-native-navigation';
 // import {I18nManager} from 'react-native'; // <--- In order to test RTL
 //@ts-expect-error
-import {LocaleConfig} from 'react-native-calendars';
-import {registerScreens} from './screens';
+import { LocaleConfig } from 'react-native-calendars';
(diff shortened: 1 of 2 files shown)


example/src/app.tsx
```diff
@@ -33,6 +33,27 @@ LocaleConfig.locales['en'] = {
 };
 LocaleConfig.defaultLocale = 'en';
 
+LocaleConfig.locales['ko'] = {
```
Inbal-Tish, line 36: This is what you need to do in your app, not here

hanseulhee: It's a proposed solution, and Read Me seems in need of revision! Thank you for answer.

hanseulhee: It's a proposed solution, and README seems in need of revision! Thank you for answer. @Inbal-Tish

Inbal-Tish: Thanks. I just merged a PR for the readme change. Changing the app.tsx is affecting the demo so it can't get in

PR #2490, opened 2024-07-02
created new props to fix timeline width

ammarNoon:
Currently, the timeline events and hour rows are taking the width of a screen, which makes the event box overflow
added a prop where you can pass your container width to contain everything within

Before:
Screenshot 2024-07-02 at 12 51 48 PM

After:
Screenshot 2024-07-02 at 12 56 15 PM

### DIFF
--- a/src/timeline/Timeline.tsx
+++ b/src/timeline/Timeline.tsx
@@ -115,6 +115,10 @@ export interface TimelineProps {
   timelineLeftInset?: number;
   /** Identifier for testing */
   testID?: string;
+  /**
+   * Width for the timeline component
+   */
+  availableWidth?: number;
 }
 
 const Timeline = (props: TimelineProps) => {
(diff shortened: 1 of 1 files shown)


src/timeline/Timeline.tsx
```diff
@@ -247,7 +252,7 @@ const Timeline = (props: TimelineProps) => {
       // @ts-expect-error
       ref={scrollView}
       style={styles.current.container}
-      contentContainerStyle={[styles.current.contentStyle, {width: constants.screenWidth}]}
+      contentContainerStyle={[styles.current.contentStyle, {width: (availableWidth ?? constants.screenWidth)}]}
```
Inbal-Tish, line 255: I would prefer not to have a prop just for the content container width but instead, allow overriding/merge to the 'contentContainerStyle'. Maybe by exposing 'scrollViewProps' prop where you can pass any ScrollView props and override some (!) of its props.

PR #2497, opened 2024-07-10
feat(infiniteAgendaList): support passing scrollViewProps

yakirza17:
Since the `AgendaListProps` extends `SectionListProps` we can receive any section list props.
This PR just moves them to the right place inside `InfiniteAgendaList`

### DIFF
--- a/src/expandableCalendar/infiniteAgendaList.tsx
+++ b/src/expandableCalendar/infiniteAgendaList.tsx
@@ -47,7 +47,7 @@ const InfiniteAgendaList = ({
   onEndReached,
   onEndReachedThreshold,
   ...others
-}: AgendaListProps) => {
+}: Omit<AgendaListProps, 'viewOffset'>) => {
   const {date, updateSource, setDate} = useContext(Context);
 
   const style = useRef(styleConstructor(theme));
--- a/src/expandableCalendar/infiniteAgendaList.tsx
+++ b/src/expandableCalendar/infiniteAgendaList.tsx
@@ -47,7 +47,6 @@ const InfiniteAgendaList = (props: AgendaListProps) => {
     renderItem,
     onEndReached,
     onEndReachedThreshold,
-    refreshControl
   } = props;
 
   const {date, updateSource, setDate} = useContext(Context);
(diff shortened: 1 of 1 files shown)


src/expandableCalendar/infiniteAgendaList.tsx
```diff
@@ -244,7 +243,7 @@ const InfiniteAgendaList = (props: AgendaListProps) => {
       layoutProvider={layoutProvider}
       onScroll={_onScroll}
       onVisibleIndicesChanged={_onVisibleIndicesChanged}
-      scrollViewProps={{onMomentumScrollEnd: _onMomentumScrollEnd, nestedScrollEnabled: true, refreshControl}}
+      scrollViewProps={{nestedScrollEnabled: true, ...props, onMomentumScrollEnd: _onMomentumScrollEnd}}
```
Inbal-Tish, line 246: You're passing all the props... this can cause a problem for example to the onScroll we pass the _onScroll and now scrollViewProps gets the props.onScroll. I don't like it. IMO is is better to create a proper const _scrollViewProps that passes only the relevant props with regard to what we want the user to override and what not

src/expandableCalendar/infiniteAgendaList.tsx
```diff
@@ -244,7 +242,7 @@ const InfiniteAgendaList = (props: AgendaListProps) => {
       layoutProvider={layoutProvider}
       onScroll={_onScroll}
       onVisibleIndicesChanged={_onVisibleIndicesChanged}
-      scrollViewProps={{onMomentumScrollEnd: _onMomentumScrollEnd, nestedScrollEnabled: true, refreshControl}}
+      scrollViewProps={{nestedScrollEnabled: true, ...others, onMomentumScrollEnd: _onMomentumScrollEnd}}
```
Inbal-Tish, line 245: We want to allow users to change 'nestedScrollEnabled'?

yakirza17: It's a good question, but it looks to me depending on the usage, so yes.

Inbal-Tish: `viewOffset` prop is used in AgendaList component. 
It shouldn't affect anything since it's not a ScrollView prop, so I guess it's ok..

Inbal-Tish: Or we can omit is from the type here

Inbal-Tish: `const InfiniteAgendaList = (props: Omit<AgendaListProps, 'viewOffset'>) => {
`

yakirza17: Yes, do you want that I will add it?

Inbal-Tish (approved):

KrisLau:
Is there documentation anywhere on how to use InfiniteAgendaList?

PR #2502, opened 2024-08-03
fix(AgendaList):  renderSectionHeader do not match the expected signature

bravesoul349:
The renderSectionHeader prop is not a custom property of AgendaList; it is expected to match the signature of the SectionList, which aligns with the current typing.

It is also useful in cases when users do not want to render headers when data is empty.

### DIFF
--- a/src/expandableCalendar/agendaList.tsx
+++ b/src/expandableCalendar/agendaList.tsx
@@ -213,7 +213,7 @@ const AgendaList = (props: AgendaListProps) => {
     const title = info?.section?.title;
 
     if (renderSectionHeader) {
-      return renderSectionHeader(title);
+      return renderSectionHeader(info);
     }
 
     const headerTitle = getSectionTitle(title);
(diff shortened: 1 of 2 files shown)


NguyenVuAnh1996:
Hi, I am the person who you replied yesterday / 2 days ago.

First of all, thank you for your reply.

Regarding your commit, I see that if you simply change the prop of renderSectionHeader, it may break other people's codes, who had already put a title string into the parameter of renderSectionHeader. Instead, I suggest we create another prop that does this.

Btw, I am not well-knowledged with how github works, but if I click add my own review then this commit will be automatically approved to the main branch. Is that correct?
If so then does it only take 2 random people to break the whole project?

bravesoul349:
@NguyenVuAnh1996, you're right; this change will disrupt users' existing code. I'm not sure why it's currently implemented this way, but technically, the prop should have been like this from the start. Let's get input from others first. We can either introduce a new prop and update the existing typing to avoid confusion for users, or we can categorize this under "breaking changes."

NguyenVuAnh1996:
tbh I am trying to find another library to implement a task-calendar.
This library seems so chaotic, and if I am right, 2 random people can just make any changes to this, and also, since the original coders are clearly pro-ukraine just by looking at the homepage, I can see that it attracts some people who want to sabotage.

Issue #2194, opened 2023-03-21
Delay when using WeekCalendar and AgendaList together.


aleberguer:
## Description

Hello! I’m using the CalendarProvider to show a page with a `WeekCalendar` and an `AgendaList` working together. Everything looks good except for a delay when tapping on a day in the WeekCalendar. I’m attaching a video with this issue.

I was digging into the code, and I found that there is a debounce of 1 second in the [`scrollToSection` function](https://github.com/wix/react-native-calendars/blob/00c287776e21eb21cb82b2b863e15b82dc711c66/src/expandableCalendar/agendaList.tsx#L167) that is causing this delay. I tried to remove the debounce, and the delay is gone.

I'm not sure why this debounce is there, but maybe we can add a prop to decide whether to use the debounce or not?

## Expected Behavior

When tapping a day in the WeekCalendar, the AgendaList scrolls immediately, and there is no noticeable delay between the tap in WeekCalendar and the scrolling.

## Observed Behavior

When tapping a day in the WeekCalendar, there is a delay in scrolling the AgendaList to the day selected.

## Environment

`react-native-calendars@1.1289.0`
`react-native@0.65.2`

This is happening on both iOS/Android emulators and physical devices.

## Reproducible Demo

```
 <CalendarProvider date="2022-01-07">
        <WeekCalendar hideDayNames firstDay={1} />
        <AgendaList
          sections={ITEMS}
          renderItem={CardItem}
          horizontal
          pagingEnabled
        />
</CalendarProvider>
```
## Video

https://user-images.githubusercontent.com/5271876/226691487-1925e483-cc17-4dc2-9684-d8b2be645c48.mov

brianjuhl:
Also experiencing this but when I remove the debounce the delay seems to persist for some reason. Have used patch package, removed node_modules, cleared expo cache, etc etc.

aleberguer:
@Inbal-Tish, what are your thoughts on this issue? 
I can work on making a PR, but I would like to know the best approach here. I don't know what the debounce is expected to solve, but it is making the calendar very laggy and not the best user experience.

Can we add a prop to the AgendaList to decide whether to use the debounce? 
Or maybe a prop to give the debounce time, 1000 seems pretty arbitrary here.

zkteco-home:
same issue

juanmigdr:
Hey, I have had the same issue and created this patch which works as a temporary fix until the team creates a PR:

[...]

PR #2503, opened 2024-08-04
fix(AgendaList): Allow user to control delay in scrollToSection

bravesoul349:
Solves https://github.com/wix/react-native-calendars/issues/2194

### DIFF
--- a/docsRNC/docs/Components/AgendaList.md
+++ b/docsRNC/docs/Components/AgendaList.md
@@ -46,6 +46,11 @@ Whether to block the date change in calendar (and calendar context provider) whe
 Whether to enable scrolling the agenda list to the next date with content when pressing a day without content  
 <span style={{color: 'grey'}}>boolean</span>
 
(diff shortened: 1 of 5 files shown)


bravesoul349:
I have set the default value to 0 because this is the expected default behaviour.

Inbal-Tish:
> I have set the default value to 0 because this is the expected default behaviour by most of the users to have no delay.

The delay is intentional, without which the debounce is meaningless... I understand the need to control this delay but the default should be kept to the current behavior otherwise this might cause issues to current implementations that relay on this delay.

bravesoul349:
> > I have set the default value to 0 because this is the expected default behaviour by most of the users to have no delay.
> 
> The delay is intentional, without which the debounce is meaningless... I understand the need to control this delay but the default should be kept to the current behavior otherwise this might cause issues to current implementations that relay on this delay.

I have updated the code as you suggested.

Issue #2280, opened 2023-07-13
`ExpandableCalendar` expand/collapse animation is so laggy


jagamypriera:
## Description

`ExpandableCalendar` expand/collapse animation is so laggy. 
See video below

## Reproducible Demo

https://github.com/wix/react-native-calendars/assets/4131801/4cff02e1-aab5-46f6-a334-da098e71b7ca

minh-dai:
+1

teodorkdev:
+1

asevims:
+1

yuezonglun:
+1

netojose:
+1

yuezonglun:
这是来自QQ邮箱的假期自动回复邮件。您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

Dundyne:
Have you found a fix for this?

SoohyungJung:
+1

guilhermeferreira7:
+1

yuezonglun:
这是来自QQ邮箱的假期自动回复邮件。您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

jurij:
no fix?

ShanTheShan:
Facing this exact same issue right. Dug around and found there are other open and closed issues regarding this that havent been fixed after years. I suggest looking elsewhere.

yuezonglun:
这是来自QQ邮箱的假期自动回复邮件。您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

yuezonglun:
这是来自QQ邮箱的假期自动回复邮件。您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

yuezonglun:
这是来自QQ邮箱的假期自动回复邮件。您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

PR #2504, opened 2024-08-04
fix(ExpandableCalender): laggy animation

bravesoul349:
solves https://github.com/wix/react-native-calendars/issues/2280

### DIFF
--- a/src/expandableCalendar/index.tsx
+++ b/src/expandableCalendar/index.tsx
@@ -384,22 +384,21 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
       _height.current = toValue || newValue;
       _isOpen = _height.current >= threshold; // re-check after _height.current was set
 
+      resetWeekCalendarOpacity(_isOpen);
       Animated.spring(deltaY, {
         toValue: _height.current,
         speed: SPEED,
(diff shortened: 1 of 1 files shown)


src/expandableCalendar/index.tsx
```diff
@@ -384,22 +384,21 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
...
+        onCalendarToggled?.(_isOpen);
+        setPosition(() => _height.current === closedHeight ? Positions.CLOSED : Positions.OPEN);
+      });
       closeHeader(_isOpen);
-      resetWeekCalendarOpacity(_isOpen);
     }
   };
 
-  const resetWeekCalendarOpacity = (isOpen: boolean) => {
+  const resetWeekCalendarOpacity = async (isOpen: boolean) => {
```
Inbal-Tish, line 401: Why async if no one awaits?

bravesoul349: It appears that `resetWeekCalendarOpacity` is causing a minor delay and occasional fluctuations in the animation, likely due to blocking the main thread. As a result, both processes are now being executed simultaneously.

Inbal-Tish (approved):

PR #2507, opened 2024-08-15
ios - add workaround for React.codegen error

Inbal-Tish:
iOS - Podfile - add workaround for React.codegen deployment target is 11.0 while code written with iOS 12.0 feature fails the build


ethanshar (approved):
Approved. 
But not sure I understand how this related to React codegen?

Inbal-Tish:
> Approved. But not sure I understand how this related to React codegen?

React codegen pod is using code available in ios 12.0 but it's min deployment target is iOS 11.0 which errors when trying to build from xcode.

PR #2515, opened 2024-08-26
feat(infiniteAgendaList): add the ability to set item height per item

yakirza17:
In the infinite agenda list, the item height is predefined. We need a way to set multiple item sizes per elements (and not only for title and item)

### DIFF
--- a/example/src/mocks/agendaItems.ts
+++ b/example/src/mocks/agendaItems.ts
@@ -28,7 +28,7 @@ function getPastDate(numberOfDays: number) {
 export const agendaItems = [
   {
     title: dates[0],
-    data: [{hour: '12am', duration: '1h', title: 'First Yoga'}]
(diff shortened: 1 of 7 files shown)


Inbal-Tish:
@yakirza17 Can you please add usage example to the demo?

example/src/mocks/agendaItems.ts
```diff
@@ -28,7 +28,7 @@ function getPastDate(numberOfDays: number) {
 export const agendaItems = [
   {
     title: dates[0],
-    data: [{hour: '12am', duration: '1h', title: 'First Yoga'}]
+    data: [{hour: '12am', duration: '1h', title: 'First Yoga'}, {hour: '9am', duration: '1h', title: 'Long Yoga', itemType: 'Long Event'}],
```
Inbal-Tish, line 31: Better make the itemType with no spaces, like 'Long' or 'LongEvent'

Inbal-Tish (approved):

Inbal-Tish:
@yakirza17 I'm approving but please see my little comment

PR #2517, opened 2024-08-30
day base style in style.ts

anilcivan:
With this code, you will be able to edit the size of the day boxes via the theme object.


Inbal-Tish:
@anilcivan Thank you for the PR, can you please add usage example to the CalendarsScreen?

PR #2524, opened 2024-09-08
Add disableMonthScroll

LeXuanKhanh:
The current expandable calendar is missing properties to disable scroll when it is in the opened state (like CalendarList). This implementation will resolve that issue and also provide some workaround for some bugs

### DIFF
--- a/src/expandableCalendar/index.tsx
+++ b/src/expandableCalendar/index.tsx
@@ -63,6 +63,8 @@ export interface ExpandableCalendarProps extends CalendarListProps {
   allowShadow?: boolean;
   /** whether to disable the week scroll in closed position */
   disableWeekScroll?: boolean;
+  /** whether to disable the month scroll in opened position */
+  disableMonthScroll?: boolean;
   /** a threshold for opening the calendar with the pan gesture */
   openThreshold?: number;
   /** a threshold for closing the calendar with the pan gesture */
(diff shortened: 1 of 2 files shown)


src/expandableCalendar/index.tsx
```diff
@@ -580,7 +583,7 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
         onDayPress={_onDayPress}
         onVisibleMonthsChange={onVisibleMonthsChange}
         pagingEnabled
-        scrollEnabled={isOpen}
+        scrollEnabled={isOpen && !disableMonthScroll}
```
Inbal-Tish, line 586: Since CalendarList props allow passing ScrollView's `scrollEnabled` I prefer to use this one instead of adding another prop to the API.
Let's try moving line 586 before the `{...others}` (line 579) to allow overriding the 'scrollEnabled' prop by passing it instead of 'disableMonthScroll' to the ExpandableCalendar component.

LeXuanKhanh: Do you think it will confuse the user because we already had the `disableWeekScroll` prop ?  They will think that using the prop `scrollEnabled` will disable scrolling behavior for both states opened and closed, not only opened. For that, I think we should use another prop for clarity.
If you insist on making the change in your opinion please let me know.

Inbal-Tish: I understand your point, but in the case of 'disableWeekScroll' we change the component from WeekCalendar to a simple Week component. In your implementation you're not replacing the CalendarList with a simple Calendar component (which is an option), but only passing it to the CalendarList's FlatList implementation.

LeXuanKhanh: I tried to do the same logic as the WeekCalendar but seems using the normal Calendar component it will break. So in this case must we use the `scrollEnabled` ?

LeXuanKhanh: any update @Inbal-Tish ?

Issue #2526, opened 2024-09-11
[Bug] accessibilityRole in Header not variable


Abhisflyingsoon:
## Description

Accessibility property "accessibilityRole" in Header should not be hardcoded to "adjustable". It should be variable.
Noticed weird behaviour, when scrolling with one finger while the header was focused. It somehow ignored every boundary I set and showed duplicate months in a list. Changing the accessibilityRole fixed the issue for us.

PR #2527, opened 2024-09-11
Fixes issue #2526

Abhisflyingsoon:
This should fix the issue of the hardcoded accessbilityRole issue in the header component. It was causing problem with the Voicover/Talkback Screenreader functionality, it was suggesting wrong accessbilityActions when the header was in focus.

### DIFF
--- a/src/calendar/header/index.tsx
+++ b/src/calendar/header/index.tsx
@@ -13,7 +13,8 @@ import {
   ViewStyle,
   AccessibilityActionEvent,
   ColorValue,
-  Insets
+  Insets,
+  AccessibilityRole
 } from 'react-native';
 import {formatNumbers, weekDayNames} from '../../dateutils';
 import styleConstructor from './style';
@@ -68,6 +69,8 @@ export interface CalendarHeaderProps {
   current?: string;
   /** Left inset for the timeline calendar header, default is 72 */
   timelineLeftInset?: number;
+  /** Accessibility role of the header component */
+  accessibilityRole?: AccessibilityRole;
 }
 
 const accessibilityActions = [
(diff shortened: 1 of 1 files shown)


Abhisflyingsoon:
[Fixes issue](https://github.com/wix/react-native-calendars/pull/2527/commits/f8f31def9599e311754cea9da0244c4e8b1a22a0) https://github.com/wix/react-native-calendars/issues/2526

src/calendar/header/index.tsx
```diff
@@ -102,7 +102,8 @@ const CalendarHeader = forwardRef((props: CalendarHeaderProps, ref) => {
     importantForAccessibility,
     numberOfDays,
     current = '',
-    timelineLeftInset
+    timelineLeftInset,
+    accessibilityRole
```
Inbal-Tish, line 106: Where this prop is declared?

src/calendar/header/index.tsx
```diff
@@ -279,7 +280,7 @@ const CalendarHeader = forwardRef((props: CalendarHeaderProps, ref) => {
       testID={testID}
       style={propsStyle}
       accessible
-      accessibilityRole={'adjustable'}
+      accessibilityRole={accessibilityRole}
```
Inbal-Tish, line 283: You're removing the existing value without even set it as the default...

Inbal-Tish:
@Abhisflyingsoon Thank you for the PR. Please add a description to summarize the issue and your solution to it.

src/calendar/header/index.tsx
Abhisflyingsoon: defined in the props. Did you mean that?

src/calendar/header/index.tsx
Abhisflyingsoon: added fallback value to "adjustable"

PR #2555, opened 2024-11-18
Fixes stale data in Agenda

edumana:
This PR fixes an issue where Agenda fails to re-render when there is a change in items and the _selected_ day remains the same. Currently the only fix is to programmatically switch _selected_ to another day and set it back to the original day to force Agenda to re-render items correctly.


mbifulco:
Hi team - wondering if there's anything we can do to help with this proposal!

mbifulco:
Just a bump here - happy to talk through an alternative approach if that's more apropos?

PR #2557, opened 2024-11-21
Calendar - add 'disabledByWeekDays' prop

Inbal-Tish:
Calendar - add 'disabledByWeekDays' prop + cleaning props

mads-4545

### DIFF
--- a/example/src/screens/calendarScreen.tsx
+++ b/example/src/screens/calendarScreen.tsx
@@ -37,7 +37,7 @@ const CalendarScreen = () => {
   const renderCalendarWithSelectableDate = () => {
     return (
       <Fragment>
-        <Text style={styles.text}>Calendar with selectable date</Text>
+        <Text style={styles.text}>Selectable date</Text>
         <Calendar
           testID={testIDs.calendars.FIRST}
           enableSwipeMonths
@@ -50,11 +50,11 @@ const CalendarScreen = () => {
     );
   };
 
-  const renderCalendarWithWeekNumbers = () => {
+  const renderCalendarWithWeekNumbersAndSpinner = () => {
     return (
       <Fragment>
-        <Text style={styles.text}>Calendar with week numbers</Text>
-        <Calendar style={styles.calendar} hideExtraDays showWeekNumbers/>
+        <Text style={styles.text}>Week numbers and spinner</Text>
+        <Calendar style={styles.calendar} hideExtraDays showWeekNumbers displayLoadingIndicator/>
       </Fragment>
     );
   };
@@ -62,7 +62,7 @@ const CalendarScreen = () => {
   const renderCalendarWithMinAndMaxDates = () => {
     return (
       <Fragment>
-        <Text style={styles.text}>Calendar with min and max dates</Text>
+        <Text style={styles.text}>Min and max dates</Text>
         <Calendar
           style={styles.calendar}
           hideExtraDays
@@ -78,7 +78,7 @@ const CalendarScreen = () => {
   const renderCalendarWithMarkedDatesAndHiddenArrows = () => {
     return (
       <Fragment>
-        <Text style={styles.text}>Calendar with marked dates and hidden arrows</Text>
+        <Text style={styles.text}>Marked dates and hidden arrows</Text>
         <Calendar
           style={styles.calendar}
           current={INITIAL_DATE}
@@ -92,7 +92,6 @@ const CalendarScreen = () => {
             [getDate(10)]: {disabled: true, activeOpacity: 0, disableTouchEvent: false}
           }}
           hideArrows={true}
-          // disabledByDefault={true}
         />
       </Fragment>
     );
@@ -101,7 +100,7 @@ const CalendarScreen = () => {
   const renderCalendarWithMultiDotMarking = () => {
     return (
       <Fragment>
-        <Text style={styles.text}>Calendar with multi-dot marking</Text>
+        <Text style={styles.text}>Multi-dot marking</Text>
         <Calendar
           style={styles.calendar}
           current={INITIAL_DATE}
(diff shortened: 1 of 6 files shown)


src/calendar/calendar.api.json
```diff
@@ -165,7 +175,7 @@
     {
       "name": "disabledDaysIndexes",
       "type": "number[]",
-      "description": "Whether to apply custom disable color to selected day names indexes"
+      "description": "Whether to apply custom disable color to selected day names indexes (Sunday=0)"
```
nitzanyiz, line 178: if `firstDay` is not 0 then the Sunday is not 0.

src/calendar/calendar.api.json
```diff
@@ -92,9 +102,9 @@
       "description": "Whether to disable days by default"
     },
     {
-      "name": "allowSelectionOutOfRange",
-      "type": "boolean",
-      "description": "Whether to allow selection of dates before minDate or after maxDate"
+      "name": "disabledByWeekDays",
+      "type": "number[]",
+      "description": "Disable dates by days of the week (Sunday=0)"
```
nitzanyiz, line 107: Same here. If for example, we set `firstDay` to be Monday then passing `disabledByWeekDays={[1]}` disables Tuesday.

src/calendar/header/index.tsx
```diff
@@ -37,36 +41,37 @@ export interface CalendarHeaderProps {
   hideDayNames?: boolean;
   /** Hide month navigation arrows */
   hideArrows?: boolean;
-  /** Replace default arrows with custom ones (direction can be 'left' or 'right') */
+  /** Replace default arrows with custom ones (direction = 'left' || 'right') */
```
nitzanyiz, line 44: maybe same way as in typescript?
```suggestion
  /** Replace default arrows with custom ones (direction = 'left' | 'right') */
```

src/calendar/header/index.tsx
```diff
@@ -37,36 +41,37 @@ export interface CalendarHeaderProps {
...
   onPressArrowRight?: (method: () => void, month?: XDate) => void; //TODO: replace with string
-  /** Left & Right arrows. Additional distance outside of the buttons in which a press is detected, default: 20 */
+  /** Left & Right arrows. Additional distance outside of the buttons in which a press is detected. Default = 20 */
   arrowsHitSlop?: Insets | number;
   /** Disable left arrow */
   disableArrowLeft?: boolean;
   /** Disable right arrow */
   disableArrowRight?: boolean;
-  /** Apply custom disable color to selected day names indexes */
+  /** Apply custom disable color to selected day names indexes (Sunday=0) */
```
nitzanyiz, line 56: Same here with the `firstDay`.

src/calendar/index.tsx
```diff
@@ -301,9 +298,6 @@ Calendar.propTypes = {
   ...CalendarHeader.propTypes,
   ...Day.propTypes,
   theme: PropTypes.object,
-  firstDay: PropTypes.number,
-  displayLoadingIndicator: PropTypes.bool,
-  showWeekNumbers: PropTypes.bool,
```
nitzanyiz, line 306: Why did we remove this?

example/src/screens/calendarScreen.tsx
```diff
@@ -173,6 +172,9 @@ const CalendarScreen = () => {
             [getDate(26)]: {color: 'gray'},
             [getDate(27)]: {endingDay: true, color: 'gray'}
           }}
+          disabledDaysIndexes={[3]}
+          disabledByWeekDays={[3]}
+          disableAllTouchEventsForDisabledDays
```
nitzanyiz, line 177: A comment on this prop. onPress is still triggered when pressing a disabled date. Should we fix in this pr?

src/calendar/calendar.api.json
Inbal-Tish: That's a good point but not correct. Date().getDay() returns the day of the week index (Sun = 0) and the date-state-manager is checking against that in the disabledByWeekDays feature. So no matter what the first day is the date is always on the same day of the week (ex. 20-11-2024 is always Wednesday).

src/calendar/calendar.api.json
Inbal-Tish: No. No matter what is the first day, if you want to disable Mondays you should pass [1]

src/calendar/header/index.tsx
Inbal-Tish: Same answer...

src/calendar/index.tsx
Inbal-Tish: `showWeekNumbers` is a CalendarHeader's prop, and CalendarProps inherits from it so it gets it from there and there's no need to redefine it

example/src/screens/calendarScreen.tsx
Inbal-Tish: It only affects markingType = 'period' so I prefer to address it in a different PR since there are many gaps btw the basic day and the period day that we should sync

src/calendar/calendar.api.json
nitzanyiz: Just making sure, what do we expect to happen here?
```
firstDay={1}
disabledDaysIndexes={[0]}
```
This is the result. Monday, which is in index 0, is shown as disabled.
image

src/calendar/calendar.api.json
nitzanyiz: This is what happens if (Tuesday is disabled)
```
firstDay={1}
disabledDaysIndexes={[1]}
```
image

src/calendar/index.tsx
nitzanyiz: I'm not sure if they're set automatically somehow but I don't see that `CalendarHeader` even have `propTypes` set anywhere.

src/calendar/calendar.api.json
Inbal-Tish: If you remove the markingType prop, I'll expect that the first day displayed will be Monday and alll the Sundays will be disabled.
You're checking on a period marking... as mentioned, there's an issue since basic day and period day are not the same.

src/calendar/index.tsx
Inbal-Tish: oh you mean the prop type... well, this should be removed all over but there was an issue there that I need to confirm first

src/calendar/calendar.api.json
Inbal-Tish: Sunday have to be 0. This is a bug that need to be fixed, but fixing it is a breaking change so I'll add it to the next major version

nitzanyiz (approved):

PR #2559, opened 2024-11-24
Fix/period day styling

Inbal-Tish:
PeriodDay - sync styling with BasicDay

Note: The default for disabled and inactive dot colors were changed.


nitzanyiz (approved):

deezyomar:
Is it possible to hide the week view on the calendar? I want users to be able to pan to view the week and then the month, rather than having the week view displayed by default.

The week view is displayed by default.

Screenshot 2025-01-06 at 6 47 35 PM

PR #2588, opened 2025-01-19
Added forwardRef to AgendaList

nitzanyiz:
Added forwardRef to AgendaList (This is needed to support it with the `CollapsingHeader` component). Currently it throws a warning and the `CollapsingHeader` doesn't work well.

### DIFF
--- a/src/expandableCalendar/agendaList.tsx
+++ b/src/expandableCalendar/agendaList.tsx
@@ -19,7 +19,7 @@ import {
   ViewToken,
 } from 'react-native';
 
-import {useDidUpdate} from '../hooks';
+import {useDidUpdate, useCombinedRefs} from '../hooks';
 import {getMoment} from '../momentResolver';
 import {isToday, isGTE, sameDate} from '../dateutils';
 import {parseDate} from '../interface';
@@ -41,7 +41,7 @@ const viewabilityConfig = {
  * @extends: SectionList
  * @example: https://github.com/wix/react-native-calendars/blob/master/example/src/screens/expandableCalendar.js
  */
-const AgendaList = forwardRef((props: AgendaListProps) => {
+const AgendaList = forwardRef((props: AgendaListProps, ref: any) => {
   if (props.infiniteListProps) {
     return <InfiniteAgendaList {...props} />;
   }
@@ -69,7 +69,7 @@ const AgendaList = forwardRef((props: AgendaListProps) => {
   const {date, updateSource, setDate, setDisabled} = useContext(Context);
 
   const style = useRef(styleConstructor(theme));
-  const list = useRef<any>();
+  const list = useCombinedRefs(ref);
   const _topSection = useRef(sections[0]?.title);
   const didScroll = useRef(false);
   const sectionScroll = useRef(false);
@@ -228,6 +228,7 @@ const AgendaList = forwardRef((props: AgendaListProps) => {
     <SectionList
       stickySectionHeadersEnabled
       {...props}
+      // @ts-expect-error
       ref={list}
       keyExtractor={_keyExtractor}
       showsVerticalScrollIndicator={false}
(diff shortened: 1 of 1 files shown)


src/expandableCalendar/agendaList.tsx
```diff
@@ -255,6 +255,6 @@ AgendaList.propTypes = {
   dayFormatter: PropTypes.func,
   useMoment: PropTypes.bool,
   markToday: PropTypes.bool,
-  sectionStyle: PropTypes.oneOfType([PropTypes.object, PropTypes.number, PropTypes.array]),
```
Inbal-Tish, line 258: Style is a number when it comes from a StyleSheet

src/expandableCalendar/agendaList.tsx
```diff
@@ -41,7 +41,7 @@ const viewabilityConfig = {
  * @extends: SectionList
  * @example: https://github.com/wix/react-native-calendars/blob/master/example/src/screens/expandableCalendar.js
  */
-const AgendaList = (props: AgendaListProps) => {
+const AgendaList = forwardRef((props: AgendaListProps) => {
```
Inbal-Tish, line 44: Why youre not using the 'ref' you got now?

src/expandableCalendar/agendaList.tsx
nitzanyiz: I don't think it does. When I added the `forwardRef` I got ts errors that the type of section style isn't good.
When I printed out the typeof the style object I got this output:
`Nitzan - Typeof StylSheet object object {"marginVertical": 16}`
With this code:
```tsx
import React from 'react';
import {StyleSheet} from 'react-native';
import {Navigation} from 'react-native-navigation';
import {StaticOptions, TopBar} from 'rnn-copilot';
// import myEngineInstance from '../../engineDemo/engineInstance';
import {Assets, Spacings, Colors, asWixScreen, ThemeView, View, Text, Button, Icon} from '../../src';

export default function PlaygroundScreen() {
  console.log(`Nitzan - Typeof StylSheet object`, typeof styles.playgroundLabel, styles.playgroundLabel);
  return (
    <ThemeView flex padding-page>
      <Text headingL center style={styles.playgroundLabel}>
        Playground
      </Text>
    </ThemeView>
  );
}

PlaygroundScreen.options = new StaticOptions().withTopBar(new TopBar().withTitle('Playground')).get();

const styles = StyleSheet.create({
  playgroundLabel: {
    marginVertical: Spacings.s4
  }
});

Navigation.registerComponent('example.PlaygroundScreen', () => asWixScreen(PlaygroundScreen));

```

src/expandableCalendar/agendaList.tsx
nitzanyiz: Sorry. I missed a commit I thought I pushed. added now.

src/expandableCalendar/agendaList.tsx
Inbal-Tish: That's good news. It was a bit confusing: https://stackoverflow.com/questions/41483862/how-are-styles-mapped-to-numbers-in-react-native

Inbal-Tish: Can we tell when was this change?

Inbal-Tish: As we discussed, let's put it back and ts ignore it. I'll try to find out why this error only pops up when wrapping with forwardRef

Inbal-Tish (approved):

nitzanyiz:
@Inbal-Tish Do you think you will have some changes in calendars by the next release (maybe for widgets or something)? I might hf this if not, so we won't have to release next week.

Inbal-Tish:
> @Inbal-Tish Do you think you will have some changes in calendars by the next release (maybe for widgets or something)? I might hf this if not, so we won't have to release next week.

I have a task here but I don't know what's the plan for next week. Anyways we can hotfix it anytime

PR #2257, opened 2023-06-17
Added descending order to CalendarList

biddano:
This PR will allow users the flexibility to view CalendarList months in descending order rather than ascending order (current functionality). Added an optional descendingOrder prop to CalendarList for this (defaults to false).

PR #2594, opened 2025-02-03
Added inverted prop to invert the calendar list order

jcedborger:
This is a copy of this stale PR: https://github.com/wix/react-native-calendars/pull/2257
I needed the functionality, so I remade it and thought contributing would be good.

Adding an `inverted` prop, to make the CalendarList reverse order.


jcedborger:
@Inbal-Tish, is this something you could have a look at? :)

KurnosovNikita:
Would love to see this merged — really need the inverted prop. Could someone review it? Thanks!

PR #2598, opened 2025-02-09
TimelineList - fix for RTL under RN73

Inbal-Tish:
TimelineList - fix for RTL under RN73

### DIFF
--- a/src/infinite-list/index.tsx
+++ b/src/infinite-list/index.tsx
@@ -51,6 +51,7 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
     disableScrollOnDataChange,
     onEndReachedThreshold,
     onVisibleIndicesChanged,
+    disableRecycling,
     layoutProvider,
     onScroll,
     onEndReached,
@@ -70,8 +71,9 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
       }
     )
   );
-  const shouldUseAndroidRTLFix = useMemo(() => {
-    return constants.isAndroidRTL && isHorizontal;
+
+  const shouldFixRTL = useMemo(() => {
+    return isHorizontal && (disableRecycling || constants.isAndroidRTL);
   }, []);
 
   const listRef = useCombinedRefs(ref);
@@ -87,7 +89,7 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
     }
 
     setTimeout(() => {
-      const x = isHorizontal ? constants.isAndroidRTL ? Math.floor(data.length / 2) + 1 : Math.floor(data.length / 2) * pageWidth : 0;
+      const x = isHorizontal ? shouldFixRTL ? Math.floor(data.length / 2) + 1 : Math.floor(data.length / 2) * pageWidth : 0;
       const y = isHorizontal ? 0 : positionIndex * pageHeight;
       // @ts-expect-error
       listRef.current?.scrollToOffset?.(x, y, false);
@@ -100,7 +102,7 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
 
       const contentOffset = event.nativeEvent.contentOffset;
       const y = contentOffset.y;
-      const x = shouldUseAndroidRTLFix ? (pageWidth * data.length - contentOffset.x) : contentOffset.x;
+      const x = shouldFixRTL ? (pageWidth * data.length - contentOffset.x) : contentOffset.x;
       const newPageIndex = Math.round(isHorizontal ? x / pageWidth : y / pageHeight);
       if (pageIndex.current !== newPageIndex) {
         if (pageIndex.current !== undefined) {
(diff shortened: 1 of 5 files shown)


src/infinite-list/index.tsx
```diff
@@ -70,8 +71,9 @@ const InfiniteList = (props: InfiniteListProps, ref: any) => {
       }
     )
   );
-  const shouldUseAndroidRTLFix = useMemo(() => {
-    return constants.isAndroidRTL && isHorizontal;
+
+  const shouldFixRTL = useMemo(() => {
+    return isHorizontal && (disableRecycling || constants.isAndroidRTL);
```
ethanshar, line 76: Are you sure you need here the `disableRecycling` - why  `disableRecycling` affect the RTL fix 
Especially since you do disabled the recycling later on here
```
 disableRecycling={shouldFixRTL || disableRecycling}
```

src/timeline-list/index.tsx
```diff
@@ -49,17 +49,25 @@ export interface TimelineListProps {
...
+  infiniteListProps?: {
+    disableRecycling?: boolean;
+  }
 }
 
 const TimelineList = (props: TimelineListProps) => {
-  const {timelineProps, events, renderItem, showNowIndicator, scrollToFirst, scrollToNow, initialTime} = props;
+  const {timelineProps, events, renderItem, showNowIndicator, scrollToFirst, scrollToNow, initialTime, infiniteListProps} = props;
+  const disableRecycling = infiniteListProps?.disableRecycling;
+  const shouldFixRTL = disableRecycling || constants.isAndroidRTL; // isHorizontal = true
```
ethanshar, line 63: Same here, for the user It's not clear that the prop purpose is to fix RTL.

src/timeline-list/index.tsx
```diff
@@ -166,6 +173,7 @@ const TimelineList = (props: TimelineListProps) => {
       scrollViewProps={{
         onMomentumScrollEnd
       }}
+      disableRecycling={disableRecycling}
```
ethanshar, line 176: consider spreading {...infiniteListProps}

ethanshar (commented):
Wrote a few comments, let me know If I misunderstood something about the disableRecycling prop 
I understand that it's required to make RTL work, but not sure it should trigger the RTL fix. Am I missing something?

src/infinite-list/index.tsx
Inbal-Tish: Well, now that I fixed the x when passing disableRecycling I really don't need to actually disable the recycling... so I need another prop.
The point was that if disabling it is the fix I'll let the user pass the prop that controls it, and once the user pass it I'll apply it to the recyclerList AND fix the rtl issue that it causes.
Since I don't want to add a prop that will be redundant after the RN73 upgrade, what do you suggest can control this from the users side?

src/timeline-list/index.tsx
Inbal-Tish: The prop purpose is not fixing RTL issue, the fix for RTL issue was to disable recycling... but now not anymore...

Inbal-Tish:
@ethanshar We need to consider using RN version check to apply this fix instead of adding a prop to control it from the client side

Inbal-Tish:
> > @ethanshar We need to consider using RN version check to apply this fix instead of adding a prop to control it from the client side (unless we decide to upgrade RN version and make it the minimum)
> 
> We can start with that, but like you wrote the right solution will be to upgrade RN and release new major version
> 
> How much effort will it be to upgrade RN and remove dependency on RNN? Instead of upgrading we can create a new RN project (with RN 73) and override our existing native folders - I do it all the time and it's fairly simple - let's talk about it

Yes, creating a new demo will be best and fastest. I do think this is the right solution and maybe a good chance to clean the library from all the propTypes and other constrains, that's why it will take a bit longer.

ethanshar:
Let's focus only on upgrading the demo and fixing the RTL issues 
We don't have time for extra cleaning ATM

ethanshar (approved):
Approved. 
Let's start with upgrading the demo app as discussed right away

src/timeline-list/index.tsx
```diff
@@ -53,13 +53,14 @@ export interface TimelineListProps {
 
 const TimelineList = (props: TimelineListProps) => {
   const {timelineProps, events, renderItem, showNowIndicator, scrollToFirst, scrollToNow, initialTime} = props;
+  const shouldFixRTL = constants.isRTL && (constants.isRN73() || constants.isAndroid); // isHorizontal = true
```
yuvalsho, line 56: consider deleting commented out code

yuvalsho (approved):

src/timeline-list/index.tsx
Inbal-Tish: This comment comes to hit that we assume the list is in horizontal layout since the TimelineList only supports horizontal

PR #2599, opened 2025-02-09
ExpandableCalendar - Added height calculation on layout instead of hard coded header height

nitzanyiz:
## Description
Added dynamic header height calculation. The header height had a fixed height until now causing the `WeekCalendar` to go on top of the header and cut some of the calendar header. We've encountered this where the `WeekCalendar` was cutting the days letters.

**Ticket: MADS-4584**


nitzanyiz:
@Inbal-Tish , I've passed the onLayout to the headers - Calendar's header when there is not staticHeader and static header when there is. I Also moved the WeekCalendar to be on top of the AnimatedHeader because it was cutting it. Do you think I should add the onHeaderLayout to the animated header instead? They should be same... I'm not sure which one is better to relay on.

nitzanyiz:
Do you have any intuition on things I should check here?

src/calendar-list/index.tsx
```diff
@@ -250,12 +252,13 @@ const CalendarList = (props: CalendarListProps & ContextProp, ref: any) => {
         calendarHeight={calendarHeight}
         scrollToMonth={scrollToMonth}
         visible={isDateInRange(item)}
+        onHeaderLayout={!shouldUseStaticHeader ? onHeaderLayout : undefined}
```
Inbal-Tish, line 255: We only need it to be called once, not on every item (by default it will be called 101 times...)

src/expandableCalendar/index.tsx
```diff
@@ -205,17 +207,17 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
 
   /** Styles */
 
-  const style = useRef(styleConstructor(theme));
+  const style = useMemo(() => styleConstructor(theme, {headerHeight}), [theme, headerHeight]);
   const themeObject = Object.assign(headerStyleOverride, theme);
 
-  const _wrapperStyles = useRef({style: {height: startHeight}});
-  const _headerStyles = {style: {top: isOpen ? -HEADER_HEIGHT : 0}};
+  const _wrapperStyles = useMemo(() => ({style: {height: startHeight}}), [headerHeight]);
```
Inbal-Tish, line 213: Why _wrapperStyles depends on headerHeight?

src/expandableCalendar/index.tsx
```diff
@@ -205,17 +207,17 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
 
   /** Styles */
 
-  const style = useRef(styleConstructor(theme));
+  const style = useMemo(() => styleConstructor(theme, {headerHeight}), [theme, headerHeight]);
```
Inbal-Tish, line 210: The theme and headerHeight are not dynamic and so is the style. Since the height is set after the onLayout, why not pass it to the weekCalendarStyle directly instead of injecting to the style's weekContainer?

src/expandableCalendar/index.tsx
```diff
@@ -229,25 +231,25 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
...
   const weekCalendarStyle = useMemo(() => {
-    return [style.current.weekContainer, isOpen ? style.current.hidden : style.current.visible];
-  }, [isOpen]);
+    return [style.weekContainer, isOpen ? style.hidden : style.visible];
+  }, [isOpen, style]);
 
   const containerStyle = useMemo(() => {
-    return [allowShadow && style.current.containerShadow, propsStyle];
-  }, [allowShadow, propsStyle]);
+    return [allowShadow && style.containerShadow, propsStyle, headerHeight === 0 && {opacity: 0}];
```
Inbal-Tish, line 251: Use the style.current.hidden instead of the object

src/expandableCalendar/index.tsx
```diff
@@ -608,9 +611,9 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
       ) : (
         <Animated.View testID={`${testID}.expandableContainer`} ref={wrapper} style={wrapperStyle} {...panResponder.panHandlers}>
           {renderCalendarList()}
-          {renderWeekCalendar()}
           {!hideKnob && renderKnob()}
           {!horizontal && renderAnimatedHeader()}
+          {renderWeekCalendar()}
```
Inbal-Tish, line 616: I think we need to understand why the header is cutting the weekCalendar instead of moving it to the top.

src/calendar-list/index.tsx
nitzanyiz: Hmm you're right. I'm not sure what I can do that its not ugly here because the one that is called first is not the the first or some predefined index.

src/expandableCalendar/index.tsx
nitzanyiz: Mistake. For some reason if I remove this the drag animation is not working.

src/calendar-list/index.tsx
nitzanyiz: I could do something like this? (Not pretty though)
```tsx
    const shouldInvokeOnHeaderLayout = useRef(true);
    .
    .
    .
    const onHeaderLayoutToPass = shouldInvokeOnHeaderLayout.current ? onHeaderLayout : undefined;
    shouldInvokeOnHeaderLayout.current = false;
    return (
      <CalendarListItem
        {...calendarProps}
        testID={testId}
        markedDates={getMarkedDatesForItem(item)}
        item={item}
        style={calendarStyle}
        // @ts-expect-error - type mismatch - ScrollView's 'horizontal' is nullable
        horizontal={horizontal}
        calendarWidth={calendarWidth}
        calendarHeight={calendarHeight}
        scrollToMonth={scrollToMonth}
        visible={isDateInRange(item)}
        onHeaderLayout={!shouldUseStaticHeader ? onHeaderLayoutToPass : undefined}
      />
```

src/expandableCalendar/index.tsx
nitzanyiz: I've changed things a little according to what you said here. I removed the injecting of the header height to the style constructor. Maybe we should change `HEADER_HEIGHT` to be `DEFAULT_HEADER_HEIGHT`. And made the weekCalendarStyle change based on the headerHeight.

src/expandableCalendar/index.tsx
nitzanyiz: The change also seem to solve this.

src/expandableCalendar/index.tsx
Inbal-Tish: You can try it if it doesn't make the UI jump when updated...

src/expandableCalendar/useHeaderHeight.ts
```diff
@@ -0,0 +1,16 @@
+import {useCallback, useState} from 'react';
+import type {LayoutChangeEvent} from 'react-native';
+
+const useHeaderHeight = () => {
```
Inbal-Tish, line 4: If you think the hook is necessary here please place it in hooks/helpers directory

src/expandableCalendar/index.tsx
Inbal-Tish: So the drag animation is not working now?

src/calendar-list/index.tsx
Inbal-Tish: oh you mean set it to false after first reading it? Does it work?

src/expandableCalendar/useHeaderHeight.ts
nitzanyiz: I removed it. I thought it would a more complex functionality but seem like to wont.

src/expandableCalendar/index.tsx
nitzanyiz: I meant just to rename it

src/calendar-list/index.tsx
Inbal-Tish: What about measuring the static header with opacity 0 idea? Do you think it might be cleaner?

nitzanyiz: Seems to work also. It does call the `onLayout` 4 times at the beginning but works. Do we prefer that? It is simpler but I'm not sure about performance. The component is heavy by itself...

Inbal-Tish: What do you mean performant-wise to calculate the header height or to do it from the static header? Why it is called 4 times for the static header? Is it called once when you pass it to the Item?

nitzanyiz: I mean that now we have to render the static header regardless of its in horizontal mode and that staticHeader is passed so it might hurt performance. I'm not sure why its called 4 times because I just pass it to the underlaying view in the staticHeader. It doesn't cause a rerender because `onHeaderLayout` is called with the same value. The static header onLayout is called 4 times in master as well without the changes we made here in the branch.

src/expandableCalendar/index.tsx
Inbal-Tish: And where do you use the DEFAULT_HEADER_HEIGHT?

nitzanyiz: I don't. There are other places we use them

src/calendar-list/index.tsx
Inbal-Tish: Don't forget to add the flag and default height

nitzanyiz: Added 👌

nitzanyiz:
@Inbal-Tish, as we've talked, I've reverted the default header height change and made it 0 by default so now it will be hidden while the `headerHeight` is 0.

src/calendar-list/index.tsx
```diff
@@ -265,6 +271,7 @@ const CalendarList = (props: CalendarListProps & ContextProp, ref: any) => {
           addMonth={addMonth}
           accessibilityElementsHidden={true} // iOS
           importantForAccessibility={'no-hide-descendants'} // Android
+          onHeaderLayout={onHeaderLayout}
```
Inbal-Tish, line 274: How about passing `onHeaderLayoutToPass` here too? You said it is called 4 times...

src/expandableCalendar/index.tsx
```diff
@@ -238,16 +243,16 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
...
   const animatedHeaderStyle = useMemo(() => {
-    return [style.current.header, {height: HEADER_HEIGHT + 10, top: headerDeltaY.current}];
-  }, [headerDeltaY.current]);
+    return [style.current.header, {height: headerHeight, top: headerDeltaY.current}];
+  }, [headerDeltaY.current, headerHeight]);
 
   const weekCalendarStyle = useMemo(() => {
-    return [style.current.weekContainer, isOpen ? style.current.hidden : style.current.visible];
-  }, [isOpen]);
+    return [style.current.weekContainer, isOpen ? style.current.hidden : style.current.visible, {top: headerHeight}];
```
Inbal-Tish, line 250: You didn't remove the HEADER_HEIGHT from the style and the weekContainer.top

src/expandableCalendar/style.ts
```diff
@@ -96,7 +96,6 @@ export default function styleConstructor(theme: Theme = {}) {
       position: 'absolute',
       left: 0,
       right: 0,
-      top: HEADER_HEIGHT + (constants.isAndroid ? 8 : 9), // align row on top of calendar's first row
```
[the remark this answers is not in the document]
Inbal-Tish: Was not removed from line 7

src/calendar-list/index.tsx
Inbal-Tish: Ping

src/expandableCalendar/style.ts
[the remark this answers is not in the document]
nitzanyiz: Thought it was used somewhere else. Removed 👌

src/calendar-list/index.tsx
nitzanyiz: Missed that. Added.

src/expandableCalendar/index.tsx
```diff
@@ -370,7 +378,7 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
     onPanResponderMove: handlePanResponderMove,
     onPanResponderRelease: handlePanResponderEnd,
     onPanResponderTerminate: handlePanResponderEnd
-  }) : PanResponder.create({}), [numberOfDays, position]);
+  }) : PanResponder.create({}), [numberOfDays, position, headerHeight]);
```
Inbal-Tish, line 381: Please add a comment here for why the headerHeight is in the dep array

src/calendar-list/index.tsx
```diff
@@ -101,16 +101,19 @@ const CalendarList = (props: CalendarListProps & ContextProp, ref: any) => {
...
 
   const calendarProps = extractCalendarProps(props);
   const headerProps = extractHeaderProps(props);
   const calendarSize = horizontal ? calendarWidth : calendarHeight;
+  const shouldUseStaticHeader = staticHeader && horizontal;
 
   const [currentMonth, setCurrentMonth] = useState(parseDate(current));
 
   const shouldUseAndroidRTLFix = useMemo(() => constants.isAndroidRTL && horizontal, [horizontal]);
+  const shouldMeasureHeader = useRef(true);
```
Inbal-Tish, line 116: Please leave a comment here to explain why we're using this ref

src/calendar-list/index.tsx
```diff
@@ -250,12 +255,15 @@ const CalendarList = (props: CalendarListProps & ContextProp, ref: any) => {
...
         visible={isDateInRange(item)}
+        onHeaderLayout={onHeaderLayoutToPass}
       />
     );
   }, [horizontal, calendarStyle, calendarWidth, testID, getMarkedDatesForItem, isDateInRange, calendarProps]);
 
   const renderStaticHeader = () => {
-    if (staticHeader && horizontal) {
+    if (shouldUseStaticHeader) {
+      const onHeaderLayoutToPass = shouldMeasureHeader.current ? onHeaderLayout : undefined;
```
Inbal-Tish, line 265: You're checking 'shouldUseStaticHeader' twice...

src/expandableCalendar/index.tsx
```diff
@@ -175,20 +179,24 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
     if (!horizontal) {
       return Math.max(constants.screenHeight, constants.screenWidth);
     }
-    return CLOSED_HEIGHT + (WEEK_HEIGHT * (numberOfWeeks.current - 1)) + (hideKnob ? 12 : KNOB_CONTAINER_HEIGHT) + (constants.isAndroid ? 3 : 0);
+    return headerHeight + (WEEK_HEIGHT * (numberOfWeeks.current)) + (hideKnob ? 12 : KNOB_CONTAINER_HEIGHT) + (constants.isAndroid ? 3 : 0);
```
Inbal-Tish, line 182: Do we still need the Android fix here?

nitzanyiz: I don't know what its for. I guess now we kind of know that the `headerHeight` is bigger than the default value that was before so I guess we can remove this... Wdyt?

Inbal-Tish: Let's check. I would also want to know what the 12 is for and move it to a const to make it clearer

nitzanyiz: This seems to work
`return headerHeight + (WEEK_HEIGHT * (numberOfWeeks.current)) + (hideKnob ? 0 : KNOB_CONTAINER_HEIGHT)` and feels more right.

Inbal-Tish (approved):

boris-szl:
@KrisLau where you able to resolve the issue? I face the exact same problem. Upgrading to 1.1310.0 from 1.309.0 introduced the blank calendar....

KrisLau:
@boris-szl nope I haven't heard anything back from the maintainers

nitzanyiz:
Hi @boris-szl, @KrisLau, sorry for the late reply.
Can you give me a snippet to reproduce this?

PR #2604, opened 2025-02-20
Accessibility - add header reader after month change

Inbal-Tish:
Accessibility - add header reader after month change


src/calendar-list/index.tsx
```diff
@@ -263,8 +267,8 @@ const CalendarList = (props: CalendarListProps & ContextProp, ref: any) => {
           style={staticHeaderStyle}
           month={currentMonth}
           addMonth={addMonth}
-          accessibilityElementsHidden={true} // iOS
-          importantForAccessibility={'no-hide-descendants'} // Android
+          accessibilityElementsHidden={accessibilityElementsHidden} // iOS
```
nitzanyiz, line 270: This doesn't seem to focus correctly on ios now. Lets check after the RN0.73 upgrade your doing 👌

Inbal-Tish:
@nitzanyiz Let's try it now after the RN73 upgrade

nitzanyiz:
When we focus the calendar header there are two containers in the header that can be focused both giving the same effect - the adjusting and reading on month change.
There is a big one and a smaller one.
When we are focused on the smaller one and then swipe up and the months changes it starts reading the month and then cuts off, focuses the big container and then reads again.
When we are focused on the larger container this bug doesn't occur and the month reads like it should.

nitzanyiz:
On android there are also the two containers but the cut reading is not happening. How ever it reads extra things something like "emtpy ttfs..."

Inbal-Tish:
@nitzanyiz When running on devices I get the Calendar view only and can't reproduce the mentioned double focus.
What I did notice is that on Android only it reads the day names (s, m, t etc.) but I could not find a way to ignore it.
I suggest to merge this fix and get to this issue separately.

nitzanyiz (approved):

PR #2607, opened 2025-02-25
RN73 upgrade

Inbal-Tish:
RN73 upgrade

Lint fix: https://github.com/wix/react-native-calendars/pull/2608


babel.config.js
```diff
@@ -1,3 +1,20 @@
...
+      {
+        extensions: ['.js', '.jsx', '.ts', '.tsx'],
+        root: ['.'],
+        alias: {
+          'react-native-calendars': './src/index.ts',
+          commons: './src/commons',
+          services: './src/services',
+          utils: './src/utils',
+          hooks: './src/hooks.ts',
+          style: './src/style.ts'
```
ethanshar, line 15: Are you using these other aliases in src files? (commons, services, etc..) if not better remove them
If you do (I didn't find usages), we need to make sure the built version of Calendars go thru this babel, otherwise it might break consumers, including us

example/package.json
ethanshar, line 1: I wonder if and why we have a package.json in the example folder? is it in use?

example/src/app.tsx
```diff
@@ -1,14 +1,12 @@
 import './wdyr'; // <--- must be first import
-import {Navigation} from 'react-native-navigation';
-// import {I18nManager} from 'react-native'; // <--- In order to test RTL
+import React from 'react';
+import {AppRegistry} from 'react-native';
 //@ts-expect-error
```
ethanshar, line 4: I figure that if you include this file in tsconfig, you won't receive TS errors about this

example/src/screens/menuScreen.tsx
ethanshar, line 1: Im getting some warnings (yellow highlights in vscode) on code formatting  
I think that you need to install prettier-eslint package so vscode would be able to format based on prettier+eslint rules

jest.config.js
ethanshar, line 1: Notice that tests don't pass at the moment, probably why the build fails 
I believe you are missing some configurations here. Specifically the `transformIgnorePatterns` (based on the tests errors I see)

ethanshar (commented):
Looks good though the build fails, see my comments

example/src/app.tsx
Inbal-Tish: True, but I have other issues with LocaleConfig that I didn't want to deal with now. I'll see if I have time for that, I'll try to fix them

ethanshar (approved):

PR #2608, opened 2025-02-26
fix lint errors (rn73 linter)

Inbal-Tish:
fix lint errors (rn73 linter)

### DIFF
--- a/src/expandableCalendar/agendaList.tsx
+++ b/src/expandableCalendar/agendaList.tsx
@@ -42,6 +42,10 @@ const viewabilityConfig = {
  * @example: https://github.com/wix/react-native-calendars/blob/master/example/src/screens/expandableCalendar.js
  */
 const AgendaList = forwardRef((props: AgendaListProps, ref: any) => {
+  if (props.infiniteListProps) {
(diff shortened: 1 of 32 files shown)


src/expandableCalendar/agendaList.tsx
```diff
@@ -42,10 +42,6 @@ const viewabilityConfig = {
  * @example: https://github.com/wix/react-native-calendars/blob/master/example/src/screens/expandableCalendar.js
  */
 const AgendaList = forwardRef((props: AgendaListProps, ref: any) => {
-  if (props.infiniteListProps) {
-    return <InfiniteAgendaList {...props} />;
-  }
-
```
nitzanyiz, line 48: You had to move this because of the conditional hooks?

nitzanyiz (approved):

src/expandableCalendar/agendaList.tsx
Inbal-Tish: Yes

Inbal-Tish: I think I'll revert it and add ignore on the other PR

nitzanyiz (approved):
👌

PR #2617, opened 2025-03-11
fix: Agenda dates disappearing on component load

### LINKED ISSUE #2591: Agenda dates disappearing on component load
Hi everyone, first time posting a question here so please bear with me.

I have been working on a project in react native where I want the user to be able to create, view, edit and delete calendar events. I have been implementing react-native-calendars' Agenda component. However, I have begun to run into an issue where when the app loads, the dates at the top disappear after about 1/2 seconds (as show in the screenshot). The dates would reappear once an event has been added or if i expand calendar and reset it to normal (single date row at the top)

My general process is that I have a function called initCalendar(). In this function data is retrieved from a database which is then refactored to match the required syntax for Agenda (that is with date [YYYY-MM-dd] as the key) and set to the state variable "events".

If you require any code snippets, let me know as I am not sure what would be important to share.

I would be grateful for any advice.

I originally had the events be refactored etc then passed into a custom component that holds the Agenda but i thought that may have been the reason but now that I have refactored it and tried checking the format of my event data after refactoring and all seems good. When i launch my app though, the dates appear for like 1/2 a second then disappear again I am lost.

This issue only occurs on my Android emulator running API 35. Tried on iOS and browser and no issues there.

Here is an image of what I mean:

Image

akravchukabto:
Fixes  #2591

### DIFF
--- a/src/agenda/index.tsx
+++ b/src/agenda/index.tsx
@@ -258,7 +258,10 @@ export default class Agenda extends Component<AgendaProps, State> {
     // scroll position actually changes (it would stay at 0, when scrolled to the top).
     this.setScrollPadPosition(this.initialScrollPadPosition(), false);
     // delay rendering calendar in full height because otherwise it still flickers sometimes
-    setTimeout(() => this.setState({calendarIsReady: true}), 0);
+    setTimeout(() => {
+      this.setState({calendarIsReady: true});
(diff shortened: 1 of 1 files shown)


ArturSousaFerreira (changes requested):
try and not resolve in my case

PR #2622, opened 2025-03-19
Fix: Adjenda crash in RN 0.78.0

dmohler20:
This fixes the Error: Maximum update depth exceeded error that occurs on the agenda in RN 0.78.0, but checking to see if a re render is necessary before doing so.

### DIFF
--- a/src/agenda/reservation-list/index.tsx
+++ b/src/agenda/reservation-list/index.tsx
@@ -107,37 +107,46 @@ class ReservationList extends Component<ReservationListProps, State> {
   }
 
   componentDidUpdate(prevProps: ReservationListProps) {
-    if (this.props.topDay && prevProps.topDay && prevProps !== this.props) {
+    if (this.props.topDay && prevProps.topDay) {
       if (!sameDate(prevProps.topDay, this.props.topDay)) {
         this.setState({reservations: []},
-          () => this.updateReservations(this.props)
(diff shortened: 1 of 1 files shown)


emmanuel-bonin (approved):

emmanuel-bonin:
Same here, it fixes the issue when applying this modification

NiharR27 (approved):
works after the above. cc @Inbal-Tish 🙏🏽

redsniper90:
I hope you merge this asap....honestly, 1 month for a simple fix that prevents many to fully use this library is a lil bit too much...

felipematheusmello:
It worked for me on Android. Checking for IOS and I'll return a feedback for you( Thanks a lot)

src/agenda/reservation-list/index.tsx
```diff
@@ -107,38 +107,47 @@ class ReservationList extends Component<ReservationListProps, State> {
...
       }
     }
   }
-
   updateDataSource(reservations: DayAgenda[]) {
     this.setState({reservations});
   }
 
-  updateReservations(props: ReservationListProps) {
+  updateReservations(props: ReservationListProps, prevProps: ReservationListProps) {
```
lucascarrias, line 127: Why do you need to pass the `prevProps` here if it is not used at `updateReservations`?

PR #2623, opened 2025-03-25
Remove scroll delay in AgendaList

nitzanyiz:
Added leading invocation to scrollToSection in AgendaList and InfiniteList AgendaList.

Ticket:
MADS-4420

### DIFF
--- a/src/expandableCalendar/agendaList.tsx
+++ b/src/expandableCalendar/agendaList.tsx
@@ -142,7 +142,7 @@ const AgendaList = forwardRef((props: AgendaListProps, ref: any) => {
     return sectionTitle;
   }, []);
 
-  const scrollToSection = useCallback(debounce((d) => {
+  const scrollToSection = useCallback((d) => {
     const sectionIndex = scrollToNextEvent ? getNextSectionIndex(d) : getSectionIndex(d);
     if (isUndefined(sectionIndex)) {
       return;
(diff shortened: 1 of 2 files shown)


src/expandableCalendar/agendaList.tsx
```diff
@@ -142,7 +141,7 @@ const AgendaList = forwardRef((props: AgendaListProps, ref: any) => {
     return sectionTitle;
   }, []);
 
-  const scrollToSection = useCallback(debounce((d) => {
+  const scrollToSection = useCallback((d) => {
```
Inbal-Tish, line 144: As we discussed, removing the debounce only going to cause an issue with the list's updates from others source. Let's add the option to control this debounce time

Inbal-Tish (approved):

src/expandableCalendar/infiniteAgendaList.tsx
```diff
@@ -145,7 +145,7 @@ const InfiniteAgendaList = ({
         _onMomentumScrollEnd(); // the RecyclerListView doesn't trigger onMomentumScrollEnd when calling scrollToSection
       }, 500);
     }
-  }, 1000, {leading: false, trailing: true}), [sections]);
+  }, 1000, {leading: true, trailing: true}), [sections]);
```
yakirza17, line 148: Why do we need both leading: true, trailing: true?
It will lead to 2 invocations
@nitzanyiz @Inbal-Tish

nitzanyiz: The leading was set so for a single press it will invoke immediately. The trailing was set so the last one would be called if there are multiple calls to this. Its a compromise so we don't lower the timing or remove it entirely.

PR #2540, opened 2024-10-09
fix: re-render on theme prop changes

aminmeshk:
See #2539 for explanation. 
Added `useEffect` to update styles when theme prop changes.

PR #2626, opened 2025-03-26
Replace the calendar ref style with a state-based style to fix theme changes

youcefswoo:
The issue I had was that when I changed my phone's theme to dark mode or light mode, the calendar didn't apply the new style. So, I made some changes to make it work, and it works perfectly on my end. I believe this approach can also be used in other components like the agenda.


rklomp:
Related to #2540 ?

PR #2627, opened 2025-03-27
refactor(ESModules): replace requires by imports in source files

### LINKED ISSUE #2396: Mixing require and import
Please make our job easier by filling this template out to completion. If you're requesting a feature instead of reporting a bug, please feel free to skip the Environment and Reproducible Demo sections.

## Description

I'm using vite to include this in a react-native-web project and rollup is leaving the require statements in the code without bundling the dependencies because you are mixing import and require statements in the same file. See [import](https://github.com/wix/react-native-calendars/blob/master/src/componentUpdater.ts#L6) and [require](https://github.com/wix/react-native-calendars/blob/master/src/componentUpdater.ts#L8)

## Expected Behavior

I expect a library to stick with either require or import. Not both

## Observed Behavior

```
Uncaught ReferenceError: require is not defined
```

## Environment

* `npm ls react-native`:

Also specify:

1. Device/emulator/simulator & OS version:

## Reproducible Demo

Please provide a minimized reproducible demonstration of the problem you're reporting.

Issues that come with minimal repro's are resolved much more quickly than issues where a maintainer has to reproduce themselves.

## Screenshots

Screenshots or gifs of the issue and the suggested fix will help us move faster with the review process.

JohnBerd:
This package is incompatible with storybook 8+ with `vite` because `require` is not allowed in `ES modules`.

Fixes #2396

### DIFF
--- a/src/dateutils.ts
+++ b/src/dateutils.ts
@@ -1,5 +1,7 @@
-const XDate = require('xdate');
-const {toMarkingFormat} = require('./interface');
+import XDateClass from 'xdate';
+import {toMarkingFormat} from './interface';
+
+const XDate = XDateClass as any;
 
 const latinNumbersPattern = /[0-9]/g;
 
--- a/src/images.d.ts
+++ b/src/images.d.ts
@@ -1,0 +1,4 @@
+declare module '*.png' {
+  const value: import('react-native').ImageSourcePropType;
+  export default value;
+}
--- a/src/calendar/header/index.tsx
+++ b/src/calendar/header/index.tsx
@@ -19,6 +19,8 @@ import {
 import {formatNumbers, weekDayNames} from '../../dateutils';
 import styleConstructor from './style';
 import {Theme, Direction} from '../../types';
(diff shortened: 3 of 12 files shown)


src/images.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.png' {
+  const value: import('react-native').ImageSourcePropType;
+  export default value;
+}
```
JohnBerd, line 4: Without this file, `typescript` does not allow to import `png`

src/dateutils.ts
```diff
@@ -1,5 +1,7 @@
-const XDate = require('xdate');
-const {toMarkingFormat} = require('./interface');
+import XDateClass from 'xdate';
+import {toMarkingFormat} from './interface';
+
+const XDate = XDateClass as any;
```
JohnBerd, line 4: `require` types modules as `any`, since this change is leading to a lot of typing problems, I prefered to keep the same behavior as before

JohnBerd:
I need to know if it is worth resolving conflicts or this should be closed

PR #2628, opened 2025-03-30
ExpandableCalendar - disable auto date selection

Inbal-Tish:
ExpandableCalendar - disable auto date selection by passing an array of navigationTypes to point on which event to disable the auto selection.

ticker 4419


src/expandableCalendar/week.tsx
```diff
@@ -56,6 +56,11 @@ const Week = React.memo((props: WeekProps) => {
   const dayProps = extractDayProps(props);
   const currXdate = useMemo(() => parseDate(current), [current]);
 
+  const getDayState = (day: XDate) => {
+    const isControlled = disableDaySelection;
```
ethanshar, line 60: let's move it back to state-manager to encapsulate this logic

src/calendar/index.tsx
```diff
@@ -187,23 +187,26 @@ const Calendar = (props: CalendarProps & ContextProp) => {
     );
   };
 
-  const renderDay = (day: XDate, id: number) => {
-    const dayProps = extractDayProps(props);
+  const getDayState = (day: XDate) => {
+    const isControlled = isEmpty(props.context);
```
ethanshar, line 191: let's move it back to state-manager to encapsulate this logic

ethanshar (approved):

src/expandableCalendar/week.tsx
Inbal-Tish: Done

PR #2633, opened 2025-04-06
Hide ExpandableCalendar header when going out of container

nitzanyiz:
Added overflow hidden so vertical calendar header won't go outside of the component.

### DIFF
--- a/src/expandableCalendar/index.tsx
+++ b/src/expandableCalendar/index.tsx
@@ -260,7 +260,7 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
   }, [allowShadow, propsStyle, headerHeight]);
 
   const wrapperStyle = useMemo(() => {
-    return {height: deltaY};
+    return {height: deltaY, overflow: 'hidden'};
   }, [deltaY]);
 
   const numberOfDaysHeaderStyle = useMemo(() => {
--- a/src/expandableCalendar/index.tsx
+++ b/src/expandableCalendar/index.tsx
@@ -256,11 +256,11 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
   }, [isOpen, headerHeight]);
 
   const containerStyle = useMemo(() => {
-    return [allowShadow && style.current.containerShadow, propsStyle, headerHeight === 0  && style.current.hidden];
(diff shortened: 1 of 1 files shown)


src/expandableCalendar/index.tsx
```diff
@@ -260,7 +260,7 @@ const ExpandableCalendar = (props: ExpandableCalendarProps) => {
   }, [allowShadow, propsStyle, headerHeight]);
 
   const wrapperStyle = useMemo(() => {
-    return {height: deltaY};
+    return {height: deltaY, overflow: 'hidden'};
```
Inbal-Tish, line 263: You have an error here since this is the animated style. You can move it to the containerStyle.

Inbal-Tish:
@nitzanyiz Can you please check why the header doesn't disappear when you open the calendar with know press?

src/expandableCalendar/index.tsx
nitzanyiz: I believe we need to add animation for opening. There is only one for the closing of the header. We can add this change to the closeHeader function (and rename it)
```tsx
headerDeltaY.current.setValue(Number(_headerStyles.style.top)); // set the start position for the animated value
if (!horizontal) {
  Animated.spring(headerDeltaY.current, {
    toValue: isOpen ? -headerHeight : 0,
    speed: SPEED / 10,
    bounciness: 1,
    useNativeDriver: false
  }).start();
}
```

Inbal-Tish:
> @Inbal-Tish I believe we need to add animation for opening. There is only one for the closing of the header. We can add this change to the closeHeader function (and rename it)
> 
> ```tsx
> headerDeltaY.current.setValue(Number(_headerStyles.style.top)); // set the start position for the animated value
> if (!horizontal) {
>   Animated.spring(headerDeltaY.current, {
>     toValue: isOpen ? -headerHeight : 0,
>     speed: SPEED / 10,
>     bounciness: 1,
>     useNativeDriver: false
>   }).start();
> }
> ```

Sounds good. Do you want to do it in this PR or a new one?

Inbal-Tish (approved):

nitzanyiz:
I'll create a different pr. Lets keep this simple.

PR #2634, opened 2025-04-06
ExpandableCalendar - Added imperative Calendar Toggling

nitzanyiz:
Added imperative toggling of the calendar


src/index.ts
```diff
@@ -6,7 +6,7 @@ export type {CalendarListProps} from './calendar-list';
 export {default as Agenda} from './agenda';
 export type {AgendaProps} from './agenda';
 export {default as ExpandableCalendar} from './expandableCalendar';
-export type {ExpandableCalendarProps} from './expandableCalendar';
+export type {ExpandableCalendarProps, ExpandableCalendarRef} from './expandableCalendar';
```
Inbal-Tish, line 9: Why are you exposing the ref? The user should hold it's own ref and access the imperative methods using it

nitzanyiz: its just the type so the user can you it.

src/expandableCalendar/index.tsx
```diff
@@ -94,7 +98,7 @@ const headerStyleOverride = {
  * @example: https://github.com/wix/react-native-calendars/blob/master/example/src/screens/expandableCalendar.js
  */
 
-const ExpandableCalendar = (props: ExpandableCalendarProps) => {
+const ExpandableCalendar: ForwardRefRenderFunction<ExpandableCalendarRef, ExpandableCalendarProps> = (props, ref) => {
```
Inbal-Tish, line 101: Why not simply wrap with 'forwardRef'? What do we need `ForwardRefRenderFunction<ExpandableCalendarRef, ExpandableCalendarProps> ` for?

src/index.ts
Inbal-Tish: See what we're doing in the other components

nitzanyiz: Removed.

example/src/screens/expandableCalendarScreen.tsx
```diff
@@ -13,7 +14,7 @@ const ITEMS: any[] = agendaItems;
 interface Props {
   weekView?: boolean;
 }
-
+const CHEVRON = require('../img/chevronUp.png');
```
Inbal-Tish, line 17: This source doesn't look good, it's too big. You can use the next and change the rotation degrees if you want or create another source from it

example/src/screens/expandableCalendarScreen.tsx
```diff
@@ -86,8 +118,11 @@ const styles = StyleSheet.create({
     paddingRight: 20
   },
   header: {
-    backgroundColor: 'lightgrey'
+    flexDirection: 'row',
+    justifyContent: 'center',
+    alignItems: 'center'
```
Inbal-Tish, line 123: Please add  marginVertical: 10

example/src/screens/expandableCalendarScreen.tsx
nitzanyiz: Change to use next

Inbal-Tish (approved):

francoangulo:
Hello @nitzanyiz this adds a lot of flexibility to the calendar, could it be possible to close() / open() it programmatically instead of toggling() it?

PR #2638, opened 2025-04-09
Patch recyclerlistview

Inbal-Tish:
Add patch for recyclerlistview with fix on VistualRenderer updateOffset() for RTL in RN73 and above

Yarn 3.4.1 (min 2 is required)

Ticket 4668


ethanshar:
@Inbal-Tish 
There's not much to review here, do you know why the pathed file is empty? Doesn't it suppose to have the fixed code of the file or something? 
I wonder if will work when installed in another package, can we create a temp version to test in uilib?

Inbal-Tish:
> @Inbal-Tish There's not much to review here, do you know why the pathed file is empty? Doesn't it suppose to have the fixed code of the file or something? I wonder if will work when installed in another package, can we create a temp version to test in uilib?

Fixed. I parched the dist file instead

PR #2641, opened 2025-04-16
ExpandableCalendar - default header height

nitzanyiz:


src/expandableCalendar/index.tsx
```diff
@@ -141,10 +141,10 @@ const ExpandableCalendar = forwardRef<ExpandableCalendarRef, ExpandableCalendarP
   } = props;
 
   const [screenReaderEnabled, setScreenReaderEnabled] = useState(false);
-  const [headerHeight, setHeaderHeight] = useState(0);
+  const [headerHeight, setHeaderHeight] = useState(constants.isWeb ? 78 : 0);
   const onHeaderLayout = useCallback(({nativeEvent: {layout: {height}}}: LayoutChangeEvent) => {
-      setHeaderHeight(height);
-  }, []);
+      !constants.isWeb && setHeaderHeight(height);
```
ethanshar, line 146: You mentioned that on web the event called with height `0` so maybe we can set the header height if it's different than zero and not if it's not web, this way, once they upgrade RNW it will work, right?

ethanshar (approved):
I Approved, but see my suggestion

src/expandableCalendar/index.tsx
nitzanyiz: Yeah. Guess we can do that

nitzanyiz: Yeah, I guess we can do that. Changed.

ethanshar (approved):

PR #2642, opened 2025-04-16
Removed hidden for accessible calendar

nitzanyiz:
We can (and should probably) hf this.

### DIFF
--- a/src/expandableCalendar/index.tsx
+++ b/src/expandableCalendar/index.tsx
@@ -269,7 +269,7 @@ const ExpandableCalendar = forwardRef<ExpandableCalendarRef, ExpandableCalendarP
   }, [isOpen, headerHeight]);
 
   const containerStyle = useMemo(() => {
-    return [allowShadow && style.current.containerShadow, propsStyle, headerHeight === 0 && !screenReaderEnabled  && style.current.hidden, {overflow: 'hidden'} as const];
+    return [allowShadow && style.current.containerShadow, propsStyle, headerHeight === 0 && style.current.hidden, {overflow: 'hidden'} as const];
   }, [allowShadow, propsStyle, headerHeight, screenReaderEnabled]);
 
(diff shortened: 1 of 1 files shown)


src/expandableCalendar/index.tsx
```diff
@@ -269,8 +269,8 @@ const ExpandableCalendar = forwardRef<ExpandableCalendarRef, ExpandableCalendarP
   }, [isOpen, headerHeight]);
 
   const containerStyle = useMemo(() => {
-    return [allowShadow && style.current.containerShadow, propsStyle, headerHeight === 0  && style.current.hidden, {overflow: 'hidden'} as const];
-  }, [allowShadow, propsStyle, headerHeight]);
+    return [allowShadow && style.current.containerShadow, propsStyle, headerHeight === 0 && !screenReaderEnabled  && style.current.hidden, {overflow: 'hidden'} as const];
```
Inbal-Tish, line 272: I prefer connecting the accessibility calendar to the onLayout then adding this patch

src/expandableCalendar/index.tsx
```diff
@@ -269,8 +269,8 @@ const ExpandableCalendar = forwardRef<ExpandableCalendarRef, ExpandableCalendarP
   }, [isOpen, headerHeight]);
 
   const containerStyle = useMemo(() => {
-    return [allowShadow && style.current.containerShadow, propsStyle, headerHeight === 0  && style.current.hidden, {overflow: 'hidden'} as const];
-  }, [allowShadow, propsStyle, headerHeight]);
+    return [allowShadow && style.current.containerShadow, propsStyle, headerHeight === 0 && style.current.hidden, {overflow: 'hidden'} as const];
+  }, [allowShadow, propsStyle, headerHeight, screenReaderEnabled]);
```
Inbal-Tish, line 273: You forgot the dependency here

Inbal-Tish (approved):

PR #2645, opened 2025-04-20
ExpandableCalendar - TogglePosition with disablePan

nitzanyiz:
Removed `bouneToPosition` disabling with disablePan. Disabled header closing on day press with disablePan.


src/expandableCalendar/index.tsx
```diff
@@ -397,8 +397,8 @@ const ExpandableCalendar = forwardRef<ExpandableCalendarRef, ExpandableCalendarP
 
   /** Animated */
 
-  const bounceToPosition = (toValue = 0) => {
-    if (!disablePan) {
+  const bounceToPosition = (toValue = 0, manual = false) => {
+    if (!disablePan || manual) {
```
Inbal-Tish, line 401: Can't we just remove the condition?

Inbal-Tish (approved):

Issue #2646, opened 2025-04-21
ExpandableCalendar initial position "open" is partially cut off in latest version (v1.1311.0)


juanmigdr:
**Description:**

In the latest version of `react-native-calendars`, setting the `ExpandableCalendar`'s initial position to `"open"` does not fully display the calendar. Only part of the calendar appears, and it seems to be cut off visually.

This issue can be reproduced using one of the existing examples provided in the repo — specifically, the [`TimelineScreen` example](https://github.com/wix/react-native-calendars/blob/master/example/src/screens/timelineCalendarScreen.tsx).

This behavior was **not** present in a previous version (`1.1307.0`) where the calendar would correctly render in its fully expanded state when the `initialPosition` prop was set to `Positions.OPEN`.

**Code Sample:**
I just added the following to the example `initialPosition={Positions.OPEN}`

```tsx
import groupBy from "lodash/groupBy";
import filter from "lodash/filter";
import find from "lodash/find";

import React, { Component } from "react";
import { Alert } from "react-native";
import {
  ExpandableCalendar,
  TimelineEventProps,
  TimelineList,
  CalendarProvider,
  TimelineProps,
  CalendarUtils,
} from "react-native-calendars";

import { timelineEvents, getDate } from "./timelineEvents";
import { Positions } from "react-native-calendars/src/expandableCalendar";

const INITIAL_TIME = { hour: 9, minutes: 0 };
const EVENTS: TimelineEventProps[] = timelineEvents;
export default class TimelineCalendarScreen extends Component {
  state = {
    currentDate: getDate(),
    events: EVENTS,
    eventsByDate: groupBy(EVENTS, (e) =>
      CalendarUtils.getCalendarDateString(e.start)
    ) as {
      [key: string]: TimelineEventProps[];
    },
  };

  marked = {
    [`${getDate(-1)}`]: { marked: true },
    [`${getDate()}`]: { marked: true },
    [`${getDate(1)}`]: { marked: true },
    [`${getDate(2)}`]: { marked: true },
    [`${getDate(4)}`]: { marked: true },
  };

  onDateChanged = (date: string, source: string) => {
    console.log("TimelineCalendarScreen onDateChanged: ", date, source);
    this.setState({ currentDate: date });
  };

  onMonthChange = (month: any, updateSource: any) => {
    console.log("TimelineCalendarScreen onMonthChange: ", month, updateSource);
  };

  createNewEvent: TimelineProps["onBackgroundLongPress"] = (
[...]

PR #2661, opened 2025-05-14
Fix ExpandableCalendar StartHeight calculations

KimGnab:
**Problem Cause:**
When `ExpandableCalendar` was configured with `initialPosition={Positions.OPEN}`, the calendar's total open height `startHeight` was calculated before the actual height of the header section `headerHeight` was determined. This led to headerHeight being treated as `0` or an incorrect value, causing the calendar content to appear clipped or the layout to be misaligned, even though the calendar was in an open state.

**Solution:**
1. Fixed that when the `headerHeight` value changes, the logic that calculates the height when the calendar is open `getOpenHeight` immediately reflects this and recalculates the startHeight correctly.
2.  Also, synchronize the correctly calculated startHeight value to the object that manages native component styles internally `_wrapperStyles` to ensure the correct height is applied to the actual UI.

#2646


nitzanyiz (approved):
Thank you, @KimGnab 🙏

PR #2668, opened 2025-05-20
Timeline - fix api file and add one for TimelineList

Inbal-Tish:
Timeline - fix api file and add one for TimelineList

### DIFF
--- a/src/timeline-list/timelineList.api.json
+++ b/src/timeline-list/timelineList.api.json
@@ -1,0 +1,47 @@
+{
+  "name": "TimelineList",
+  "description": "TimelineList component",
+  "images": [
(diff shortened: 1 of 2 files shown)


src/timeline-list/timelineList.api.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "TimelineList",
+  "description": "TimelineList component",
+  "images": [
+    "https://github.com/wix/react-native-calendars/blob/master/demo/assets/timeline-calendar.gif?raw=true"
+  ],
+  "extends": ["InfiniteList"],
```
nitzanyiz, line 7: It doesn't seem to extend `InfiniteList`.

nitzanyiz (approved):

PR #2707, opened 2025-07-16
fix: correctly type style override props

AmauryLiet:
Current typing incorrectly enforces passing styling props as:
```tsx
<Calendar
    theme={{
      stylesheet: {
        calendar: {
          header: {
            dayTextAtIndex0: { color: 'fuchsia', fontSize: 42 },
          }
        }
      }
    }}
/>
```
although this leads to ignored styling. Correct usage is:
```tsx
<Calendar
    theme={{
        dayTextAtIndex0: { color: 'fuchsia', fontSize: 42 },
      }
    }}
/>
```

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -75,28 +75,16 @@ export interface Theme {
   arrowWidth?: number;
   weekVerticalMargin?: number;
   reservationsBackgroundColor?: string;
-  stylesheet?: {
-    calendar?: {
-      main?: object;
-      header?: object;
-    };
-    day?: {
-      basic?: object;
-      period?: object;
-    };
-    dot?: object;
-    marking?: object;
-    'calendar-list'?: {
-      main?: object;
-    };
-    agenda?: {
-      main?: object;
-      list?: object;
-    };
-    expandable?: {
-      main?: object;
-    };
-  };
+  'stylesheet.calendar.main'?: object;
+  'stylesheet.calendar.header'?: object;
+  'stylesheet.day.basic'?: object;
+  'stylesheet.day.period'?: object;
+  'stylesheet.dot'?: object;
+  'stylesheet.marking'?: object;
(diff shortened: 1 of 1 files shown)


SWARVY:
@AmauryLiet

Thank you so much for uploading this PR! 

I'd like to share a suggestion regarding styling consistency that I've noticed while working with the library. Currently, there are two different approaches for styling calendar components in the original source code:

1. [Other Custom style properties](https://github.com/wix/react-native-calendars/blob/master/src/calendar/style.ts): Can be modified using stylesheet.~... approach
```ts
export default function getStyle(theme: Theme = {}) {
    ...other things...,
    ...(theme['stylesheet.calendar.main'] || {})
  });
}
```

2. [ExpandableCalendar properties](https://github.com/wix/react-native-calendars/blob/master/src/expandableCalendar/style.ts): Must use the currently defined types for styling to be applied
```ts
export default function styleConstructor(theme: Theme = {}) {
  const appStyle = {...defaultStyle, ...theme};

  return StyleSheet.create({
    ...other things...,
    ...(theme?.stylesheet?.expandable?.main || {})
  });
}
```

As a developer using this library, I've found this inconsistency to be a bit confusing when trying to understand which styling approach to use. It would be really helpful if we could have a more consistent styling pattern across all calendar components.

Suggestion: Would it be possible to consider standardizing the styling approach (either all stylesheet.~... or all current types) and applying it uniformly? This would make the library much easier to use and maintain for the community.

AmauryLiet:
@Inbal-Tish any input on this one? or a recommendation for another reviewer?

AmauryLiet:
@SWARVY it would totally make sense to uniformize the styling, you could submit another PR in that sense. However keep in mind that this would be a breaking change and will need an appropriate version bump

PR #2713, opened 2025-07-30
fix agenda reservation list exceeded update depth for expo sdk53

raeinlee:
This pull request will fix the issue mentioned here: #2705 

Fix: Avoid infinite update loop in Agenda > ReservationList on Expo SDK 53+

**Problem**

When using react-native-calendars in Expo SDK 53+, the app crashes due to a Maximum update depth exceeded error in the ReservationList component of the Agenda.

This was caused by an infinite re-render triggered in componentDidUpdate, due to the state being set on every prop change without proper guarding.

**Solution**

Updated the componentDidUpdate method to check for meaningful changes in props (showOnlySelectedDayItems, items, and selectedDay) before calling setState or updating reservations.

This patch was submitted using @dangthequan 's contribution.

PR #2742, opened 2025-11-17
Fix ReservationList infinite update loop under React 18/19

mazdak:
## Summary

This change fixes a potential infinite update loop in the Agenda’s ReservationList when used with newer React versions (18/19+).

The current implementation of ReservationList.componentDidUpdate uses `prevProps !== this.props` as part of its guard:

With React 18/19 and the new architecture, props objects are often recreated on each render. As a result, `prevProps !== this.props` can be true on every update, even when the values have not changed. Combined with setState in componentDidUpdate, this can cause ReservationList to continually update, leading to:

-  Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate.
-  A component stack pointing at ReservationList → Agenda.

## What this PR changes

ReservationList.componentDidUpdate is updated to only respond when:

- The visible topDay actually changes (by value), or
- The items reference changes while topDay stays the same.

This keeps the original behavior (recomputing reservations when topDay moves or items change) but avoids treating a new props object as a guaranteed change. In practice this:

- Prevents unnecessary updateReservations calls on every render.
- Avoids “Maximum update depth exceeded” loops under React 18/19/new architecture.

## Impact

- No public API changes.
- No behavior change for callers that update topDay or items explicitly.
- Improves compatibility with React 18/19 and React Native new architecture, where props identity is not stable.

I’ve tested this change against a real app using Agenda under React 19 / RN 0.81, where it eliminates the infinite update crash (ReservationList → Agenda) without regressing scrolling behavior.

### DIFF
--- a/src/agenda/reservation-list/index.tsx
+++ b/src/agenda/reservation-list/index.tsx
@@ -107,14 +107,31 @@ class ReservationList extends Component<ReservationListProps, State> {
   }
 
   componentDidUpdate(prevProps: ReservationListProps) {
-    if (this.props.topDay && prevProps.topDay && prevProps !== this.props) {
-      if (!sameDate(prevProps.topDay, this.props.topDay)) {
-        this.setState({reservations: []},
-          () => this.updateReservations(this.props)
-        );
-      } else {
-        this.updateReservations(this.props);
-      }
+    const {topDay, items} = this.props;
+    const {topDay: prevTopDay, items: prevItems} = prevProps;
+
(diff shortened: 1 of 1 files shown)


feri-irawan:
We need this

Dcrosssensei:
What happened to that? This PR is the only solution to the current problem. we need this solution in production

justinchan23:
this didnt work for me but https://github.com/wix/react-native-calendars/pull/2713 did

PR #2761, opened 2026-01-29
Fix/dedupe and 2748

M-i-k-e-l:
Add the request in #2748 along with `yarn dedupe` & `pos install`

The comment from the PR's code:
Prevent FlatList from rendering with empty data to avoid React Native bug
where FlatList gets stuck and won't render items added later.
See: https://github.com/facebook/react-native/issues/39421


adids1221 (approved):
🚀

PR #2771, opened 2026-03-16
Migrate StyleSheet.absoluteFillObject

jenskuhrjorgensen:
Migrate StyleSheet.absoluteFillObject which has been removed in react-native 0.85: https://github.com/facebook/react-native/releases/tag/v0.85.0-rc.0


jerone:
Without this change, long press is breaking. :shipit:

PR #2780, opened 2026-04-18
Fix showSixWeeks page() using aligned span instead of /6 heuristic

Norcy:
## Problem
When `showSixWeeks` is true, `page()` used `(daysBefore + monthLength) / 6 >= 6` to decide whether to extend `to` by 7 days. This does not match the real week-aligned cell count and caused:
- An extra 7th row in some months (e.g. **March 2026** with `firstDay=1`).
- Missing trailing next-month cells when the heuristic skipped the extension (e.g. **May 2026** stayed at 5 weeks).

## Solution
After aligning `from` and `to` to week boundaries, extend by one week only when `from.diffDays(to) + 1 < 42`. Remove the `/6` heuristic and the unused `daysBefore` variable.

## Testing
`yarn test src/dateutils.spec.js`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/dateutils.ts
+++ b/src/dateutils.ts
@@ -136,7 +136,6 @@ export function page(date: XDate, firstDayOfWeek = 0, showSixWeeks = false) {
   firstDayOfWeek = firstDayOfWeek || 0;
 
   const from = days[0].clone();
-  const daysBefore = from.getDay();
 
   if (from.getDay() !== fdow) {
     from.addDays(-(from.getDay() + 7 - fdow) % 7);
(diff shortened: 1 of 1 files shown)


CreatlV:
Thanks for tackling this — using the aligned page span is much clearer than the current `/ 6` heuristic.

There is one remaining edge case: February 2026 with `firstDay={0}`. After alignment, `from` through `to` spans 28 days, so adding a single week produces 35 days (five rows), not 42.

Could the extension add the exact missing number of days instead?

```ts
const daysOnPage = from.diffDays(to) + 1;

if (showSixWeeks && daysOnPage < 42) {
  to.addDays(42 - daysOnPage);
}
```

That gives 28 → 42, 35 → 42, and leaves 42 unchanged. It would be good to add regression cases for February 2026 with a Sunday start, May 2026 with a Monday start, and a month that naturally spans six rows. I also checked this calculation across 1900–2100 with `firstDay` values 0–6; every `showSixWeeks` page resolves to 42 days.

